PR #125, opened 2022-03-29
Feature/#66 - Refactor JobConfig + Add StandaloneConfig

khaitruong922:
- Seperate standalone config properties from JobConfig into StandaloneConfig class.
- Do some refactoring to reduce boilerplate code.


florian-vuillemot (dismissed):
Good job !
Blocked until V1.0.0

taipy/core/config/config.py
```diff
@@ -123,7 +123,6 @@ def configure_global_app(
     def configure_job_executions(
         cls,
         mode: str = None,
-        nb_of_workers: Union[int, str] = None,
```
jrobinAV: Should remove the parameter in the docstring as well.

taipy/core/config/job_mode_config.py
```diff
@@ -0,0 +1,5 @@
+from typing import Any, Dict
+
+
+class JobModeConfig:
```
jrobinAV: Is this class meant to be exposed?

taipy/core/config/standalone_config.py
```diff
@@ -0,0 +1,18 @@
+from typing import Any, Dict, Union
+
+from taipy.core.config.job_mode_config import JobModeConfig
+
+
+class StandaloneConfig(JobModeConfig):
```
jrobinAV: Is this class exposed to the user?

taipy/core/config/standalone_config.py
```diff
@@ -0,0 +1,18 @@
...
+    """
+    Holds configuration fields related to the job executions.
+
+    Parameters:
+        nb_of_workers (int): The maximum number of running workers to execute jobs. It must be a positive integer.
+            The default value is 1.
+        other_properties (dict): A dictionary of additional properties.
+    """
```
jrobinAV: I don't see these fields.

taipy/core/config/job_config.py
```diff
@@ -13,70 +15,63 @@ class JobConfig:
...
+            if value is not None:
+                self.config[k] = value
+
+    @classmethod
+    def _config_cls(cls, mode: str) -> Type[JobModeConfig]:
+        if mode == cls._DEFAULT_MODE:
+            return StandaloneConfig
+        dep = f"taipy.{mode}"
+        if not util.find_spec(dep):
+            raise DependencyNotInstalled(mode)
```
jrobinAV: I'm not too fond of coupling the name of the package with a field in the config. Imagine we want to replace the `taipy-airflow` name with something else (whatever the reason). Then, we will need to change this implementation and change the API (the config field is exposed to the user).

I would prefer to map the mode and the package name explicitly.

taipy/core/config/job_config.py
```diff
@@ -13,70 +15,63 @@ class JobConfig:
...
 
     _MODE_KEY = "mode"
     _DEFAULT_MODE = "standalone"
 
-    _NB_OF_WORKERS_KEY = "nb_of_workers"
-    _DEFAULT_NB_OF_WORKERS = 1
+    def __init__(self, mode: str = None, **properties):
+        self.mode = mode or self._DEFAULT_MODE
+        self.config_cls = self._config_cls(self.mode)
+        self.config = self._create_config(self.config_cls, **properties)
```
jrobinAV: Shouldn't they be hidden with an underscore?

taipy/core/config/job_config.py
```diff
@@ -13,70 +15,63 @@ class JobConfig:
     Parameters:
         mode (str): The Taipy operating mode. By default, the "standalone" mode is set. On Taipy enterprise,
             the "airflow" mode is available.
-        nb_of_workers (int): The maximum number of running workers to execute jobs. It must be a positive integer.
-            The default value is 1.
```
jrobinAV, line 28: Since JobConfig is exposed, I believe we want to keep this documentation.

jrobinAV (changes requested):
1 - At this point, I don't really get the purpose of this refactoring. It brings more complexity. I guess I need to review the PR on airflow for a better understanding.

2 - Since you are using inheritance, can't we just make JobConfig a parent class, and Standalone its child class? I would remove some complexity. Or am I missing something?

3- It seems to work, but can you test if the job_config is well exported and reloaded to/from a TOML file? With and without an environment variable value.

taipy/core/config/job_mode_config.py
khaitruong922: No I think

taipy/core/config/job_config.py
khaitruong922: I would prefer that too

taipy/core/config/standalone_config.py
khaitruong922: I will update the docstring

taipy/core/config/config.py
khaitruong922: It is already removed

taipy/core/config/job_config.py
khaitruong922: It is now belong to StandaloneConfig

jrobinAV: I would keep this documentation. I would also keep the nb_of_worker parameter as Optional with a default value =1. And I would explicitly document that it is used only in standalone mode.

The purpose of this is to keep autocompletion for the basic usage of this method, which is the standalone mode.

taipy/core/config/job_config.py
```diff
@@ -24,70 +26,68 @@ class JobConfig:
...
-    _NB_OF_WORKERS_KEY = "nb_of_workers"
-    _DEFAULT_NB_OF_WORKERS = 1
+    _MODE_TO_MODULE: Dict[str, str] = {
+        "airflow": "taipy.airflow",
+    }
 
-    def __init__(self, mode: str = None, nb_of_workers: Union[int, str] = None, **properties):
-        self.mode = mode
+    def __init__(self, mode: str = None, **properties):
+        self.mode = mode or self._DEFAULT_MODE
```
jrobinAV, line 40: Following the previous comment, I would let the nb_of_workers parameter in the method signature.

taipy/core/config/job_config.py
```diff
@@ -24,70 +26,68 @@ class JobConfig:
...
+        self._config_cls = self._get_config_cls(self.mode)
+        self._config = self._create_config(self._config_cls, **properties)
```
jrobinAV, line 42: Following the previous comment, I would do:

```suggestion
        self._config = self._create_config(self._config_cls, nb_of_workers, **properties)
```

jrobinAV (commented):
I would like to keep the autocompletion for the nb_of_workers in standalone mode.
What do you think of my comment guys? @khaitruong922 @florian-vuillemot 

If we keep the nb_of_workers in the job_config init signature, and in the Config.configure_job_execution method, I believe we can hide the StandaloneConfig class.

jrobinAV (approved):

jrobinAV:
For me, it s good to go on develop.

PR #153, opened 2022-04-11
Feature/#27 release and publish

florian-vuillemot:
https://app.zenhub.com/workspaces/taipy-613ef328affde6000ff77fb3/issues/avaiga/taipy/27


.github/workflows/release.yml
```diff
@@ -25,6 +42,10 @@ jobs:
         run: echo "##[set-output name=tag;]$(echo ${GITHUB_REF#refs/heads/release/})"
         id: extract_branch
 
+      - name: Ensure tag is and package version are equals
```
jrobinAV:
```suggestion
      - name: Ensure tag and package version are equals
```
```suggestion
      - name: Ensure tag is and package version are equals
```

.github/workflows/release.yml
```diff
@@ -1,21 +1,38 @@
-name: release workflow
+name: Create Github Release and Push on Pypi
 
 on:
   push:
     branches:
       - 'release/**'
```
jrobinAV: I would change this. 
I believe we don't want to release and publish each time we push on a release/* branch.
I believe it is more convenient to manually trigger the release and publish action.

So that we can create a release branch, share it with others, eventually add new fixes/features, review it, etc. Then, when we are happy about the version, we can release, tag, and publish it.

What do you think? @FabienLelaquais @florian-vuillemot

jrobinAV: I would change this. 
I believe we don't want to release and publish each time we push on a release/* branch.
I believe it is more convenient to manually trigger the release and publish action.

So that we can create a release branch, share it with others, eventually add new fixes/features, review it, etc. Then, when we are happy about the version, we can release, tag, and publish it.

What do you think? @FabienLelaquais @florian-vuillemot

jrobinAV (changes requested):

.github/workflows/release.yml
FabienLelaquais: I completely agree with that!

Generating a release on each push of a release branch is fine for me.
But publishing should be a delicate and rare decision. Manual then.

jrobinAV (dismissed):

.github/workflows/release.yml
```diff
@@ -1,29 +1,47 @@
...
+    runs-on: ubuntu-18.04
 
     steps:
       - uses: actions/checkout@v2
 
-      - name: Extract branch tag
-        shell: bash
-        run: echo "##[set-output name=tag;]$(echo ${GITHUB_REF#refs/heads/release/})"
-        id: extract_branch
+      - name: Ensure package version is well set
```
FabienLelaquais, line 42: is well set -> is properly set
or
is correct

FabienLelaquais (commented):
Nickel

PR #156, opened 2022-04-13
Update dependencies

florian-vuillemot:


Pipfile
```diff
@@ -4,12 +4,11 @@ verify_ssl = true
...
 
 [packages]
-networkx = ">=2.6"
-numpy = ">=1.21"
-openpyxl = ">=3.0"
-pandas = ">=1.3"
-sqlalchemy = ">=1.3"
-toml = ">=0.10"
+networkx = "==2.6"
+openpyxl = "==3.0.3"
```
FabienLelaquais, line 8: Is == required here??

&gt;=3.0 wasn't enough? Or maybe we're lacking a &lt;[something]

setup.py
```diff
@@ -19,13 +19,11 @@
...
 requirements = [
-    "networkx",
-    "numpy",
-    "openpyxl",
-    "pandas",
-    "simple-websocket",
-    "sqlalchemy",
-    "toml",
+    "networkx>=2.6",
+    "openpyxl>=3.0.3",
```
FabienLelaquais, line 23: Something like that, yes.

FabienLelaquais (dismissed):
With my little comments... Thx.

Pipfile
jrobinAV: I am fine with setting an exact version if it is too complicated to follow the compatibility constraints.

jrobinAV (dismissed):

Pipfile
florian-vuillemot: _Pipfile_ is for development only, for installation _setup.py_ is used.
I think, setting minimal versions as constraints in the _Pipfile_ will ensure developers to work with the minimal version that a user can have on its machine.
If we need new functionality of a package, we should update the version in the _Pipfile_ and thus the _setup.py_ and we are sure to not break the user as we did.

florian-vuillemot: For _openpyxl_ only. Check https://pypi.org/project/openpyxl/#history
We should deal with the "patch/technical/.." version. There are using it as we used the "minor".

FabienLelaquais: I guess my point was: do we know of any issues with openpyxl&gt;=3.0?

jrobinAV (approved):

Pipfile
florian-vuillemot: Oh ^^' 
Yes, failed until the 3.0.3 with the following error:
image

PR #168, opened 2022-04-26
bug/fix airflow path in jobconfig

toan-quach:


taipy/core/config/job_config.py
```diff
@@ -33,7 +33,7 @@ class JobConfig:
     _DEFAULT_MODE = "standalone"
 
     _MODE_TO_MODULE: Dict[str, str] = {
-        "airflow": "taipy.airflow.scheduler",
+        "airflow": "taipy.airflow",
         "enterprise": "taipy.enterprise.core.scheduler.scheduler",
     }
```
khaitruong922: How about creating 2 different dicts: one for mapping to the Config class module and one for Scheduler class (to be more explicit), as enterprise is not a job mode and we don't have Config for enterprise.

toan-quach: ah yes exactly!! that's what I was talking about a "weird spot" in my approach. Have already pushed :D

taipy/core/config/job_config.py
```diff
@@ -33,8 +33,13 @@ class JobConfig:
...
-        "airflow": "taipy.airflow.scheduler",
-        "enterprise": "taipy.enterprise.core.scheduler.scheduler",
+        "airflow": "taipy.airflow",
+        "enterprise": "taipy.enterprise",
+    }
+
+    _MODE_TO_SCHEDULER_PATH: Dict[str, str] = {
+        "airflow": ".scheduler",
+        "enterprise": ".core.scheduler.scheduler",
     }
```
khaitruong922, line 43: We can add _MODE_TO_CONFIG_PATH mapping and refactor the logic in _get_config_cls too.

khaitruong922 (approved):

taipy/core/config/job_config.py
toan-quach: imo, the current version of _get_config_cls is good already, the reason the need for mode to scheduler is because enterprise and airflow version has different base path for config. What do you think? :D

khaitruong922: It is good already but in the future there may be cases where the Config class is not in the root of the package so I think it is better to have explicit path. What do you think?

toan-quach: Hm true, but I would say we can postpone it until the use case rises up as I dont think we have a need for it atm :D

PR #180, opened 2022-05-03
Feature/#179 - Replace "edition" with "edit"

khaitruong922:
- Replace "edition" with "edit" in all places.
- For public API, make the "edition" version to reference the "edit" version but add deprecated docstring and log a warning if they are used.
- Fix some mypy issues.


taipy/core/data/csv.py
```diff
@@ -32,12 +32,12 @@ class CSVDataNode(DataNode):
         id (str): The unique identifier of this data node.
         name (str): A user-readable name of this data node.
         parent_id (str): The identifier of the parent (pipeline_id, scenario_id, cycle_id) or `None`.
-        last_edition_date (datetime): The date and time of the last edition.
+        last_edit_date (datetime): The date and time of the last edit.
```
florian-vuillemot, line 35:
```suggestion
        last_edit_date (datetime): The date and time of the last modification.
```

florian-vuillemot:
```suggestion
        last_edit_date (datetime): The date and time of the last modification.
```

taipy/core/_scheduler/_job_dispatcher.py
```diff
@@ -53,14 +53,14 @@ def _dispatch(self, job: Job):
...
+            self.__unlock_edit_on_outputs(job)
             self.__logger.info(f"job {job.id} is skipped.")
 
     @staticmethod
     def _needs_to_run(task: Task) -> bool:
         """
-        Returns True if the task outputs are in cache and if the output's last edition date is prior the input's last
-        edition date.
+        Returns True if the task outputs are in cache and if the output's last edit date is prior the input's last
+        edit date.
```
florian-vuillemot: I find this commentary really complex. Can we have something simpler ?

For ex:
```suggestion
        Returns True if the task has no output or if at least one input was modified since the latest run.
```

taipy/core/data/data_node.py
```diff
@@ -53,12 +53,12 @@ class DataNode(_Entity):
         name (str): A user-readable name of this data node.
         parent_id (str): The identifier of the parent (pipeline_id, scenario_id, cycle_id) or
             None.
-        last_edition_date (datetime): The date and time of the last edition.
+        last_edit_date (datetime): The date and time of the last edit.
```
florian-vuillemot:
```suggestion
        last_edit_date (datetime): The date and time of the last modification.
```

taipy/core/data/in_memory.py
```diff
@@ -35,12 +35,12 @@ class InMemoryDataNode(DataNode):
         name (str): A user-readable name of this data node.
         parent_id (str): The identifier of the parent (pipeline_id, scenario_id, cycle_id) or
             `None`.
-        last_edition_date (datetime): The date and time of the last edition.
+        last_edit_date (datetime): The date and time of the last edit.
```
khaitruong922:
```suggestion
        last_edit_date (datetime): The date and time of the last modification.
```

taipy/core/data/pickle.py
```diff
@@ -31,12 +31,12 @@ class PickleDataNode(DataNode):
         name (str): A user-readable name of this data node.
         parent_id (str): The identifier of the parent (pipeline_id, scenario_id, cycle_id) or
             `None`.
-        last_edition_date (datetime): The date and time of the last edition.
+        last_edit_date (datetime): The date and time of the last edit.
```
khaitruong922:
```suggestion
        last_edit_date (datetime): The date and time of the last modification.
```

taipy/core/data/sql.py
```diff
@@ -36,12 +36,12 @@ class SQLDataNode(DataNode):
         name (str): A user-readable name of this data node.
         parent_id (str): The identifier of the parent (pipeline_id, scenario_id, cycle_id) or
             None.
-        last_edition_date (datetime): The date and time of the last edition.
+        last_edit_date (datetime): The date and time of the last edit.
```
khaitruong922:
```suggestion
        last_edit_date (datetime): The date and time of the last modification.
```

taipy/core/data/data_node.py
```diff
@@ -76,34 +76,49 @@ def __init__(
...
+        self._last_edit_date = val
+
     @property  # type: ignore
     @_self_reload("data")
     def last_edition_date(self):
-        return self._last_edition_date
+        """
+        Deprecated. Use last_edit_date instead.
+        """
+        self.__logger.warning("last_edition_date is deprecated. Use last_edit_date instead.")
```
florian-vuillemot: Add: https://docs.python.org/3/library/warnings.html

taipy/core/data/excel.py
```diff
@@ -40,12 +40,12 @@ class ExcelDataNode(DataNode):
         name (str): A user-readable name of this data node.
         parent_id (str): The identifier of the parent (pipeline_id, scenario_id, cycle_id) or
             `None`.
-        last_edition_date (datetime): The date and time of the last edition.
+        last_edit_date (datetime): The date and time of the last edit.
```
khaitruong922:
```suggestion
        last_edit_date (datetime): The date and time of the last modification.
```

taipy/core/data/generic.py
```diff
@@ -31,12 +31,12 @@ class GenericDataNode(DataNode):
         name (str): A user-readable name of the data node.
         parent_id (str): The identifier of the parent (pipeline_id, scenario_id, cycle_id) or
             `None`.
-        last_edition_date (datetime): The date and time of the last edition.
+        last_edit_date (datetime): The date and time of the last edit.
```
khaitruong922:
```suggestion
        last_edit_date (datetime): The date and time of the last modification.
```

taipy/core/data/data_node.py
khaitruong922: https://stackoverflow.com/questions/9595009/warnings-warn-vs-logging-warning

taipy/core/data/data_node.py
```diff
@@ -229,37 +259,51 @@ def write(self, data, job_id: Optional[JobId] = None):
...
+        self.lock_edit()
+
+    def unlock_edit(self, at: datetime = None, job_id: JobId = None):
+        """Unlocks the edit of the data node.
+
+        The _last_edit_date_ is updated.
 
         Parameters:
-            at (datetime): The optional datetime of the last edition.
+            at (datetime): The optional datetime of the last edit.
```
florian-vuillemot:
```suggestion
            at (datetime): The optional datetime of the last modification.
```

taipy/core/data/data_node.py
```diff
@@ -371,14 +415,14 @@ def __getitem__(self, items):
     @_self_reload("data")
     def is_ready_for_reading(self):
         """Indicate if this data node is ready for reading.
-        
+
         Returns:
-            bool: False if the data is locked for edition or if the data has never been written.
+            bool: False if the data is locked for edit or if the data has never been written.
```
florian-vuillemot:
```suggestion
            bool: False if the data is locked for modification or if the data has never been written.
```

taipy/core/_scheduler/_job_dispatcher.py
jrobinAV: better phrasing, indeed.

taipy/core/data/_data_repository.py
```diff
@@ -118,10 +118,10 @@ def _from_model(self, model: _DataNodeModel):
             id=model.id,
             name=model.name,
             parent_id=model.parent_id,
-            last_edition_date=datetime.fromisoformat(model.last_edition_date) if model.last_edition_date else None,
+            last_edit_date=datetime.fromisoformat(model.last_edit_date) if model.last_edit_date else None,
```
jrobinAV, line 121: What about something like : 
```suggestion
            last_edit_date = datetime.fromisoformat(model.last_edit_date) if model.last_edit_date else datetime.fromisoformat(model.last_edition_date) if model.last_edition_date else None,
```
So we accept to read the last_edition_date field but we don't write it. Using such a mechanism, the old JSON files could be still compatible with the new field name. For that, we need to keep temporarily the 2 names in the dataNodeModel class.

taipy/core/data/_data_repository.py
```diff
@@ -118,10 +118,10 @@ def _from_model(self, model: _DataNodeModel):
...
             job_ids=model.job_ids,
             validity_period=validity_period,
-            edition_in_progress=model.edition_in_progress,
+            edit_in_progress=model.edit_in_progress,
```
jrobinAV, line 124: same here.

jrobinAV (changes requested):
I am thinking that we are breaking the API here, but not only. we are also renaming a field stored in JSON files. That means the user will need to migrate the JSON files (or delete all of them). We can help the user to migrate by accepting old versions.

To accept old versions, we may want to keep in DataModel both names, and in DataNodeRepository, we may want to read the old name if the new name is not set. But of course, when writing the DataModel, we only write the new name. So the data are migrated from the old version to the new one on the fly when they are read. 

What do you think?

taipy/core/data/_data_repository.py
khaitruong922: Since last_edition_date is no longer in the model, model.last_edition_date will throw AttributeError. Do you think we should also store it in the model?

khaitruong922: Is there anyway we can read them without storing in the model?

florian-vuillemot: Can we automatically save the model if it's the old version?

jrobinAV: The principle I propose is to be able to read both old and new models, but we only write the new model.

taipy/core/_repository/_fs_base.py
```diff
@@ -128,9 +128,6 @@ def _get_by_config_and_parent_ids(self, config_id: str, parent_id: Optional[str]
                 return entity
         return None
 
-    def _build_model(self, model_data: Dict) -> ModelType:
-        return self.model.from_dict(model_data)  # type: ignore
-
```
khaitruong922, line 133: This is not used anywhere so I deleted it.

taipy/core/data/_data_model.py
```diff
@@ -45,10 +45,10 @@ def from_dict(data: Dict[str, Any]):
             storage_type=data["storage_type"],
             name=data["name"],
             parent_id=data["parent_id"],
-            last_edition_date=data["last_edition_date"],
+            last_edit_date=data.get("last_edit_date", data.get("last_edition_date")),
             job_ids=data["job_ids"],
             validity_days=data["validity_days"],
             validity_seconds=data["validity_seconds"],
-            edition_in_progress=bool(data["edition_in_progress"]),
+            edit_in_progress=bool(data.get("edit_in_progress", data.get("edition_in_progress", False))),
```
khaitruong922, line 52: @jrobinAV @florian-vuillemot 

I have found a way that to load the deprecated properties from the JSON file without storing them in the model.  Basically, the new properties just look for both versions and prefer the new one. Eventually, we will write the new properties only. What do you think?

jrobinAV (dismissed):

jrobinAV (approved):

toan-quach (approved):

PR #194, opened 2022-05-10
Feature - Create manager factory

khaitruong922:
This is the baseline for us to work on the overriding manager tickets.

The factory will check if taipy-enterprise is installed, then return the authorized class. Otherwise, it will return the taipy-core class.

### DIFF
--- a/taipy/core/_manager/_manager_factory.py
+++ b/taipy/core/_manager/_manager_factory.py
@@ -29,50 +29,44 @@ class _ManagerFactory:
     def _scenario_manager(cls) -> Type[_ScenarioManager]:
         if cls._has_enterprise():
             return _load_fct(
-                cls._TAIPY_ENTERPRISE_CORE_MODULE + ".scenario._authorized_scenario_manager",
-                "_AuthorizedScenarioManager",
+                cls._TAIPY_ENTERPRISE_CORE_MODULE + ".scenario._scenario_manager",
+                "_ScenarioManager",
             )  # type: ignore
         return _ScenarioManager
 
     @classmethod
     def _data_manager(cls) -> Type[_DataManager]:
         if cls._has_enterprise():
-            return _load_fct(
-                cls._TAIPY_ENTERPRISE_CORE_MODULE + ".data._authorized_data_manager", "_AuthorizedDataManager"
-            )  # type: ignore
+            return _load_fct(cls._TAIPY_ENTERPRISE_CORE_MODULE + ".data._data_manager", "_DataManager")  # type: ignore
         return _DataManager
 
     @classmethod
     def _cycle_manager(cls) -> Type[_CycleManager]:
         if cls._has_enterprise():
             return _load_fct(
-                cls._TAIPY_ENTERPRISE_CORE_MODULE + ".cycle._authorized_cycle_manager", "_AuthorizedCycleManager"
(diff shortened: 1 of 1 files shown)


taipy/core/_manager/_manager_factory.py
```diff
@@ -0,0 +1,80 @@
...
+
+class _ManagerFactory:
+    _TAIPY_ENTERPRISE_MODULE = "taipy.enterprise"
+    _TAIPY_ENTERPRISE_CORE_MODULE = _TAIPY_ENTERPRISE_MODULE + ".core"
+
+    @classmethod
+    def _scenario_manager(cls) -> Type[_ScenarioManager]:
+        if cls._has_enterprise():
+            return _load_fct(
+                cls._TAIPY_ENTERPRISE_CORE_MODULE + ".scenario._authorized_scenario_manager",
```
jrobinAV: I would name the module in the enterprise version _scenarion_manager and not _authorized_scenario_manager. The purpose is to hide from the community version what is done in the enterprise version, and how it is implemented.
```suggestion
                cls._TAIPY_ENTERPRISE_CORE_MODULE + ".scenario._scenario_manager",
```

taipy/core/_manager/_manager_factory.py
```diff
@@ -0,0 +1,80 @@
...
+                "_AuthorizedScenarioManager",
```
jrobinAV:
```suggestion
                "_ScenarioManager",
```

jrobinAV (dismissed):

taipy/core/_manager/_manager_factory.py
khaitruong922: Isn't it counterintuitive to not let the community know what we provide in the enterprise version?

toan-quach: i think the idea is to keep people from copying our product :D

jrobinAV: exactly

taipy/core/_manager/_manager_factory.py
```diff
@@ -0,0 +1,74 @@
...
+from taipy.core.task._task_manager import _TaskManager
+
+
+class _ManagerFactory:
+    _TAIPY_ENTERPRISE_MODULE = "taipy.enterprise"
+    _TAIPY_ENTERPRISE_CORE_MODULE = _TAIPY_ENTERPRISE_MODULE + ".core"
+
+    @classmethod
+    def _scenario_manager(cls) -> Type[_ScenarioManager]:
+        if cls._has_enterprise():
```
toan-quach, line 30: @jrobinAV should we also include if the config mode stated to use the enterprise module?

khaitruong922: That would require a new config value in the global app. What do you think? @jrobinAV  @toan-quach

toan-quach: yep, the reason I'm asking is in the use case where enterprise users dont want to have authorisation, how should we approach it 😄

FabienLelaquais: I guess, if the user WANTS to use authentication and the rest of it, a configuration of how to access the authentication system will have to be provided.
If no such config is provided.... then there will be no authentication...

jrobinAV: Yes. I believe we will need to have a test on the config at a point. This is not blocking the PR anyway.

jrobinAV (approved):

PR #196, opened 2022-05-12
Feature/#190 - Create Scenario Manager from factory

khaitruong922:
- Move _manager.py into _manager folder.
- Create Scenario Manager from manager factory.
- Refactor Scenario API to be consistent with using `import taipy.core as tp` since we are building ScenarioManager from factory in taipy.py.


toan-quach (commented):
I'm quite confused as to why we're moving _manager.py to folder _manager and not just keep it at common?

jrobinAV (changes requested):
I don't like the new code design, actually.
I prefer to have a domain-based code design, meaning we split the code by domain (datanode, task, scenario...) instead of a layer-based code design (controllers, managers, repositories, model, ...).

So I would keep the _scenario_manager in the scenario package.
I would also split the _ManagerFactory into multiple classes ex: 
having a _ScenarioManagerFactory in the scenario package.

khaitruong922:
@jrobinAV So we can split them into seperate factory, and move the logic of checking if enterprise is used into the Config. What do you think?

FabienLelaquais (dismissed):
👌

jrobinAV:
> @jrobinAV So we can split them into seperate factory, and move the logic of checking if enterprise is used into the Config. What do you think?

I don't think this should be in the config. It is not really linked to the back-end config, but to the taipy version you have. 

We can have the logic duplicated in each factory. In the end it is just one line of code: 

```
util.find_spec(cls._TAIPY_ENTERPRISE_MODULE) is not None
```

Otherwise, if you really want to factorize this piece of code, we can have a static method in module common._utils.py. Up to you.

khaitruong922:
> I'm quite confused as to why we're moving _manager.py to folder _manager and not just keep it at common?

The common should contain util functions or enums so putting the _Manager class feel kinda out of place. Since now we are having _manager directory, I think it would be a better place for _Manager class.

taipy/core/scenario/_scenario_manager_factory.py
```diff
@@ -0,0 +1,24 @@
...
+from taipy.core._manager._manager_factory import _ManagerFactory
+from taipy.core.common._utils import _load_fct
+from taipy.core.scenario._scenario_manager import _ScenarioManager
+
+
+class _ScenarioManagerFactory(_ManagerFactory):
+    @classmethod
+    def _build_manager(cls) -> Type[_ScenarioManager]:
+        if cls._using_enterprise():
+            return _load_fct(cls._TAIPY_ENTERPRISE_CORE_MODULE + ".scenario._scenario_manager")  # type: ignore
```
jrobinAV: isn't it missing an argument to the function _load_fct ?

jrobinAV (commented):
I found these 4 direct usages of _ScenarioManager. Can you check if this is an oversight?
- taipy.core._manager._manager._Manager._delete_entities_of_multiple_types
- taipy.core.common._reload._get_manager
- taipy.core.cycle._cycle_manager._CycleManager._get_owned_entities
- taipy.core.scenario.scenario.Scenario.__setstate__

jrobinAV (approved):

PR #199, opened 2022-05-13
feature/#28 added src folder

toan-quach:


jrobinAV:
It seems very complicated for little added value. LEt s discuss this tomorrow. I may miss some points.

toan-quach:
Obsolete

Issue #175, opened 2022-04-28
Do not save SECRET in storage_folder


florian-vuillemot:
Taipy can deal with secrets through the "ENV" syntax in Config but can't with Entities.

So, if you create an Entity from a Config using the "ENV" syntax, you will not be able to see the secret in the config file, but you will be able to see it in the entity file.

```

from taipy import Config
import taipy as tp
import os

os.environ['foo'] = 'bar'
Config.configure_global_app(storage_folder='.data')
data = Config.configure_data_node(id="data", properties={"secret": "ENV['foo']"})

...

print(f"{os.environ['foo']} is present in './data/data_node/{scenario.data.id}'")
```

PR #203, opened 2022-05-16
Fix/#175 do not save secrets in entities

jrobinAV:
I implemented a fix for taipy-core #175.

From now, the Config._applied_config object still does contain the compiled config, but the environment values are not interpreted. For instance, the config field contains values like 'ENV[NAME]' (before, the value was interpreted and replaced by the value of the corresponding environment variable).
I then needed to ensure that every time a config field is accessed, it is interpreted on the fly through the method: _tpl._replace_templates. I created getters/setters properties for that purpose.

Here is the expected behavior of the usage of a config object:
```
def test_data_node_with_env_variable_value():
    with mock.patch.dict(os.environ, {"FOO": "pickle", "BAR": "baz"}):
        Config.configure_data_node("data_node", storage_type="ENV[FOO]", prop="ENV[BAR]")
        assert Config.data_nodes["data_node"].prop == "baz"
        assert Config.data_nodes["data_node"].properties["prop"] == "baz"
        assert Config.data_nodes["data_node"]._properties["prop"] == "ENV[BAR]"
        assert Config.data_nodes["data_node"].storage_type == "pickle"
        assert Config.data_nodes["data_node"]._storage_type == "ENV[FOO]"
```

First consequence: When exporting the applied config, we do not export the env variable values, but the templates. And so secrets are not serialized.

Second consequence: When creating an entity, the properties of the config are copied before they got interpreted and replaced. When serializing the entity through a repository, we do not serialize interpreted values. And so secrets are not serialized.
I then needed to ensure that every time an entity field is accessed, it is interpreted on the fly through the method: _tpl._replace_templates. I created getters/setters properties for that purpose.

Here is the behavior of the usage of an entity object:
```
def test_data_node_with_env_variable_value_not_serialized(self):
    with mock.patch.dict(os.environ, {"FOO": "bar"}):
        assert self.data_node._properties["prop"] == "ENV[FOO]"
        assert self.data_node.properties["prop"] == "bar"
        assert self.data_node.prop == "bar"
```

### DIFF
--- a/taipy/core/config/_config_template_handler.py
+++ b/taipy/core/config/_config_template_handler.py
@@ -12,6 +12,7 @@
 import os
 import re
 
+from taipy.core.common._properties import _Properties
 from taipy.core.common.frequency import Frequency
 from taipy.core.common.scope import Scope
 from taipy.core.exceptions.exceptions import InconsistentEnvVariableError, MissingEnvVariableError
@@ -29,6 +30,8 @@ class _ConfigTemplateHandler:
         if isinstance(template, list):
             return [cls._replace_template(item, type, required, default) for item in template]
         if isinstance(template, dict):
+            return {str(k): cls._replace_template(v, type, required, default) for k, v in template.items()}
+        if isinstance(template, _Properties):
             return {str(k): cls._replace_template(v, type, required, default) for k, v in template.items()}
         return cls._replace_template(template, type, required, default)
 
--- a/taipy/core/config/data_node_config.py
+++ b/taipy/core/config/data_node_config.py
@@ -51,17 +51,33 @@ class DataNodeConfig:
 
     def __init__(self, id: str, storage_type: str = None, scope: Scope = None, **properties):
         self.id = _validate_id(id)
-        self.storage_type = storage_type
-        self.scope = scope
+        self._storage_type = storage_type
+        self._scope = scope
         self.properties = properties
         if self.properties.get(self._IS_CACHEABLE_KEY) is None:
             self.properties[self._IS_CACHEABLE_KEY] = self._DEFAULT_IS_CACHEABLE_VALUE
 
     def __getattr__(self, item: str) -> Optional[Any]:
-        return self.properties.get(item)
+        return _tpl._replace_templates(self.properties.get(item))
 
     def __copy__(self):
-        return DataNodeConfig(self.id, self.storage_type, self.scope, **copy(self.properties))
+        return DataNodeConfig(self.id, self._storage_type, self._scope, **copy(self.properties))
+
+    @property
+    def storage_type(self):
+        return _tpl._replace_templates(self._storage_type)
+
+    @storage_type.setter  # type: ignore
+    def storage_type(self, val):
+        self._storage_type = val
+
+    @property
+    def scope(self):
+        return _tpl._replace_templates(self._scope)
+
+    @scope.setter  # type: ignore
+    def scope(self, val):
+        self._scope = val
 
     @classmethod
     def default_config(cls, id):
@@ -69,10 +85,10 @@ class DataNodeConfig:
 
     def _to_dict(self):
         as_dict = {}
-        if self.storage_type is not None:
-            as_dict[self._STORAGE_TYPE_KEY] = self.storage_type
-        if self.scope is not None:
-            as_dict[self._SCOPE_KEY] = self.scope
+        if self._storage_type is not None:
+            as_dict[self._STORAGE_TYPE_KEY] = self._storage_type
+        if self._scope is not None:
+            as_dict[self._SCOPE_KEY] = self._scope
         as_dict.update(self.properties)
         return as_dict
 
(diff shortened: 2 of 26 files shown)


khaitruong922 (dismissed):

jrobinAV:
After better thinking, I believe the following is not great:
data_node.properties["prop"] != data_node.prop

taipy/core/config/_config_template_handler.py
```diff
@@ -30,6 +31,8 @@ def _replace_templates(cls, template, type=str, required=True, default=None):
             return [cls._replace_template(item, type, required, default) for item in template]
         if isinstance(template, dict):
             return {str(k): cls._replace_template(v, type, required, default) for k, v in template.items()}
+        if isinstance(template, _Properties):
+            return {str(k): cls._replace_template(v, type, required, default) for k, v in template.items()}
```
florian-vuillemot, line 35:
```suggestion
        if isinstance(template, dict) or isinstance(template, _Properties):
            return {str(k): cls._replace_template(v, type, required, default) for k, v in template.items()}
```

taipy/core/config/data_node_config.py
```diff
@@ -51,49 +51,72 @@ class DataNodeConfig:
...
+    @storage_type.setter  # type: ignore
+    def storage_type(self, val):
+        self._storage_type = val
+
+    @property
+    def properties(self):
+        res = {}
+        for k, v in self._properties.items():
+            res[k] = _tpl._replace_templates(v)
+        return res
```
florian-vuillemot:
```suggestion
        return {k: _tpl._replace_templates(v) for k, v in self._properties.items()}
```

taipy/core/config/task_config.py
```diff
@@ -58,10 +58,21 @@ def __init__(
...
     def __copy__(self):
-        return TaskConfig(self.id, copy(self._inputs), self.function, copy(self._outputs), **copy(self.properties))
+        return TaskConfig(self.id, copy(self._inputs), self.function, copy(self._outputs), **copy(self._properties))
+
+    @property
+    def properties(self):
+        res = {}
+        for k, v in self._properties.items():
+            res[k] = _tpl._replace_templates(v)
+        return res
```
florian-vuillemot:
```suggestion
        return {k: _tpl._replace_templates(v) for k, v in self._properties.items()}
```

tests/core/data/test_data_manager.py
```diff
@@ -36,6 +36,7 @@ def test_create_data_node_and_modify_properties_does_not_modify_config(self):
         assert dn_config.properties.get("foo") == "bar"
         assert dn_config.properties.get("baz") is None
 
+        # dn._properties["baz"] = "qux"
```
florian-vuillemot: ?

florian-vuillemot (dismissed):

taipy/core/config/global_app_config.py
```diff
@@ -47,51 +47,83 @@ def __init__(
...
+    @clean_entities_enabled.setter  # type: ignore
+    def clean_entities_enabled(self, val):
+        self._clean_entities_enabled = val
+
+    @property
+    def properties(self):
+        res = {}
+        for k, v in self._properties.items():
+            res[k] = _tpl._replace_templates(v)
+        return res
```
jrobinAV:
```suggestion
return {k: _tpl._replace_templates(v) for k, v in self._properties.items()}
```

taipy/core/config/pipeline_config.py
```diff
@@ -31,17 +31,28 @@ class PipelineConfig:
...
     def __copy__(self):
-        return PipelineConfig(self.id, copy(self._tasks), **copy(self.properties))
+        return PipelineConfig(self.id, copy(self._tasks), **copy(self._properties))
+
+    @property
+    def properties(self):
+        res = {}
+        for k, v in self._properties.items():
+            res[k] = _tpl._replace_templates(v)
+        return res
```
jrobinAV:
```suggestion
return {k: _tpl._replace_templates(v) for k, v in self._properties.items()}
```

taipy/core/config/scenario_config.py
```diff
@@ -59,11 +59,22 @@ def __init__(
...
         comp = None if self.comparators is None else self.comparators
-        return ScenarioConfig(self.id, copy(self._pipelines), self.frequency, copy(comp), **copy(self.properties))
+        return ScenarioConfig(self.id, copy(self._pipelines), self.frequency, copy(comp), **copy(self._properties))
+
+    @property
+    def properties(self):
+        res = {}
+        for k, v in self._properties.items():
+            res[k] = _tpl._replace_templates(v)
+        return res
```
jrobinAV:
```suggestion
return {k: _tpl._replace_templates(v) for k, v in self._properties.items()}
```

tests/core/data/test_data_manager.py
jrobinAV:
```suggestion
```

jrobinAV:
> After better thinking, I believe the following is not great: data_node.properties["prop"] != data_node.prop

This is fixed!

PR #204, opened 2022-05-19
Add max version in setup.py

florian-vuillemot:
- Setting the maximum version of packages will limit the risk of API changes that can impact our product.
- Test package installation on Windows, Linux and Mac
- Test the package installation each day

https://app.zenhub.com/workspaces/taipy-613ef328affde6000ff77fb3/issues/avaiga/taipy/38

### DIFF
--- a/.github/workflows/setuptools.yml
+++ b/.github/workflows/setuptools.yml
@@ -54,7 +54,7 @@ jobs:
           echo "[print(k) for k in extras_require.keys()]" >> setup_packages.py
           python setup_packages.py | xargs -I% bash -c "pip install .[%]"
 
-          rm -r taipy  # Removal of Taipy code to ensure the use of the installed package.
+          rm -r taipy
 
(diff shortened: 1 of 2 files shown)


.github/workflows/setuptools.yml
```diff
@@ -5,14 +5,16 @@ on:
     branches: [ develop ]
   pull_request:
     branches: [ develop ]
+  schedule:
+    - cron: "0 8 * * *"
```
jrobinAV, line 9: Every day at 8AM? 

OK. Why, not.

jrobinAV (approved):

.github/workflows/setuptools.yml
florian-vuillemot: One day a week is enough? On Monday?

jrobinAV: Don't know. Let's try every day, and adjust later if needed.

Issue #184, opened 2022-05-09
DataNodeConfig fields should be prefixed by `default_` when the field should/could be overridden in the entity.


jrobinAV:
Several fields in data node configs such as file_path should be prefixed by `default` to let know the user that if he/she instantiates 2 entities from the config, the property in the entities should/could be changed.

An alternative would be to find a method like in Pickle data node to have a unique name.

jrobinAV:
@khaitruong922, Since the authentication topic is blocked by the Authorization context manager, you should get back to this ticket.

PR #209, opened 2022-05-26
Feature #184 - rename path to default_path and make path settable

khaitruong922:


taipy/core/data/pickle.py
```diff
@@ -47,8 +49,10 @@ class PickleDataNode(DataNode):
     """
 
     __STORAGE_TYPE = "pickle"
-    __PICKLE_FILE_NAME = "path"
+    __PICKLE_PATH_KEY = "path"
+    __PICKLE_DEFAULT_PATH_KEY = "default_path"
     __DEFAULT_DATA_VALUE = "default_data"
+    _MANAGER_NAME: str = "data"
```
jrobinAV: It is already in the parent class DataNode.
```suggestion
```

taipy/core/data/pickle.py
```diff
@@ -79,32 +83,40 @@ def __init__(
             edit_in_progress,
             **properties,
         )
-        self.__pickle_file_path = self.__build_path()
```
jrobinAV, line 82: I would use the name __pickle_path. At this point, it is not a default path anymore. It is the real path of the data node.

taipy/core/data/pickle.py
```diff
@@ -79,32 +83,40 @@ def __init__(
...
-    def path(self) -> str:
-        return self.__pickle_file_path
+    @property  # type: ignore
+    @_self_reload(_MANAGER_NAME)
+    def path(self) -> Any:
+        return self._pickle_default_path
+
+    @path.setter  # type: ignore
+    def path(self, value: Any):
+        self.properties[self.__PICKLE_DEFAULT_PATH_KEY] = value
```
jrobinAV, line 103: You don't need to change the default path value in properties, but the actual path. So if needed, you can retrieve what was the default path after you changed the path.
```suggestion
        self._pickle_path = value
```

taipy/core/data/pickle.py
```diff
@@ -79,32 +83,40 @@ def __init__(
             edit_in_progress,
             **properties,
         )
-        self.__pickle_file_path = self.__build_path()
-        if not self._last_edit_date and os.path.exists(self.__pickle_file_path):
+        self._pickle_default_path = self.__build_path()
+        if not self._last_edit_date and os.path.exists(self._pickle_default_path):
```
jrobinAV:
```suggestion
        self._pickle_path = self.__build_path()
        if not self._last_edit_date and os.path.exists(self._pickle_path):
```

taipy/core/data/pickle.py
```diff
@@ -79,32 +83,40 @@ def __init__(
...
             self.unlock_edit()
-        if default_value is not None and not os.path.exists(self.__pickle_file_path):
+        if default_value is not None and not os.path.exists(self._pickle_default_path):
```
jrobinAV:
```suggestion
        if default_value is not None and not os.path.exists(self._pickle_path):
```

taipy/core/data/pickle.py
```diff
@@ -79,32 +83,40 @@ def __init__(
...
     def storage_type(cls) -> str:
         return cls.__STORAGE_TYPE
 
-    @property
-    def path(self) -> str:
-        return self.__pickle_file_path
+    @property  # type: ignore
+    @_self_reload(_MANAGER_NAME)
+    def path(self) -> Any:
+        return self._pickle_default_path
```
jrobinAV:
```suggestion
        return self._pickle_path
```

taipy/core/data/pickle.py
```diff
@@ -79,32 +83,40 @@ def __init__(
...
+        self.properties[self.__PICKLE_DEFAULT_PATH_KEY] = value
 
     @property
     def is_file_generated(self) -> bool:
-        return self.__PICKLE_FILE_NAME not in self.properties
+        return self.__PICKLE_DEFAULT_PATH_KEY not in self.properties and self.__PICKLE_PATH_KEY not in self.properties
 
     def _read(self):
-        return pickle.load(open(self.__pickle_file_path, "rb"))
+        return pickle.load(open(self._pickle_default_path, "rb"))
```
jrobinAV:
```suggestion
        return pickle.load(open(self._pickle_path, "rb"))
```

taipy/core/data/pickle.py
```diff
@@ -79,32 +83,40 @@ def __init__(
...
 
     def _write(self, data):
-        pickle.dump(data, open(self.__pickle_file_path, "wb"))
+        pickle.dump(data, open(self._pickle_default_path, "wb"))
```
jrobinAV:
```suggestion
        pickle.dump(data, open(self._pickle_path, "wb"))
```

jrobinAV:
I also pushed a fix on the method is_file_generated. Since the logic is slightly different now.
@khaitruong922 please review my latest commit.

taipy/core/data/pickle.py
khaitruong922: I tried this initially but it does not work because _pickle_path is a computed property, so we should set the properties to trigger the saving into JSON file.

taipy/core/data/pickle.py
khaitruong922: We cannot reference the constant in the parent class without using self or cls. I will use DataNode._MANAGER_NAME to reference the manager name.

taipy/core/data/pickle.py
jrobinAV: you can trigger the saving using the setter decorator : 
@_self_setter(_MANAGER_NAME)

khaitruong922: I have tried that method but it does not work. I think the problem is that the _pickle_path is a custom property itself. I think there is no need to retrieve the default path from the entity itself, which can be retrieved from the config.

tests/core/data/test_csv_data_node.py
```diff
@@ -45,7 +45,7 @@ def test_create(self):
         assert dn.last_edition_date is None
         assert dn.job_ids == []
         assert not dn.is_ready_for_reading
-        assert dn.path == path
+        assert dn._path == path
```
jrobinAV: We want to test that path still returns the right value.

jrobinAV (changes requested):
The ticket is not complete. The work on the entities is done but the work on config is not done.

1- We want in DataNodeConfig to stop exposing the field `path` but we want to expose `default_path`. Something like:
```
config = Config.configure_data_node(name="foo", default_path="foo.csv")
assert config.default_path == "foo.csv"
config.default_path = "bar.csv"
assert config.default_path == "bar.csv"
```
2- When building a data node entity from the DataNodeConfig, we still want to expose the `path` being built from the `config.default_path`. Keeping the default path is useless. We want to be able to do the following:
```
config = Config.configure_data_node(name="foo", default_path="foo.csv")
data_node = _DataManager._get_or_create(config)
assert data_node.path == "foo.csv"
data_node.path == "bar.csv"
assert data_node.path == "bar.csv"
```
3- We want to keep compatibility with previous versions. Old configs and old entities should still work.

khaitruong922:
I think they are all done. Let me add some tests to verify that.

khaitruong922:
```
config.default_path = "bar.csv"
````

This does not work because we have not implement __setattr__ for data node config. Is it necessary to have it?
No it is not. thx

jrobinAV (changes requested):
The method signaturs of `Config.configure_csv_data_node` and `Config.configure_excel_data_node` must be updated.

jrobinAV (approved):

PR #212, opened 2022-05-26
Feature/#210 - Add args kwargs to managers and Taipy API

khaitruong922:


khaitruong922:
The _CycleManager._create method takes in properties as kwargs, and we do not want to save the `token` inside the cycle properties. I would propose to make the `properties` a parameter so that we can put **kwargs. This method is not exposed to end user so we can refactor it and propagate to taipy-enterprise.

What do you think? @jrobinAV @toan-quach

khaitruong922:
There is a big issue I found when working on this ticket is that a class method can call other class methods, which triggers a lot of unnecessary auth checks. For example, if you hard delete a scenario, it will also check the auth when deleting each pipeline, task and data node of that scenario, which is a performance issue. I'm thinking of passing a value to skip the auth check when you call within a class method. 

Do you have any idea? @toan-quach @jrobinAV

jrobinAV:
> There is a big issue I found when working on this ticket is that a class method can call other class methods, which triggers a lot of unnecessary auth checks. For example, if you hard delete a scenario, it will also check the auth when deleting each pipeline, task and data node of that scenario, which is a performance issue. I'm thinking of passing a value to skip the auth check when you call within a class method.
> 
> Do you have any idea? @toan-quach @jrobinAV

We should put this in another ticket. As a first step, this implementation is ok.

taipy/core/cycle/_cycle_manager.py
```diff
@@ -18,18 +18,25 @@
...
+        cls,
+        frequency: Frequency,
+        name: str = None,
+        creation_date: datetime = None,
+        display_name=None,
+        *args,
+        **properties
     ):
         creation_date = creation_date if creation_date else datetime.now()
         start_date = _CycleManager._get_start_date_of_cycle(frequency, creation_date)
```
jrobinAV:
```suggestion
        start_date = _CycleManager._get_start_date_of_cycle(frequency, creation_date, *args, **kwargs)
        end_date = _CycleManager._get_end_date_of_cycle(frequency, start_date, *args, **kwargs)
```

jrobinAV: This implementation does not work. args and kwargs must still contain the enterprise property at this point !!!!

taipy/core/data/_data_manager.py
```diff
@@ -34,23 +35,25 @@ def _get_or_create(
...
-        return cls._create_and_set(data_node_config, parent_id)
+        return cls._create_and_set(data_node_config, parent_id, *args, **kwargs)
 
     @classmethod
-    def _create_and_set(cls, data_node_config: DataNodeConfig, parent_id: Optional[str]) -> DataNode:
-        data_node = cls.__create(data_node_config, parent_id)
-        cls._set(data_node)
+    def _create_and_set(cls, data_node_config: DataNodeConfig, parent_id: Optional[str], *args, **kwargs) -> DataNode:
+        data_node = cls.__create(data_node_config, parent_id, *args, **kwargs)
+        cls._set(data_node, parent_id, *args, **kwargs)
```
jrobinAV: Why is the parent_id variable in the set method ???

taipy/core/data/_data_manager.py
```diff
@@ -65,29 +68,29 @@ def __create(cls, data_node_config: DataNodeConfig, parent_id: Optional[str]) ->
...
-    def _clean_pickle_files(cls, data_nodes: Iterable[Union[DataNode, DataNodeId]]):
+    def _clean_pickle_files(cls, data_nodes: Iterable[Union[DataNode, DataNodeId]], *args, **kwargs):
         for data_node in data_nodes:
-            cls._clean_pickle_file(data_node)
+            cls._clean_pickle_file(data_node, *args, **kwargs)
 
     @classmethod
-    def _delete(cls, data_node_id: DataNodeId):  # type:ignore
+    def _delete(cls, data_node_id: DataNodeId, *args, **kwargs):  # type:ignore
         cls._clean_pickle_file(data_node_id)
```
jrobinAV:
```suggestion
        cls._clean_pickle_file(data_node_id, *args, **kwargs)
```

taipy/core/pipeline/_pipeline_manager.py
```diff
@@ -17,51 +17,58 @@
...
+        cls._set(pipeline, *args, **kwargs)
 
     @classmethod
-    def _get_or_create(cls, pipeline_config: PipelineConfig, scenario_id: Optional[ScenarioId] = None) -> Pipeline:
+    def _get_or_create(
+        cls, pipeline_config: PipelineConfig, scenario_id: Optional[ScenarioId] = None, *args, **kwargs
+    ) -> Pipeline:
         pipeline_id = Pipeline._new_id(pipeline_config.id)
         task_manager = _TaskManagerFactory._build_manager()
         tasks = [
```
jrobinAV, line 73: missing args, kwargs

taipy/core/pipeline/_pipeline_manager.py
```diff
@@ -74,12 +81,17 @@ def _get_or_create(cls, pipeline_config: PipelineConfig, scenario_id: Optional[S
...
-        cls, pipeline: Union[PipelineId, Pipeline], callbacks: Optional[List[Callable]] = None, force: bool = False
+        cls,
+        pipeline: Union[PipelineId, Pipeline],
+        callbacks: Optional[List[Callable]] = None,
+        force: bool = False,
+        *args,
+        **kwargs,
     ):
         callbacks = callbacks or []
         pipeline_id = pipeline.id if isinstance(pipeline, Pipeline) else pipeline
```
jrobinAV, line 97: missing args, kwargs

taipy/core/pipeline/_pipeline_manager.py
```diff
@@ -96,22 +108,22 @@ def __get_status_notifier_callbacks(pipeline: Pipeline) -> List:
         return [partial(c, pipeline) for c in pipeline.subscribers]
 
     @classmethod
-    def _hard_delete(cls, pipeline_id: PipelineId):
+    def _hard_delete(cls, pipeline_id: PipelineId, *args, **kwargs):
         pipeline = cls._get(pipeline_id)
```
jrobinAV:
```suggestion
        pipeline = cls._get(pipeline_id, *args, **kwargs)
```

taipy/core/scenario/_scenario_manager.py
```diff
@@ -44,53 +45,54 @@ class _ScenarioManager(_Manager[Scenario]):
...
+    ):
         if scenario is None:
-            scenarios = cls._get_all()
+            scenarios = cls._get_all(*args, **kwargs)
             for scn in scenarios:
-                cls.__remove_subscriber(callback, scn)
+                cls.__remove_subscriber(callback, scn, *args, **kwargs)
             return
 
         cls.__remove_subscriber(callback, scenario)
```
jrobinAV:
```suggestion
        cls.__remove_subscriber(callback, scenario, *args, **kwargs)
```

taipy/core/scenario/_scenario_manager.py
```diff
@@ -103,101 +105,101 @@ def _create(
...
             raise UnauthorizedTagError(f"Tag `{tag}` not authorized by scenario configuration `{scenario.config_id}`")
         if scenario.cycle:
-            old_tagged_scenario = cls._get_by_tag(scenario.cycle, tag)
+            old_tagged_scenario = cls._get_by_tag(scenario.cycle, tag, *args, **kwargs)
             if old_tagged_scenario:
                 old_tagged_scenario.remove_tag(tag)
-                cls._set(old_tagged_scenario)
+                cls._set(old_tagged_scenario, *args, **kwargs)
         scenario._add_tag(tag)
         cls._set(scenario)
```
jrobinAV:
```suggestion
        cls._set(scenario, *args, **kwargs)
```

taipy/core/scenario/_scenario_manager.py
```diff
@@ -103,101 +105,101 @@ def _create(
...
-    def _untag(cls, scenario: Scenario, tag: str):
+    def _untag(cls, scenario: Scenario, tag: str, *args, **kwargs):
         scenario._remove_tag(tag)
-        cls._set(scenario)
+        cls._set(scenario, *args, **kwargs)
 
     @classmethod
-    def _delete(cls, scenario_id: ScenarioId, **kwargs):  # type: ignore
+    def _delete(cls, scenario_id: ScenarioId, *args, **kwargs):  # type: ignore
         if cls._get(scenario_id).is_primary:
```
jrobinAV:
```suggestion
        if cls._get(scenario_id, *args, **kwargs).is_primary:
```

taipy/core/scenario/_scenario_manager.py
```diff
@@ -230,16 +232,16 @@ def __get_config(scenario: Scenario):
         return Config.scenarios.get(scenario.config_id, None)
 
     @classmethod
-    def _hard_delete(cls, scenario_id: ScenarioId):
+    def _hard_delete(cls, scenario_id: ScenarioId, *args, **kwargs):
         scenario = cls._get(scenario_id)
```
jrobinAV:
```suggestion
        scenario = cls._get(scenario_id, *args, **kwargs)
```

taipy/core/task/_task_manager.py
```diff
@@ -68,31 +73,33 @@ def _get_or_create(
         return task
```
jrobinAV, line 72: missing , *args, **kwargs

taipy/core/task/_task_manager.py
```diff
@@ -68,31 +73,33 @@ def _get_or_create(
...
+    def __save_data_nodes(cls, data_nodes, *args, **kwargs):
         data_manager = _DataManagerFactory._build_manager()
         for i in data_nodes:
-            data_manager._set(i)
+            data_manager._set(i, *args, **kwargs)
 
     @classmethod
-    def _hard_delete(cls, task_id: TaskId):
+    def _hard_delete(cls, task_id: TaskId, *args, **kwargs):
         task = cls._get(task_id)
```
jrobinAV:
```suggestion
        task = cls._get(task_id, *args, **kwargs)
```

taipy/core/task/_task_manager.py
```diff
@@ -68,31 +73,33 @@ def _get_or_create(
...
-    def _submit(cls, task: Union[TaskId, Task], callbacks: Optional[List[Callable]] = None, force: bool = False):
+    def _submit(
+        cls, task: Union[TaskId, Task], callbacks: Optional[List[Callable]] = None, force: bool = False, *args, **kwargs
+    ):
         task_id = task.id if isinstance(task, Task) else task
-        task = cls._get(task_id)
+        task = cls._get(task_id, *args, **kwargs)
         if task is None:
             raise NonExistingTask(task_id)
         cls._scheduler().submit_task(task, callbacks, force)
```
jrobinAV, line 104: This should be protected as well. Must create another ticket for that !

taipy/core/pipeline/_pipeline_manager.py
jrobinAV: Scheduler.submit should recieve *args, **kwargs. New ticket !

jrobinAV (dismissed):
This implementation cannot work. 
I proposed something. Please review it.

Plus I added some missing parameters a bit everywhere.

taipy/core/cycle/_cycle_manager.py
```diff
@@ -24,37 +24,55 @@
...
             frequency, properties, creation_date=creation_date, start_date=start_date, end_date=end_date, name=name
         )
-        cls._set(cycle)
+        cls._set(cycle, *args, **properties)
         return cycle
 
+    @classmethod
+    def _clean_properties(cls, *args, **kwargs):
+        pass
+
```
khaitruong922, line 54: Putting this method without implementation is quite ambigious. How about we defined all the allowed properties for a cycle then filter it, instead of omitting the `token`? What do you think?

jrobinAV: We have implemented it as a dictionary so the user can have custom properties. Let's consider for example that the user wants to add a description, a creator name, or a list of user comments, ...

taipy/core/task/_task_manager.py
jrobinAV: https://github.com/Avaiga/taipy-enterprise/issues/33

jrobinAV (approved):

PR #223, opened 2022-06-08
Perf improvement 

florian-vuillemot:


taipy/core/_repository/_fs_base.py
```diff
@@ -90,29 +90,44 @@ def __init__(self, model: Type[ModelType], dir_name: str):
...
-    @property
-    def _directory(self) -> pathlib.Path:
-        return self.dir_path
-
     def load(self, model_id: str) -> Entity:
-        return self.__to_entity(self.__get_model_filepath(model_id))
+        try:
+            return self.__to_entity(self.__get_model_filepath(model_id))
+        except FileNotFoundError:
+            raise ModelNotFound(str(self.dir_path), model_id)
```
toan-quach, line 99: since we're calling to_entity, I think EntityNotFound is a better name

taipy/core/_repository/_fs_base.py
```diff
@@ -121,30 +136,62 @@ def _delete_many(self, model_ids: Iterable[str]):
...
-            raise ModelNotFound(str(self._directory), model_id)
+    def __get_model_filepath(self, model_id) -> pathlib.Path:
+        return self.dir_path / f"{model_id}.json"
 
-        return filepath
+    def __to_entity(self, filepath, by=None):
+        with open(filepath, "r") as f:
+            d = f.read()
+        if by and by not in d:
+            return None
```
toan-quach: is there a better default value or maybe we can also raise an InvalidModel error

taipy/core/_repository/_fs_base.py
```diff
@@ -121,30 +136,62 @@ def _delete_many(self, model_ids: Iterable[str]):
...
+        data = json.loads(d, cls=_CustomDecoder)
+        model = self.model.from_dict(data)  # type: ignore
+        return self._from_model(model)
 
-    def __to_entity(self, filepath):
+    def __to_entity_by(self, filepath, by):
```
toan-quach: the naming is confusing with __to_entity, can we combine both function? I think __to_entities would be better a name

taipy/core/scenario/scenario.py
```diff
@@ -168,12 +168,14 @@ def __getattr__(self, attribute_name):
         protected_attribute_name = _validate_id(attribute_name)
         if protected_attribute_name in self._properties:
             return _tpl._replace_templates(self._properties[protected_attribute_name])
-        if protected_attribute_name in self.pipelines:
-            return self.pipelines[protected_attribute_name]
-        for pipeline in self.pipelines.values():
-            if protected_attribute_name in pipeline.tasks:
-                return pipeline.tasks[protected_attribute_name]
-            for task in pipeline.tasks.values():
+        pipelines = self.pipelines
```
toan-quach, line 174: this makes so much sense!

toan-quach (dismissed):
LGTM! just some minor remarks :D

taipy/core/_repository/_fs_base.py
florian-vuillemot: Good point. This is the historical implementation, but we could change it.

florian-vuillemot: @jrobinAV  what do you think?

taipy/core/_repository/_fs_base.py
florian-vuillemot: In our case, since we are iterating, `None` is simpler.

taipy/core/_repository/_fs_base.py
florian-vuillemot: This is a good idea, so I combined both as you proposed. But we are not returning a list. What do you think of the new implementation? Is it simpler?

taipy/core/_manager/_manager.py
```diff
@@ -59,6 +59,13 @@ def _get_all(cls) -> List[EntityType]:
         """
         return cls._repository._load_all()
 
+    @classmethod
+    def _get_all_by(cls, by) -> List[EntityType]:
+        """
+        Returns all entities.
+        """
+        return cls._repository._load_all_by(by)
+
```
jrobinAV, line 68: If we add a new method in the managers, we need to authorize it in enterprise version.

taipy/core/_repository/_fs_base.py
jrobinAV: absolutely

taipy/core/data/_data_manager.py
```diff
@@ -38,11 +39,33 @@ def _get_or_create(
...
         parent_id = pipeline_id if scope == Scope.PIPELINE else scenario_id if scope == Scope.SCENARIO else None
 
-        if dn_from_parent := cls._repository._get_by_config_and_parent_ids(data_node_config.id, parent_id):
+        if dn_from_parent := cls._repository._get_by_config_and_parent_id(data_node_config.id, parent_id):
             return dn_from_parent
 
         return cls._create_and_set(data_node_config, parent_id)
 
+    @classmethod
+    def _get_or_creates(
```
jrobinAV: get_or_creates is not good to me. 
Rename to _get_or_create_multiple or something like that.

taipy/core/data/_data_manager.py
```diff
@@ -38,11 +39,33 @@ def _get_or_create(
         scope = data_node_config.scope
```
jrobinAV: Is the method _get_or_create still used?

taipy/core/pipeline/pipeline.py
```diff
@@ -52,12 +52,13 @@ def __init__(
...
         parent_id: Optional[str] = None,
         subscribers: List[Callable] = None,
+        is_consistent=None
     ):
         self.config_id = _validate_id(config_id)
         self.id: PipelineId = pipeline_id or self._new_id(self.config_id)
         self.parent_id = parent_id
         self._tasks = {task.config_id: task for task in tasks}
-        self.is_consistent = self.__is_consistent()
+        self.is_consistent = is_consistent or self.__is_consistent()
```
jrobinAV: I don't think your new implementation works regarding `is_consistent`. The is_consistent attribute must be recomputed each time we modify the dag.

taipy/core/_manager/_manager.py
```diff
@@ -59,6 +59,13 @@ def _get_all(cls) -> List[EntityType]:
         """
         return cls._repository._load_all()
 
+    @classmethod
+    def _get_all_by(cls, by) -> List[EntityType]:
```
jrobinAV, line 63: Signatures for methods like get_all_by should be like : 
`def _get_all_by(cls, attribute_name, attribute_value) -> List[EntityType]:`
otherwise it is really error prone.

taipy/core/_repository/_fs_base.py
```diff
@@ -121,30 +136,60 @@ def _delete_many(self, model_ids: Iterable[str]):
...
         return filter(lambda e: getattr(e, attribute, None) == value, self._load_all())
 
-    def __get_model_filepath(self, model_id, raise_if_not_exist=True) -> pathlib.Path:
-        filepath = self._directory / f"{model_id}.json"
+    def __get_model_filepath(self, model_id) -> pathlib.Path:
+        return self.dir_path / f"{model_id}.json"
 
-        if raise_if_not_exist and not filepath.exists():
-            raise ModelNotFound(str(self._directory), model_id)
+    def __to_entity(self, filepath, by: Optional[str] = None, bys: Optional[List[str]] = None) -> Entity:
```
jrobinAV: the by parameter should be a key/value pair.
The signature should be like : 
def __to_entity(cls, filepath, name_value: Optional[List[str, str]]) -> EntityType:

taipy/core/task/_task_manager.py
```diff
@@ -67,6 +67,46 @@ def _get_or_create(
         cls._set(task)
         return task
 
+    @classmethod
+    def _get_or_creates(
```
jrobinAV: bad name

jrobinAV (changes requested):
raw_historical_data.csv must be anonymized before merging

taipy/core/_repository/_fs_base.py
florian-vuillemot: Let's go for EntityNotFound !

taipy/core/_manager/_manager.py
florian-vuillemot: Ok, we have to do it.

taipy/core/_manager/_manager.py
florian-vuillemot: I'm not in love with the "by" name. Do you have any better idea? 
cc @toan-quach

taipy/core/data/_data_manager.py
florian-vuillemot: At least in tests. We have to check for: datanode, task and pipeline.

taipy/core/data/_data_manager.py
florian-vuillemot: bulk_get_or_create ? batch_get_or_create ?

jrobinAV: yes why not.

taipy/core/data/_data_manager.py
jrobinAV: SHould not we use the `bulk_get_or_create` So we are using the real-life code?

taipy/core/pipeline/pipeline.py
florian-vuillemot: Is it necessary when we are getting the Pipeline from our repo?  
This function is time-consuming.

taipy/core/_manager/_manager.py
jrobinAV: that is quite standard for document-based queries.

taipy/core/pipeline/pipeline.py
jrobinAV: It depends if this attribute has been recomputed before to store it.

taipy/core/_manager/_manager.py
florian-vuillemot: We are not searching y attribute name or value. We are just searching if something contains this value. 
What do you think  of this signature `def _get_all_by(cls, should_contain: str)`

taipy/core/_repository/_fs_base.py
florian-vuillemot: We missed a part of the implementation.
We used ModelNotFound because we know nothing of the type of entity we are searching for.
I think -- I don't remember -- that we design it for throwing "ModelNotFound" because technically, we are searching for a Model -- we are passing a modelID --. But this error should be handled by the caller layer that will return the correct error like "TaskNotFound".
So this makes sense to me.

What do you think?

toan-quach: Absolutely agree :D

toan-quach: On the layer that called the function, we are returning an error log (with the entity class name), I think this is good enough

taipy/core/_manager/_manager.py
toan-quach: how about def _get_all_if_contains(cls, query/should_contain)?

taipy/core/_repository/_fs_base.py
toan-quach: I think `contains` is better than `name_value` though

taipy/core/data/_data_manager.py
toan-quach: This created some inconsistency between datamanager and the rest of the managers (datamanager doesn't have get_or_create). I dont think it's a problem, but think it's something to be mindful about to make sure the code is clean and consistent :D

taipy/core/_manager/_manager.py
florian-vuillemot: @jrobinAV

taipy/core/data/_data_manager.py
```diff
@@ -29,21 +29,6 @@ class _DataManager(_Manager[DataNode]):
...
-        parent_id = pipeline_id if scope == Scope.PIPELINE else scenario_id if scope == Scope.SCENARIO else None
-
-        if dn_from_parent := cls._repository._get_by_config_and_parent_id(data_node_config.id, parent_id):
-            return dn_from_parent
-
-        return cls._create_and_set(data_node_config, parent_id)
-
     @classmethod
     def _bulk_get_or_create(
         cls,
```
jrobinAV, line 34: Would it make more sense to pass `data_node_configs` as a `Set[DataNodeConfig]` instead of an Iterable?
The return type could be a `Dict[DataNodeConifg, DataNode]`. What do you think?

taipy/core/_manager/_manager.py
jrobinAV: I personally prefer `get_all_by` which is more generic and does not expose how we are searching for the attribute provided.
We are at the manager level. We absolutely don't want to expose how it is implemented. By the way, at the repository level, we don't want either to expose the fact that we are just checking for a string matching in a file.

taipy/core/_repository/_fs_base.py
jrobinAV: yep

taipy/core/_manager/_manager.py
khaitruong922: I would prefer `get_all_by_key` for more clarity or just `find` for simplicity.

@jrobinAV I thought that anyone can read the code so there is nothing we can and should hide. For the end users, normally they would not access the manager and repository directly so I wonder if this is a problem.

jrobinAV: Yes, indeed we cannot hide it.
What I meant is that the logic of the file system implementation should stay at the repository level. The API (the method name) should be generic so we don't care if it is implemented with FS or database. Basically, we want to separate responsibilities between managers and repositories.

taipy/core/_manager/_manager.py
jrobinAV: @florian can you override this method on enterprise version?

taipy/core/data/_data_manager.py
```diff
@@ -29,19 +30,25 @@ class _DataManager(_Manager[DataNode]):
     _ENTITY_NAME = DataNode.__name__
 
     @classmethod
-    def _get_or_create(
+    def _bulk_get_or_create(
```
jrobinAV, line 33: Must be protected and overridden in enterprise version.

taipy/core/task/_task_manager.py
```diff
@@ -45,27 +45,45 @@ def _set(cls, task: Task):
         super()._set(task)
 
     @classmethod
-    def _get_or_create(
+    def _bulk_get_or_create(
```
jrobinAV, line 48: Must be protected and overridden in enterprise version.

jrobinAV (dismissed):

taipy/core/_repository/_fs_base.py
```diff
@@ -125,36 +138,92 @@ def _delete_many(self, model_ids: Iterable[str]):
...
+        data = json.loads(file_content, cls=_CustomDecoder)
+        model = self.model.from_dict(data)  # type: ignore
+        return self._from_model(model)
+
     def __create_directory_if_not_exists(self):
-        if not self.dir_path.exists():
-            self.dir_path.mkdir(parents=True, exist_ok=True)
+        self.dir_path.mkdir(parents=True, exist_ok=True)
+
+    def __match_file_and_get_entity(self, filepath, config_and_parent_ids, read_entity_retry=None):
```
jrobinAV:
```suggestion
    def __match_file_and_get_entity(self, filepath, config_and_parent_ids, retry: Optional[int] = 0):
```

taipy/core/_repository/_fs_base.py
```diff
@@ -125,36 +138,92 @@ def _delete_many(self, model_ids: Iterable[str]):
...
+                    entity = self.__model_to_entity(file_content)
+                    if entity.parent_id == parent_id and entity.config_id == config_id.id:
+                        return config_id, parent_id, entity
+            except Exception as e:
+                read_entity_retry = (
+                    Config.global_config.read_entity_retry or 0 if read_entity_retry is None else read_entity_retry
+                )
+                if read_entity_retry == 0:
+                    raise e
+                return self.__match_file_and_get_entity(filepath, config_and_parent_ids, read_entity_retry - 1)
```
jrobinAV:
```suggestion
                 if retry and retry > 0:
                     return self.__match_file_and_get_entity(filepath, config_and_parent_ids, 
                                                             retry=retry - 1)
                 raise e
                
```

taipy/core/_repository/_fs_base.py
```diff
@@ -125,36 +138,92 @@ def _delete_many(self, model_ids: Iterable[str]):
...
+        # Design in order to optimize performance on Entity creation.
+        # Maintainability and readability were impacted.
+        res = {}
+        configs_and_parent_ids = set(configs_and_parent_ids)
 
-    def __get_model_filepath(self, model_id, raise_if_not_exist=True) -> pathlib.Path:
-        filepath = self._directory / f"{model_id}.json"
+        try:
+            for f in self.dir_path.iterdir():
+                config_id, parent_id, entity = self.__match_file_and_get_entity(f, configs_and_parent_ids)
```
jrobinAV:
```suggestion
                config_id, parent_id, entity = self.__match_file_and_get_entity(f, configs_and_parent_ids, retry=Config.global_config.read_entity_retry or 0)
```

jrobinAV (dismissed):
I would make the retry consistent. Otherwise that s good!

Issue #222, opened 2022-06-08
Lazy reloading


toan-quach:
**Description**
 Improve performance by reducing the amount of times take to reload objects

PR #225, opened 2022-06-09
feature/#222 added lazy loading to scenario_repo and pipeline_repo

toan-quach:
#222 
What I did: 

- Added lazy loading to scenario repository for pipelines and cycle
- Added lazy loading to pipeline repository for tasks
- Added lazy_loading params to related functions

**Side effect: in lazy loading, if the child entity is taken from the original entity (instead of loading from json), it will have the same reference with the child entity of the original entity.**

### DIFF
--- a/taipy/core/_manager/_manager.py
+++ b/taipy/core/_manager/_manager.py
@@ -11,10 +11,10 @@
 
 from typing import Generic, Iterable, List, TypeVar, Union
 
-from taipy.core._repository import _FileSystemRepository
 from taipy.core.common._entity_ids import _EntityIds
 from taipy.core.common._taipy_logger import _TaipyLogger
 
+from taipy.core._repository import _FileSystemRepository
 from taipy.core.exceptions.exceptions import ModelNotFound
 
 EntityType = TypeVar("EntityType")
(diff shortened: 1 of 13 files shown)


taipy/core/_manager/_manager.py
```diff
@@ -60,13 +61,14 @@ def _get_all(cls) -> List[EntityType]:
         return cls._repository._load_all()
 
     @classmethod
-    def _get(cls, entity: Union[str, EntityType], default=None) -> EntityType:
+    def _get(cls, entity: Union[str, EntityType], default=None, lazy_loading: bool = True) -> EntityType:
```
khaitruong922, line 64: I think naming the parameter `lazy` should be enough instead of `lazy_loading`.

toan-quach: lazy can be many things :D it'd be better to have a clearer name, I'm open to new ideas, lazy_loading is not my best name tbh 😄 (I'm looking up for a more conventional name if any in other packages)

florian-vuillemot:
Well, it's not what we have in mind, the card is really not explicit enough. What you did it's not a real lazy loading, it's more a cache in our context. The purpose of the card is to load Pipeline only when it's needed.

Ex:
scenario = tp.get(..) # Pipelines are not loaded
scenario.pipelines # Pipelines are loaded

I created the https://github.com/Avaiga/taipy-core/pull/227 to POC the idea. Can you have a look ?

jrobinAV:
Since it is not what we want, can we close this PR without merging.

PR #234, opened 2022-06-15
Feature/subscribe with args

joaoandre-avaiga:
This PR adds an option to create subscribers with parameters. This functionality is implemented in both `Scenario` and `Pipeline`. The main idea is that the list of subscribers of an entity is no longer just a list of callbacks but a list of a `Subscriber` object, which is a named tuple with two attributes: callback and params.

Please challenge this implementation and let me know if there are any other ideas.

### DIFF
--- a/taipy/core/common/_utils.py
+++ b/taipy/core/common/_utils.py
@@ -10,6 +10,7 @@
 # specific language governing permissions and limitations under the License.
 
 import functools
+from collections import namedtuple
 from importlib import import_module
 from operator import attrgetter
 from typing import Callable, Optional
@@ -30,7 +31,7 @@ def _get_fct_name(f) -> Optional[str]:
 
 
 def _fct_to_dict(obj):
-    fct_name = _get_fct_name(obj)
+    fct_name = _get_fct_name(obj.callable)
     if not fct_name:
         return None
     return {"fct_name": fct_name, "fct_module": obj.__module__}
@@ -38,3 +39,6 @@ def _fct_to_dict(obj):
 
 def _fcts_to_dict(objs):
     return [d for obj in objs if (d := _fct_to_dict(obj)) is not None]
+
+
+Subscriber = namedtuple("Subscriber", "callable params")
--- a/taipy/core/common/_utils.py
+++ b/taipy/core/common/_utils.py
@@ -31,13 +31,13 @@ def _get_fct_name(f) -> Optional[str]:
 
 
 def _fct_to_dict(obj):
-    fct_name = _get_fct_name(obj.callable)
+    fct_name = _get_fct_name(obj.callback)
     if not fct_name:
         return None
     return {
         "fct_name": fct_name,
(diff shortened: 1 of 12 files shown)


khaitruong922 (approved):
LGTM. I would prefer naming the values passed into a function `args` instead of `params`.

taipy/core/common/_utils.py
```diff
@@ -30,11 +31,25 @@ def _get_fct_name(f) -> Optional[str]:
 
 
 def _fct_to_dict(obj):
-    fct_name = _get_fct_name(obj)
+    params = []
+    callback = obj
```
jrobinAV, line 35: is the variable callback useful? can't we use obj directly?

taipy/core/scenario/_scenario_repository.py
```diff
@@ -69,7 +73,9 @@ def __to_pipeline_ids(pipelines) -> List[PipelineId]:
 
     @staticmethod
     def __to_cycle(cycle_id: CycleId = None) -> Optional[Cycle]:
-        return _CycleManagerFactory._build_manager()._get(cycle_id) if cycle_id else None
+        return (
+            _CycleManagerFactory._build_manager()._get(cycle_id) if cycle_id else None
+        )
```
jrobinAV, line 78:
```suggestion
        return _CycleManagerFactory._build_manager()._get(cycle_id) if cycle_id else None
```

tests/core/scenario/test_scenario_manager.py
```diff
@@ -317,7 +317,7 @@ def test_notification_subscribe(mocker):
     mocker.patch.object(_utils, "_load_fct", side_effect=[notify_1, notify_2])
 
     # test subscribing notification
-    _ScenarioManager._subscribe(notify_1, scenario)
+    _ScenarioManager._subscribe(callback=notify_1, scenario=scenario)
```
jrobinAV, line 320: Can you create a test where the callback takes multiple params?

jrobinAV (approved):
Sounds good to me. 
Please just add a test to check that we can pass multiple params.

taipy/core/common/_utils.py
joaoandre-avaiga: Obj in this case can be either a direct function or it can be a tuple(fct, params), if it is the tuple we need to check and replace the callback variable.

taipy/core/scenario/_scenario_repository.py
joaoandre-avaiga: yeah, I had an issue with black on pycharm that was fixed on 88 cols. I tought I had fixed the weird formatting, will change it.

tests/core/scenario/test_scenario_manager.py
joaoandre-avaiga: Yes! I will add the tests in the the second half of this PR which handles the jobs.

PR #249, opened 2022-06-21
DO NOT MERGE BEFORE 1.1 is released - Feature/use external config

jrobinAV:
Sorry, it is unreviewable. 
- I added a root folder `src/`
- I replaced all imports using the relative path in the source code.
- I replaced all test imports using the absolute path in the tests (with src.taipy.core...)
- I moved all config-related code to taipy-config.


khaitruong922 (commented):
Where is the `common` folder moved to?

jrobinAV:
> Where is the `common` folder moved to?

The common folder remains. Some of the files have been moved to taipy-config though.

Issue #158, opened 2022-04-14
Allow  jobs cancellation


toan-quach:
**Description**
- Expose a method in Job_manager to cancel a job from its id or the entity.

```
_JobManager.cancel(job.id)
_JobManager.cancel(job)
```

- Expose method in Taipy to cancel a job from its id or the entity.

```
tp.cancel(job.id)
tp.cancel(job)
```

- Expose method in Job to cancel itself. 
`job.cancel()`

**Acceptance Criteria**
- [ ] Ensure new code is unit tested, and check code coverage is at least 90%
- [ ] Propagate any change in the demos and run all of them to ensure there is no breaking change
- [ ] Propagate any change in taipy-rest 
- [ ] Ensure any change is well documented

PR #256, opened 2022-06-28
#158 added job cancelation

toan-quach:
TODO: 
- test job cancellation effect on wider scale pipeline
- cannot cancel job running (process cannot be canceled once started running)

When reviewing the PR, please pay extra attention on the tests. I hope to cover as much tests as possible. But some of the use cases are very hard to code (for example, when a job is canceled when running, it should push the next job to be executed)


taipy/core/common/_utils.py
```diff
@@ -52,4 +52,16 @@ def _fcts_to_dict(objs):
...
+
+    return _SchedulerFactory._build_scheduler()._processes.pop(job_id, default)  # type: ignore
+
+
+def _set_process_in_scheduler(job_id, process):
+    from taipy.core._scheduler._scheduler_factory import _SchedulerFactory
+
+    _SchedulerFactory._build_scheduler()._processes[job_id] = process
+
+
```
jrobinAV: I would try to move this logic into the scheduler.

taipy/core/_scheduler/_dispatcher/_standalone_job_dispatcher.py
```diff
@@ -39,11 +41,14 @@ def _dispatch(self, job: Job):
         future = self._executor.submit(
             self._run_wrapped_function, Config.global_config.storage_folder, job.id, job.task
         )
+        _set_process_in_scheduler(job.id, future)
```
jrobinAV: I would make the _dispatch method return the future, and I would call the _set_process method in the scheduler.

taipy/core/job/_job_manager.py
```diff
@@ -37,11 +39,24 @@ def _create(cls, task: Task, callbacks: Iterable[Callable], force=False) -> Job:
...
+    @classmethod
+    def _cancel(cls, job: Union[str, Job]):
+        job_id = job if isinstance(job, str) else job.id
+        process = _pop_process_in_scheduler(job_id)
+
+        if process:
+            process.cancel()
+            job = _JobManager._get(job) if isinstance(job, str) else job
+            job.cancelled()
+        # else:
```
jrobinAV: This is the most difficult case when the job is `RUNNING`.
But you also want to be able to cancel the jobs that don't have a process yet. for instance, the jobs that are `SUBMITTED`, `PENDING`, or `BLOCKED`.
It means you need to change the job status, but you need also to remove the job from scheduler queues (`blocked_jobs` and `jobs_to_run`).

taipy/core/_scheduler/_dispatcher/_standalone_job_dispatcher.py
```diff
@@ -39,11 +41,14 @@ def _dispatch(self, job: Job):
...
+
         future.add_done_callback(self.__release_worker)
         future.add_done_callback(partial(self._update_status_from_future, job))
 
     def __release_worker(self, _):
         self._nb_available_workers += 1
 
     def _update_status_from_future(self, job: Job, ft):
+        _pop_process_in_scheduler(job.id)
```
jrobinAV: I believe the job_manager already pops out the job_id from this dict.

taipy/core/common/_utils.py
toan-quach: What do you think about moving this to the AbstractScheduler?

taipy/core/job/_job_manager.py
toan-quach: Agree! Though I must note, the job that is running CANNOT be canceled as I mentioned earlier as the Future that are already running cannot be canceled. Reference to the issue: https://stackoverflow.com/questions/42782953/python-concurrent-futures-how-to-make-it-cancelable/45515052#45515052

taipy/core/_scheduler/_dispatcher/_standalone_job_dispatcher.py
toan-quach: Ah yes, this is for when the job is completed and not cancelled

taipy/core/_scheduler/_scheduler.py
```diff
@@ -195,3 +204,38 @@ def _update_job_config(cls):
...
+    def _set_process_in_scheduler(cls, job_id, process):
+        cls._processes[job_id] = process
+
+    @classmethod
+    def _pop_process_in_scheduler(cls, job_id, default=None):
+        return cls._processes.pop(job_id, default)  # type: ignore
+
+    @classmethod
+    def _cancel_job(cls, job):
+        if process := cls._pop_process_in_scheduler(job.id):
```
jrobinAV: What about retrieving directly the submit_id here and cancel all the jobs with the same submit_id ? WOuld it make the design easier?

jrobinAV: What about retrieving the submit_id here and cancel all the jobs with the same submit_id.
would it make the code design easier to understand?

taipy/core/task/_task_manager.py
```diff
@@ -113,4 +114,6 @@ def _submit(cls, task: Union[TaskId, Task], callbacks: Optional[List[Callable]]
         task = cls._get(task_id)
         if task is None:
             raise NonExistingTask(task_id)
-        cls._scheduler().submit_task(task, callbacks, force)
+
+        submit_id = f"SUBMISSION_{str(uuid.uuid4())}"
```
jrobinAV: Here you can pass None as submit_id and actually create it in the scheduler.submit_task method.

taipy/core/_scheduler/_scheduler.py
```diff
@@ -174,18 +180,21 @@ def __unlock_edit_on_outputs(job):
...
-    def __unblock_jobs(cls):
+    def __unblock_jobs(cls, if_canceled: bool = False):
         for job in cls.blocked_jobs:
             if not cls._is_blocked(job):
                 with cls.lock:
                     try:
-                        job.pending()
+                        if not if_canceled:
+                            job.pending()
+                            cls.jobs_to_run.put(job)
```
jrobinAV: I don't get why you need to change this in the __unblock_jobs method.

taipy/core/_scheduler/_scheduler.py
```diff
@@ -195,3 +204,38 @@ def _update_job_config(cls):
...
+    @classmethod
+    def _pop_process_in_scheduler(cls, job_id, default=None):
+        return cls._processes.pop(job_id, default)  # type: ignore
+
+    @classmethod
+    def _cancel_job(cls, job):
+        if process := cls._pop_process_in_scheduler(job.id):
+            process.cancel()  # TODO: this doesn't cancel the running process
+        job.cancelled()
+        cls.__unlock_edit_on_outputs(job)
```
jrobinAV: I don't get this line.

taipy/core/_scheduler/_scheduler.py
```diff
@@ -195,3 +204,38 @@ def _update_job_config(cls):
...
+
+        # if job in cls._needs_to_run:
+        #     cls._needs_to_run.remove(job)
+
+        if job in cls.blocked_jobs:
+            cls.blocked_jobs.remove(job)
```
jrobinAV: Needs to be locked first.

jrobinAV (changes requested):
I believe we can have a discussion to better understand what you have in mind.

taipy/core/_scheduler/_scheduler.py
```diff
@@ -174,18 +180,21 @@ def __unlock_edit_on_outputs(job):
...
 
     @classmethod
-    def __unblock_jobs(cls):
+    def __unblock_jobs(cls, if_canceled: bool = False):
         for job in cls.blocked_jobs:
             if not cls._is_blocked(job):
                 with cls.lock:
                     try:
-                        job.pending()
+                        if not if_canceled:
```
jrobinAV: At this point, you don't know if the current job has been submitted with the same submit ID than the one canceled.
What about

jrobinAV (changes requested):

taipy/core/_scheduler/_scheduler.py
toan-quach: because when you called job.cancelled(), it will trigger on_status_change which triggered __unblock_jobs, if you remove the cancelled job from blocked_jobs prior or after calling job.cancelled(), it will complain saying not being able to find the job

taipy/core/_scheduler/_scheduler.py
toan-quach: im not sure about the 1st part "What about retrieving the submit_id here", as for the 2nd part "cancel all the jobs with the same submit_id", we discussed and want to cancel only the subsequent jobs and leave the others to continue running :D

taipy/core/_scheduler/_scheduler.py
toan-quach: this is to unlock datanodes of the job because it was locked when submitted, to make sure the next time we submit, everything still works fine or in simpler words, because the job is already canceled, there's no reason to lock the dn outputs anymore. Does it make sense?

taipy/core/_scheduler/_scheduler.py
toan-quach: we're only looking at cls.blocked_jobs so the job is already locked

taipy/core/task/_task_manager.py
toan-quach: agree! in submit_task, i will check if submit_id is provided, if nothing is provided, i will generated 1

src/taipy/core/_scheduler/_scheduler.py
```diff
@@ -73,15 +75,18 @@ def submit(
...
         return res
 
     @classmethod
-    def submit_task(cls, task: Task, callbacks: Optional[Iterable[Callable]] = None, force: bool = False) -> Job:
+    def submit_task(
+        cls, task: Task, submit_id: str = None, callbacks: Optional[Iterable[Callable]] = None, force: bool = False
+    ) -> Job:
         """Submit the given `Task^` for an execution.
 
         Parameters:
```
jrobinAV, line 92: Need to update the docstring

src/taipy/core/_scheduler/_scheduler.py
```diff
@@ -185,14 +197,86 @@ def __unblock_jobs(cls):
...
-                        cls.blocked_jobs.remove(job)
                         cls.jobs_to_run.put(job)
+                        cls.__remove_blocked_job(job)
                     except:
                         cls.__logger.warning(f"{job.id} is not in the blocked list anymore.")
 
+    @classmethod
+    def __remove_blocked_job(cls, job: Job):
+        if job in cls.blocked_jobs:
+            cls.blocked_jobs.remove(job)
```
jrobinAV: I believe you don't need this method since it is already handle in `__unblock_jobs` method by the try catch.

src/taipy/core/_scheduler/_scheduler.py
```diff
@@ -175,8 +186,9 @@ def __unlock_edit_on_outputs(job):
     @classmethod
     def _on_status_change(cls, job: Job):
         if job.is_finished():
-            cls.__unblock_jobs()
-            cls.__execute_jobs()
+            if not job.is_cancelled():
```
jrobinAV: I believe you don't need this if condition since if the current job is cancelled, the subsequent jobs are also cancelled. That means they will not be in the `blocked_jobs` list nor in the `jobs_to_run` queue.

src/taipy/core/_scheduler/_scheduler.py
```diff
@@ -185,14 +197,86 @@ def __unblock_jobs(cls):
...
+            cls.__cancel_process(job)
+            job.cancelled()
+
+            # cls.__unlock_edit_on_outputs(job)
+
+        if not cls.jobs_to_run.empty():
+            cls.__execute_jobs()
+        else:
+            cls.__unblock_jobs()
+            cls.__execute_jobs()
```
jrobinAV: Why do you need to re execute some jobs after the cancelation? 
either there is a job already running that will trigger the next execution, either there is no running job that means there is nothing to run. AM I missing something here ?

src/taipy/core/_scheduler/_scheduler.py
```diff
@@ -185,14 +197,86 @@ def __unblock_jobs(cls):
...
+    def _cancel_job(cls, job):
+        to_cancel_jobs = set()
+        to_cancel_jobs.update([job])
+        to_cancel_jobs.update(cls.__find_subsequent_jobs(job.submit_id, set(job.task.output.keys())))
+        cls.__remove_jobs_from_running(to_cancel_jobs)
+        for job in to_cancel_jobs:
+            cls.__cancel_process(job)
+            job.cancelled()
+
+            # cls.__unlock_edit_on_outputs(job)
```
jrobinAV: Indeed, we need to unlock_edit for all the outputs of all the jobs canceled.

src/taipy/core/job/status.py
```diff
@@ -47,3 +47,4 @@ class Status(_ReprEnum):
     FAILED = 6
     COMPLETED = 7
     SKIPPED = 8
+    ABUNDANT = 9
```
jrobinAV:
```suggestion
    ABANDONED = 9
```

src/taipy/core/taipy.py
```diff
@@ -363,6 +363,11 @@ def delete_jobs():
     return _JobManagerFactory._build_manager()._delete_all()
 
 
+def cancel_job(job: Union[str, Job]):
+    """Cancel job"""
```
```suggestion
    """Cancel a job and set the status of the subsequent jobs to abandoned.
    
    Parameters:
        job (Job^): The job to cancel.
    """
```

tests/core/job/test_job_repository.py
```diff
@@ -64,15 +64,16 @@ def h():
         pass
 
 
-job = Job(JobId("id"), task)
+job = Job(JobId("id"), task, "submit_id")
 job._subscribers = [f, A.f, A.g, A.h, A.B.f]
-job._exceptions = [traceback.TracebackException.from_exception(Exception())]
+# job._exceptions = [traceback.TracebackException.from_exception(Exception())]
```
jrobinAV: Why this change?

jrobinAV (dismissed):
I believe it is working well!

However, I want to propose some minor code refactoring that could make the whole cancel feature more readable. Basically, I just want to move some parts of your code into other functions and rename the functions.

Tell me what you think.

src/taipy/core/_scheduler/_scheduler.py
```diff
@@ -196,3 +212,68 @@ def _update_job_config(cls):
...
+        with cls.lock:
+            cls.__remove_blocked_jobs(to_cancel_jobs)
+            cls.__remove_jobs_to_run(to_cancel_jobs)
+            cls.__cancel_processes(to_cancel_jobs)
+            cls.__unlock_edit_on_outputs(to_cancel_jobs)
+        # if not cls.jobs_to_run.empty():
+        #     cls.__execute_jobs()
+        # else:
+        #     cls.__unblock_jobs()
+        #     cls.__execute_jobs()
```
jrobinAV: I commented this part of the code 'cause I believe it is not necessary.

jrobinAV (dismissed):

taipy/core/_scheduler/_scheduler.py
jrobinAV: Absolutely!

src/taipy/core/_scheduler/_scheduler.py
toan-quach: @jrobinAV I would say in the scenario where the process pool is full, and we are able to cancel 1 process, we will need to call `__execute_jobs()` to execute the next job since the previous canceled job was never finished and it can't initiate the next job. Of course, currently, this is not needed as we can't cancel a process yet

src/taipy/core/_scheduler/_scheduler.py
toan-quach: @jrobinAV I think we should discussed this further, if we submit 1 pipeline twice, meaning we will have 2 sets of jobs based on the same pipeline => same datanodes. If we unlock output when canceling a set of jobs, it will affect the other set of jobs

src/taipy/core/_scheduler/_scheduler.py
toan-quach: @jrobinAV When canceling a job, meaning the process pool will have 1 process available. Ideally, it will be occupied with the next job in the queue is my idea. Canceling a job, if possible, will cancel a process, which means the process was never done and cannot initiate the execution of the next job. If there's no current running job to initiate the next job when done, meaning if there's still some jobs left in the queue, there's nothing that will call the next job to initiate. I'm not sure if my explanation is clear enough. Do let me know what you think

tests/core/job/test_job_repository.py
toan-quach: It wasn't used anywhere or for any purposes. I'm thinking of removing it

src/taipy/core/_scheduler/_scheduler.py
toan-quach: Im not sure I follor the 2nd part "already handle in __unblock_jobs method by the try catch."

src/taipy/core/_scheduler/_scheduler.py
jrobinAV: you are right!

tests/core/job/test_job_repository.py
jrobinAV: If we don't store it, the developer will not be able to use it for instance to debug or to show it to the end-user in a UI.

toan-quach: ah I see, sure!

src/taipy/core/taipy.py
```diff
@@ -364,7 +364,11 @@ def delete_jobs():
 
 
 def cancel_job(job: Union[str, Job]):
-    """Cancel job"""
+    """Cancel a job and set the status of the subsequent jobs to abandoned.
```
jrobinAV: to "ABANDONED" in uppercase

src/taipy/core/_scheduler/_scheduler.py
```diff
@@ -132,7 +143,10 @@ def __execute_jobs(cls):
             cls._update_job_config()
         while not cls.jobs_to_run.empty() and cls._dispatcher._can_execute():
             with cls.lock:
-                job = cls.jobs_to_run.get()
+                try:
+                    job = cls.jobs_to_run.get()
+                except:  # In case between line 144 and 145 the last job of the queue has been removed.
```
florian-vuillemot:
```suggestion
                except:  # In case between line 144 and 145 the last job of the queue has been removed.
```
To improve, please, do not line number in a comment, this will be outdated quickly.

toan-quach: Hi, I agree. @jrobinAV I don't recall writing this comment. Do you happen to know where it originated from? I think it explaining why the try catch is needed and it's safe to remove it.

taipy/core/_scheduler/_dispatcher/_standalone_job_dispatcher.py
toan-quach: From my understanding of the problem, set_process must be called before add_done_callback because some of the job is too fast, after it is done, the call back (update_status -> pop_process) will be called before the set_process was called

src/taipy/core/_scheduler/_scheduler.py
toan-quach: I believe it should called execute_jobs to execute other remaining jobs in jobs_to_run queue

jrobinAV (dismissed):

src/taipy/core/_scheduler/_scheduler.py
```diff
@@ -168,31 +183,99 @@ def _needs_to_run(task: Task) -> bool:
...
     def _update_job_config(cls):
         if Config.job_config.is_standalone:  # type: ignore
             cls._dispatcher = _StandaloneJobDispatcher()
         else:
             cls._dispatcher = _DevelopmentJobDispatcher()
+
+    @classmethod
+    def cancel_job(cls, job: Job):
+        to_cancel_jobs = set()
+        to_cancel_jobs.update([job])
```
florian-vuillemot:
```suggestion
        to_cancel_jobs = set([job])
```

src/taipy/core/job/status.py
```diff
@@ -47,3 +47,4 @@ class Status(_ReprEnum):
     FAILED = 6
     COMPLETED = 7
     SKIPPED = 8
+    ABANDONED = 9
```
florian-vuillemot, line 50: What is the difference between CANCELLED and ABANDONED ?

tests/test_complex_application.py
```diff
@@ -173,3 +193,67 @@ def test_complex():
...
+
+#     task1_div_csv_with_constant = Config.configure_task(
+#         "task1_div_csv_with_constant", div_constant_with_sleep, input=dn1_inp_csv, output=dn2_div
+#     )
+#     task2_print_pickle = Config.configure_task("task2_print_pickle", print_with_sleep, input=dn2_div)
+#     task3_populate_constant = Config.configure_task(
+#         "task3_populate_constant", return_a_number_with_sleep, output=dn2_div
+#     )
+
+#     pipeline_div_print_config = Config.configure_pipeline(
```
florian-vuillemot, line 218: ????

src/taipy/core/job/status.py
toan-quach: ABANDONED is for the subsequent jobs that will not be executed because their prior job was cancelled. Same for FAILED, when a job is FAILED, the subsequent jobs of said job will be ABANDONED

tests/test_complex_application.py
toan-quach: This is the test to test the effect of unlock_editition when submitting the same pipeline multiple times. So far, I see no problem regarding the topic. But I think the test is a bit immature to conclude that the question is solved.

tests/core/job/test_job_manager.py
```diff
@@ -150,13 +152,94 @@ def test_force_deleting_unfinished_job():
...
+
+    with lock:
+        job = _Scheduler.submit_task(task, "submit_id")
+
+        assert job.is_running()
+        assert len(_Scheduler._processes) == 1
+        _JobManager._cancel(job.id)
+        assert job.is_cancelled()
+        assert len(_Scheduler._processes) == 0
+    assert job.is_cancelled()
```
jrobinAV, line 177: I would like to tests that when killing a process, the nb of available workers is updated as well.

```suggestion
    assert _Scheduler._dispatcher._nb_available_workers == 2
    with lock:
        job = _Scheduler.submit_task(task, "submit_id")

        assert job.is_running()
        assert len(_Scheduler._processes) == 1
        assert _Scheduler._dispatcher._nb_available_workers == 1
        _JobManager._cancel(job.id)
        assert job.is_cancelled()
        assert len(_Scheduler._processes) == 0
    assert job.is_cancelled()
    assert _Scheduler._dispatcher._nb_available_workers == 2
```

jrobinAV (dismissed):

tests/core/job/test_job_manager.py
toan-quach: @jrobinAV Oh we aren't killing the process at the moment. But yes, I agree we should have that when we implement the killing process. I'll take note of that to make sure it is included next time

jrobinAV: Yes, right, but still we can still test now, no?

toan-quach: I'd say we can. Just want to stress that the flow would be a bit different, what I mean by that is after `with lock:` the process dispatched by submitting the task will be completed and it should update the nb_available_workers to 2. Anyway, it certainly is possible, I will add it to the test :D

tests/test_complex_application.py
toan-quach: And because of the many uses of sleep, this test takes quite a long time to run (sthg around 6-7s), so I commented it out

jrobinAV (approved):

PR #263, opened 2022-07-07
Feature - JSON data node

khaitruong922:


src/taipy/core/data/json.py
```diff
@@ -0,0 +1,107 @@
...
+    @property  # type: ignore
+    @_self_reload(DataNode._MANAGER_NAME)
+    def path(self):
+        return self._path
+
+    @path.setter
+    def path(self, value):
+        self.properties[self.__DEFAULT_PATH_KEY] = value
+
+    def _read(self):
```
toan-quach, line 133: I think we can have a custom class converter for this datanode type as well. You can check the CsvDataNode under function read_as if it makes sense to you

tests/core/data/test_json_data_node.py
```diff
@@ -0,0 +1,103 @@
...
+
+    def test_read_invalid_json(self):
+        path = os.path.join(pathlib.Path(__file__).parent.resolve(), "data_sample/invalid.json.txt")
+        dn = JSONDataNode("foo", Scope.PIPELINE, properties={"default_path": path})
+        with pytest.raises(ValueError):
+            dn.read()
+
+    def test_write(self, json_file):
+        json_dn = JSONDataNode("foo", Scope.PIPELINE, properties={"default_path": json_file})
+        data = {"a": 1, "b": 2, "c": 3}
```
toan-quach, line 141: can we expand the test to value being string and boolean and other primitive types like float for example?

toan-quach (changes requested):

src/taipy/core/data/csv.py
```diff
@@ -42,7 +43,7 @@ class CSVDataNode(DataNode):
         edit_in_progress (bool): True if a task computing the data node has been submitted
             and not completed yet. False otherwise.
         properties (dict[str, Any]): A dictionary of additional properties. Note that the
-            _properties_ parameter must at least contain a _"path"_ entry representing the path
+            _properties_ parameter must at least contain a _"default_path"_ entry representing the path
```
jrobinAV: We must revert that.
The `default_path` is the mandatory field of the config.
In the Datanode the mandatory field is the `path`.

src/taipy/core/data/excel.py
```diff
@@ -46,7 +47,7 @@ class ExcelDataNode(DataNode):
         edit_in_progress (bool): True if a task computing the data node has been submitted
             and not completed yet. False otherwise.
         properties (dict[str, Any]): A dictionary of additional properties. Note that the
-            _properties_ parameter must at least contain a _"path"_ entry representing the path
+            _properties_ parameter must at least contain a _"default_path"_ entry representing the path
```
jrobinAV: Same remark here. We must revert this change.

src/taipy/core/data/json.py
```diff
@@ -0,0 +1,118 @@
...
+        parent_id (str): The identifier of the parent (pipeline_id, scenario_id, cycle_id) or `None`.
+        last_edit_date (datetime): The date and time of the last modification.
+        job_ids (List[str]): The ordered list of jobs that have written this data node.
+        validity_period (Optional[timedelta]): The validity period of a cacheable data node.
+            Implemented as a timedelta. If _validity_period_ is set to None, the data_node is
+            always up-to-date.
+        edit_in_progress (bool): True if a task computing the data node has been submitted
+            and not completed yet. False otherwise.
+        properties (dict[str, Any]): A dictionary of additional properties. Note that the
+            _properties_ parameter must at least contain a _"default_path"_ entry representing the path
```
jrobinAV:
```suggestion
            _properties_ parameter must at least contain a _"path"_ entry representing the path
```

src/taipy/core/data/json.py
```diff
@@ -0,0 +1,118 @@
...
+            of the JSON file.
+    """
+
+    __STORAGE_TYPE = "json"
+    __DEFAULT_PATH_KEY = "default_path"
```
jrobinAV, line 52:
```suggestion
    __PATH_KEY = "path"
```

src/taipy/core/data/pickle.py
```diff
@@ -45,7 +45,7 @@ class PickleDataNode(DataNode):
             When creating a pickle data node, if the _properties_ dictionary contains a
             _"default_data"_ entry, the data node is automatically written with the corresponding
             _"default_data"_ value.
-            If the _properties_ dictionary contains a _"path"_ entry, the data will be stored
+            If the _properties_ dictionary contains a _"default_path"_ entry, the data will be stored
```
jrobinAV: Same remark
```suggestion
            If the _properties_ dictionary contains a _"path"_ entry, the data will be stored
```

tests/conftest.py
```diff
@@ -59,6 +59,14 @@ def excel_file(tmpdir_factory) -> str:
     return fn.strpath
 
 
+@pytest.fixture(scope="function")
+def json_file(tmpdir_factory) -> str:
+    json = pd.DataFrame([{"a": 1, "b": 2, "c": 3}, {"a": 4, "b": 5, "c": 6}])
```
jrobinAV, line 64: Can you create a more complex test with heterogeneous types, like str, int, bool, Dates, list, dictionaries

tests/core/data/test_json_data_node.py
```diff
@@ -0,0 +1,178 @@
...
+        assert dn.job_ids == []
+        assert not dn.is_ready_for_reading
+        assert dn.path == path
+
+        with pytest.raises(InvalidConfigurationId):
+            dn = JSONDataNode(
+                "foo bar", Scope.PIPELINE, name="super name", properties={"default_path": path, "has_header": False}
+            )
+
+    def test_new_csv_data_node_with_existing_file_is_ready_for_reading(self):
```
jrobinAV:
```suggestion
    def test_new_json_data_node_with_existing_file_is_ready_for_reading(self):
```

tests/core/data/test_json_data_node.py
```diff
@@ -0,0 +1,178 @@
...
+        json_dn = JSONDataNode("foo", Scope.PIPELINE, properties={"default_path": json_file})
+        data = {"a": 1, "b": json_dn}
+        with pytest.raises(TypeError):
+            json_dn.write(data)
+
+    def test_set_path(self):
+        dn = JSONDataNode("foo", Scope.PIPELINE, properties={"default_path": "foo.csv"})
+        assert dn.path == "foo.csv"
+        dn.path = "bar.csv"
+        assert dn.path == "bar.csv"
```
jrobinAV:
```suggestion
    def test_set_path(self):
        dn = JSONDataNode("foo", Scope.PIPELINE, properties={"path": "foo.json"})
        assert dn.path == "foo.json"
        dn.path = "bar.json"
        assert dn.path == "bar.json"
```

tests/core/data/test_json_data_node.py
```diff
@@ -0,0 +1,178 @@
...
+            dn = JSONDataNode(
+                "foo bar", Scope.PIPELINE, name="super name", properties={"default_path": path, "has_header": False}
+            )
+
+    def test_new_csv_data_node_with_existing_file_is_ready_for_reading(self):
+        not_ready_dn_cfg = Config.configure_data_node(
+            "not_ready_data_node_config_id", "json", default_path="NOT_EXISTING.json"
+        )
+        not_ready_dn = _DataManager._bulk_get_or_create([not_ready_dn_cfg])[not_ready_dn_cfg]
+        assert not not_ready_dn.is_ready_for_reading
```
jrobinAV, line 89: We should test that `DataNodeConfig.default_path` correctly populate `datanode.path`:

```suggestion
        not_ready_dn = _DataManager._bulk_get_or_create([not_ready_dn_cfg])[not_ready_dn_cfg]
        assert not not_ready_dn.is_ready_for_reading
        assert not_ready_dn.path == "NOT_EXISTING.json"
```

jrobinAV (changes requested):

src/taipy/core/data/csv.py
khaitruong922: The properties come from the config, not the datanode. The mandatory field should be of the config.

src/taipy/core/data/json.py
khaitruong922: This is the `default_path` property from the config.

src/taipy/core/data/pickle.py
khaitruong922: It can be either but we want the user to use `default_path`

src/taipy/core/data/csv.py
jrobinAV: Right, but the idea was to expose the `default_path` property in the config, but when instantiating the data node, the property exposed should become `path`. I don't like the idea of exposing the naming `default_path` in data node while it represents the actual path, and not the default one.

This is not only related to your PR. It is already the case for CSV and Excel data node implementation. I believe we missed something here.

Let me do an implementation proposal to illustrate what I have in mind.

src/taipy/core/data/json.py
```diff
@@ -115,4 +116,13 @@ def _cast(self, data: Any):
...
+            json.dump(data, f, indent=4, cls=EnhancedJSONEncoder)
+
+
+class EnhancedJSONEncoder(json.JSONEncoder):
+    def default(self, o):
+        if isinstance(o, (datetime, date)):
+            return o.isoformat()
+        if dataclasses.is_dataclass(o):
+            return dataclasses.asdict(o)
+        return super().default(o)
```
jrobinAV: If you create a custom encoder I believe you need to have a custom decoder as well. Other wise it cannot work properly.

jrobinAV (changes requested):

src/taipy/core/data/json.py
khaitruong922: How about letting the user pass in the encoder and decoder? There are many use cases that the user may want to have a custom encoder and decoder by themselves?

khaitruong922: The custom decoder you mention about is the `exposed_type` property. However it cannot decode deeply nested objects.

jrobinAV: What is important is that when reading the data node, the user should retrieve the exact same json that he/she wrote.

In other words, the following should always pass whatever foo is: 
```
my_JSON_data_node.write(foo)
assert foo= my_JSON_data_node.read()

```
I believe we need to accept optional attributes in the data node config. Something like encoder/decoder, converter/exposed_type, or any other more meaningful names.

src/taipy/core/data/csv.py
```diff
@@ -75,6 +75,13 @@ def __init__(
             )
         if self.__HAS_HEADER_PROPERTY not in properties.keys():
             properties[self.__HAS_HEADER_PROPERTY] = True
+        if path := properties.get(self.__PATH_KEY):
+            self._path = path
+        elif path := properties.get(self.__DEFAULT_PATH_KEY):
+            self._path = path
+            properties[self.__PATH_KEY] = path
```
khaitruong922: Should we drop the `default_path` from the properties?

jrobinAV: I believe we should keep it.

src/taipy/core/data/excel.py
```diff
@@ -47,9 +47,8 @@ class ExcelDataNode(DataNode):
         edit_in_progress (bool): True if a task computing the data node has been submitted
             and not completed yet. False otherwise.
         path (str): The path to the Excel file.
-        properties (dict[str, Any]): A dictionary of additional properties. Note that the
-            _properties_ parameter must at least contain a _"default_path"_ entry representing the path
-            of the Excel file (xlsx format).
+        properties (dict[str, Any]): A dictionary of additional properties. The _properties_
+            must have a _"default_path"_ or _"path"_ entry with the path of the JSON file.
```
jrobinAV: Excel, not JSON

src/taipy/core/data/pickle.py
```diff
@@ -83,11 +86,11 @@ def __init__(
             edit_in_progress,
             **properties,
         )
-        self.__is_file_generated = False
-        self._pickle_path = self.__build_path()
-        if not self._last_edit_date and os.path.exists(self._pickle_path):
+        if self._path is None:
+            self._path = self.__build_path()
```
jrobinAV, line 93: these two lines should be executed before executing this: `properties[self.__PATH_KEY] = self._path`. No?

khaitruong922: because the __build_path() needs the `self.id` attribute so we must put it after the `super().init()` call. We don't need to save the generated path anyway because it is based on the `id`.

jrobinAV: For consistency reason, I believe it s better to save the path even when it is generated.
May be we can just execute the following code :
```
if properties is None:
    properties = {}
default_value = properties.pop(self.__DEFAULT_DATA_KEY, None)
self.__is_file_generated = False
super().__init__(
    config_id,scope,id,name,parent_id,last_edit_date,job_ids,validity_period,  
    edit_in_progress,**properties
)
 self._path = properties.get(self.__PATH_KEY, properties.get(self.__DEFAULT_PATH_KEY))
 if self._path is None:
     self._path = self.__build_path()
 if self._path is not None:
     properties[self.__PATH_KEY] = self._path
```
What do you think?

khaitruong922: If we save the path, the next time the data node is loaded, the `is_file_generated` will return False, which makes the pickle file not cleaned when we delete the data node. I think it is better to not save to make the behavior consistent

jrobinAV: you are right. So I would propose to set it in the properties so it is stored and can be retrieved.

```
if properties is None:
    properties = {}
default_value = properties.pop(self.__DEFAULT_DATA_KEY, None)
if not file_generated:= properties.get(self.__FILE_GENERATED_KEY, None):
    properties[self.__FILE_GENERATED_KEY]=False
super().__init__(
    config_id,scope,id,name,parent_id,last_edit_date,job_ids,validity_period,  
    edit_in_progress,**properties
)
 self._path = properties.get(self.__PATH_KEY, properties.get(self.__DEFAULT_PATH_KEY))
 if self._path is None:
     self._path = self.__build_path()
 if self._path is not None:
     properties[self.__PATH_KEY] = self._path
``` 

Then it must be updated in both the `is_file_generated` property and in the `__build_path` method as well.

What do you think?

khaitruong922: I don't think this is necessary because the path can be inferred from the id and the is_generated can be inferred if the path is None, so saving them is just like having duplicate information.

jrobinAV: This field `is_generated` is still useful.
The purpose of it is to know if the path has been generated by taipy, or if the file has been created by a user. So when we delete the entity we don't delete the file if it has not been generated.

khaitruong922: I mean if the path properties is None then we already know it is generated, so saving it is kinda redundant.

jrobinAV: Yes but what if the path is not known? How do we know if it has been generated by Taipy or if it comes from the user? Either we just use the `is_generated` field, or we need to run a regexp on the path to see if it is generated or not (which is much more complex).

tests/core/data/test_json_data_node.py
```diff
@@ -0,0 +1,220 @@
...
+        class MyEnum(Enum):
+            A = 1
+            B = 2
+            C = 3
+
+        json_dn = JSONDataNode("foo", Scope.PIPELINE, properties={"default_path": json_file})
+        data = [MyEnum.A, MyEnum.B, MyEnum.C]
+        json_dn.write(data)
+        read_data = json_dn.read()
+        assert read_data == [1, 2, 3]
```
jrobinAV, line 170: This function tests the ability to write a list of integers, not an enum. Depending on what you want to test here, either you could rename the test, or you can simplify it by removing the Enum class, or fix the test to check that you can serialize an enum.

Something like the following :

```
def test_write_enum(self, json_file):
        class MyEnum(Enum):
            A,
            B,
            C,

        json_dn = JSONDataNode("foo", Scope.PIPELINE, properties={"default_path": json_file})
        data = [MyEnum.A, MyEnum.B, MyEnum.C]
        json_dn.write(data)
        read_data = json_dn.read()
        assert read_data == [MyEnum.A, MyEnum.B, MyEnum.C]
```

khaitruong922: This is to test if our default encoder works properly with enum. Without the encoder, when we write an enum into JSON, this error will occur.

```
TypeError: Object of type MyEnum is not JSON serializable
```

jrobinAV: OK great.

jrobinAV (dismissed):

toan-quach (dismissed):

toan-quach (approved):

PR #267, opened 2022-07-21
feature/#260 skeleton code for refactoring dispatcher

toan-quach:
- keep the development mode fully synchronous.
- allow exposing in standalone mode a synchronous_submit method.
- running a dispatcher as a separate thread


src/taipy/core/_scheduler/_dispatcher/_dispatcher.py
```diff
@@ -0,0 +1,136 @@
...
+    def run(self):
+        while True:
+            # if not self.scheduler.jobs_to_run.empty() and self._can_execute():
+            # self.__execute_jobs()
+            # else:
+            from time import sleep
+
+            print("going to sleep")
+            sleep(1)
+
```
jrobinAV: Instead of doing a sleep(1), we can eventually modify the __execute_jobs method to benefit from the fact that jobs_to_run.get() blocks until there is an item in the queue.

You can have a look at this [link](https://docs.python.org/3/library/queue.html#queue.Queue.get)

In any case, this is more than fine for a first version.

src/taipy/core/_scheduler/_dispatcher/_dispatcher.py
```diff
@@ -0,0 +1,136 @@
...
+        raise NotImplementedError
+
+    def __execute_jobs(self):
+        with self.lock:
+            try:
+                job = self.scheduler.jobs_to_run.get()
+            except:  # In case the last job of the queue has been removed.
+                # TODO: instantiate logger object
+                self.__logger.warning(f"{job.id} is no longer in the list of jobs to run.")
+        if job.force or self.scheduler._needs_to_run(job.task):
```
jrobinAV: I would move __needs_to_run into the dispatcher class.

jrobinAV (dismissed):
I believe it s very promising.

src/taipy/core/_scheduler/_dispatcher/_dispatcher.py
toan-quach: that' great thanks!! Then I'd say we should rearrange the condition to `if self._can_execute() and job := self.scheduler.jobs_to_run.get(block=True):` to make sure the processes are available to execute a job and we won't have the issue where we continuously getting new jobs out of the queue and no available processes to run it.

src/taipy/core/_scheduler/_dispatcher/_standalone_job_dispatcher.py
```diff
@@ -9,23 +9,34 @@
...
 
-    def __init__(self):
-        super().__init__()
-        self._executor = ProcessPoolExecutor(Config.job_config.nb_of_workers or 1)
+    def __init__(self, scheduler: _AbstractScheduler):
+        super().__init__(scheduler)
+        self._executor = ProcessPoolExecutor(Config.job_config.nb_of_workers or 1)  # type: ignore
         self._nb_available_workers = self._executor._max_workers  # type: ignore
 
+    def start(self):
```
jrobinAV: This can be moved to the parent class _JobDispatcher, no?

src/taipy/core/_scheduler/_dispatcher/_job_dispatcher.py
```diff
@@ -20,19 +24,93 @@
...
+
+    def run(self):
+        while True and not self.__STOP_FLAG:
+            try:
+                if self._can_execute() and (job := self.scheduler.jobs_to_run.get(block=True)):
+                    self._execute_job(job)
+                else:
+                    sleep(0.1)  # Limit CPU usage
+            except:  # In case the last job of the queue has been removed.
+                self.__logger.warning(f"{job.id} is no longer in the list of jobs to run.")
```
jrobinAV: What about ?

```suggestion
    def run(self):
        while not self.__STOP_FLAG and (job := self.scheduler.jobs_to_run.get(block=True)) and self._can_execute():
            try:
                self._execute_job(job)
            except:  # In case the last job of the queue has been removed.
                self.__logger.warning(f"{job.id} is no longer in the list of jobs to run.")
```

src/taipy/core/_scheduler/_dispatcher/_standalone_job_dispatcher.py
toan-quach: yes you're right :D

src/taipy/core/_scheduler/_dispatcher/_job_dispatcher.py
toan-quach: @jrobinAV so at first, i have only `while True` to keep the loop going, then I added `STOP_FLAG` to make sure we can manually stop the thread. I want to keep this part (running the loop) and the condition of dispatching a job or not separated for readability. Since the addition of `STOP_FLAG`, the while will now be only `while STOP_FLAG` but I'd still rather keep the condition `if self._can_execute() and (job := self.scheduler.jobs_to_run.get(block=True)):` separated as above. It also allows us to make use of sleep to limit the CPU usage and the try catch would make more sense too since the try-catch is for `self.scheduler.jobs_to_run.get`

jrobinAV: OK that s good for me.

src/taipy/core/_scheduler/_dispatcher/_standalone_job_dispatcher.py
```diff
@@ -9,22 +9,24 @@
...
 from ._job_dispatcher import _JobDispatcher
 
 
 class _StandaloneJobDispatcher(_JobDispatcher):
     """Manages job dispatching (instances of `Job^` class) in an asynchronous way using a ProcessPoolExecutor."""
 
-    def __init__(self):
-        super().__init__()
-        self._executor = ProcessPoolExecutor(Config.job_config.nb_of_workers or 1)
+    def __init__(self, scheduler: _AbstractScheduler):
```
jrobinAV, line 26: Side remark for the futur: 
The dispatcher does not really need a scheduler. It only needs the queue. For now, it is good as it is, but in the future, we may want to remove the scheduler dependency here, and directly pass a queue object as a parameter of the init method. It could be either an in-memory queue like today or a more complex queue object using for instance RabbitMq.

jrobinAV (dismissed):
Looks good to me.
I believe we can merge it and start testing it deeply in various context.

jrobinAV (dismissed):
GReat !

src/taipy/core/_scheduler/_dispatcher/_development_job_dispatcher.py
```diff
@@ -9,14 +9,24 @@
...
 
 
 class _DevelopmentJobDispatcher(_StandaloneJobDispatcher):
     """Manages job dispatching (instances of `Job^` class) in a synchronous way."""
 
-    def __init__(self):
-        super().__init__()
-        self._executor = _Synchronous()
+    def __init__(self, scheduler: _AbstractScheduler):
+        super().__init__(scheduler)
```
florian-vuillemot, line 21: ?

toan-quach: I'm passing the scheduler to the parent level where it will be stored in self.scheduler

src/taipy/core/_scheduler/_scheduler_factory.py
```diff
@@ -14,27 +14,81 @@
...
-        return scheduler  # type: ignore
+
+    @classmethod
+    def __build_standalone_job_dispatcher(cls, force_restart=False):
+        if isinstance(cls._dispatcher, _StandaloneJobDispatcher) and not isinstance(
+            cls._dispatcher, _DevelopmentJobDispatcher
+        ):
+            if force_restart:
+                cls._dispatcher.stop()
+                cls._dispatcher.join()  # wait for thread to be terminated TODO: investigate the effect
```
florian-vuillemot: If a thread is blocked, this will wait forever.

src/taipy/core/_scheduler/_scheduler_factory.py
```diff
@@ -14,27 +14,81 @@
...
+            else:
+                return
+        cls._dispatcher = _StandaloneJobDispatcher(cls._scheduler)
+        cls._dispatcher.start()
+
+    @classmethod
+    def __build_development_job_dispatcher(cls):
+        if isinstance(cls._dispatcher, _StandaloneJobDispatcher) and not isinstance(
+            cls._dispatcher, _DevelopmentJobDispatcher
+        ):
```
florian-vuillemot, line 81: Another choice then Standalone or Development?

src/taipy/core/_scheduler/_scheduler_factory.py
```diff
@@ -14,27 +14,81 @@
...
+            )
+        else:
+            cls._scheduler = _Scheduler
+
+        # Build dispatcher
+        if not cls._dispatcher:
+            cls._build_dispatcher()
+
+        # Initialize the scheduler
+        cls._scheduler.initialize()  # type: ignore
```
florian-vuillemot: These comments are not relevant ;)

toan-quach: Oh yes agree!

src/taipy/core/_scheduler/_scheduler_factory.py
toan-quach: Nope. At least not at the moment. There's also Airflow i think

src/taipy/core/_scheduler/_scheduler_factory.py
toan-quach: hmm agree, I think I had this issue during testing. I will remove the join()

src/taipy/core/_scheduler/_scheduler.py
```diff
@@ -262,10 +194,19 @@ def __remove_jobs_to_run(cls, jobs):
...
                 job.cancelled()
             else:
                 job.abandoned()
+
+    @staticmethod
+    def __check_and_execute_jobs_if_development_mode():
+        if Config.job_config.is_development:
+            from ._scheduler_factory import _SchedulerFactory
+
+            _SchedulerFactory._get_dispatcher()._execute_jobs_synchronously()
```
florian-vuillemot, line 212: Can we move this method to the development dispatcher directly?
My point is to not bother the Scheduler with this implementation constraint.

toan-quach: we still need the scheduler to be the one instantiating the execution. I think we can move the method but we still need to import and use it in the scheduler which will involve duplicating the import code (the reason this function was written in the beginning)

src/taipy/core/_scheduler/_scheduler_factory.py
florian-vuillemot: woo if there is Airflow, please remove it!  The Airflow Scheduler is not working anymore ^^
@jrobinAV

toan-quach: Ah yes I believe it has been removed, I was just raising a point that there might be airflow in the future. As for now, we only have 2 :D

src/taipy/core/_scheduler/_scheduler.py
jrobinAV: I agree that it is strange to have a method in scheduler only useful when it is a DevelopmentScheduler.

jrobinAV (approved):

PR #270, opened 2022-07-26
Fix - Error while deleting entities

khaitruong922:
The bug is fixed after renaming the function `__get_model_filepath` to `_get_model_filepath`. The question now is: Should we do the same for all the other methods in the fs repository?


jrobinAV:
I am afraid because we did not really understand the issue. Why can't we use __get_model_filepath with 2 underscores?
Should we rename all private methods that have 2 underscores?

@khaitruong922 any idea or assumption on that?

jrobinAV (approved):

PR #271, opened 2022-07-27
Feature - Pass config to subprocess

khaitruong922:


jrobinAV (commented):
At this point, we don't know if your proposal is working. 
You removed the part of the unit test that was exhibiting the issue by making the test fail. We don't want to change the test, we want to change the code so the test is passing.

jrobinAV:
Plus the new code breaks 9 unit tests. All the tests that are running some tasks in standalone mode.

PR #278, opened 2022-08-03
bug/#274 improve job cancellation message

toan-quach:


src/taipy/core/__init__.py
```diff
@@ -18,6 +18,7 @@
 from .pipeline.pipeline import Pipeline
 from .scenario.scenario import Scenario
 from .taipy import (
+    cancel_job,
```
jrobinAV, line 21: well spotted !

jrobinAV (approved):

PR #281, opened 2022-08-05
Feature - Delete primary scenario

khaitruong922:
Allow delete primary scenario when it is the only scenario in the cycle

### DIFF
--- a/src/taipy/core/scenario/_scenario_manager.py
+++ b/src/taipy/core/scenario/_scenario_manager.py
@@ -203,8 +203,13 @@ class _ScenarioManager(_Manager[Scenario]):
 
     @classmethod
     def _delete(cls, scenario_id: ScenarioId):  # type: ignore
-        if cls._get(scenario_id).is_primary:
-            raise DeletingPrimaryScenario
+        scenario = cls._get(scenario_id)
+        if scenario.is_primary:
+            scenarios = cls._get_all_by_cycle(scenario.cycle)
+            if len(scenarios) > 1:
+                raise DeletingPrimaryScenario
+            _CycleManagerFactory._build_manager()._hard_delete(scenario.cycle.id)
+            return
         super()._delete(scenario_id)
 
     @classmethod
--- a/src/taipy/core/scenario/_scenario_manager.py
+++ b/src/taipy/core/scenario/_scenario_manager.py
@@ -208,8 +208,6 @@ class _ScenarioManager(_Manager[Scenario]):
             if len(cls._get_all_by_cycle(scenario.cycle)) > 1:
                 raise DeletingPrimaryScenario
             _CycleManagerFactory._build_manager()._delete(scenario.cycle.id)
-            super()._delete(scenario_id)
-            return
         super()._delete(scenario_id)
 
     @classmethod
--- a/src/taipy/core/scenario/_scenario_manager.py
+++ b/src/taipy/core/scenario/_scenario_manager.py
@@ -205,10 +205,10 @@ class _ScenarioManager(_Manager[Scenario]):
     def _delete(cls, scenario_id: ScenarioId):  # type: ignore
         scenario = cls._get(scenario_id)
         if scenario.is_primary:
-            scenarios = cls._get_all_by_cycle(scenario.cycle)
-            if len(scenarios) > 1:
+            if len(cls._get_all_by_cycle(scenario.cycle)) > 1:
                 raise DeletingPrimaryScenario
-            _CycleManagerFactory._build_manager()._hard_delete(scenario.cycle.id)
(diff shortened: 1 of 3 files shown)


jrobinAV (dismissed):
Great !

Maybe you can update the docstring (Ref man documentation) of taipy.delete method. For example, you can add a note to explain how the primary scenario deletion works, like you did in the user man.

src/taipy/core/scenario/_scenario_manager.py
```diff
@@ -203,8 +203,13 @@ def _untag(cls, scenario: Scenario, tag: str):
 
     @classmethod
     def _delete(cls, scenario_id: ScenarioId):  # type: ignore
-        if cls._get(scenario_id).is_primary:
-            raise DeletingPrimaryScenario
+        scenario = cls._get(scenario_id)
+        if scenario.is_primary:
+            if len(cls._get_all_by_cycle(scenario.cycle)) > 1:
```
toan-quach, line 208: I think you can wrap this as another function (delete_primary_scenario) to reduce duplicate code with the lines in _hard_delete

src/taipy/core/scenario/_scenario_manager.py
```diff
@@ -203,8 +203,13 @@ def _untag(cls, scenario: Scenario, tag: str):
...
+                raise DeletingPrimaryScenario
+            _CycleManagerFactory._build_manager()._delete(scenario.cycle.id)
+            super()._delete(scenario_id)
```
toan-quach: I think this line is not needed due to the line below right?

src/taipy/core/scenario/_scenario_manager.py
```diff
@@ -244,7 +249,10 @@ def __get_config(scenario: Scenario):
     def _hard_delete(cls, scenario_id: ScenarioId):
         scenario = cls._get(scenario_id)
         if scenario.is_primary:
-            raise DeletingPrimaryScenario
+            if len(cls._get_all_by_cycle(scenario.cycle)) > 1:
```
toan-quach, line 250: can wrap this in a function (delete_primary_scenario ) to improve readability and reduce codes :D

src/taipy/core/scenario/_scenario_manager.py
```diff
@@ -244,7 +249,10 @@ def __get_config(scenario: Scenario):
...
+                raise DeletingPrimaryScenario
+            _CycleManagerFactory._build_manager()._hard_delete(scenario.cycle.id)
+            return
```
toan-quach, line 253: what about using else here? :D

tests/core/scenario/test_scenario_manager.py
```diff
@@ -203,10 +203,14 @@ def test_create_and_delete_scenario():
     assert scenario_1.properties["name"] == name_1
     assert scenario_1.tags == set()
 
-    with pytest.raises(DeletingPrimaryScenario):
-        _ScenarioManager._delete(
-            scenario_1.id,
-        )
+    cycle_id_1 = scenario_1.cycle.id
+    assert _CycleManager._get(cycle_id_1).id == cycle_id_1
```
toan-quach, line 207: can you add another test about raising DeletingPrimaryScenario when the cycle has more than 1 scenario?

khaitruong922: It is already in line 230.

src/taipy/core/scenario/_scenario_manager.py
khaitruong922: I would prefer avoid using else since we already raise the exception here.

src/taipy/core/scenario/_scenario_manager.py
khaitruong922: We have different logic in delete and hard_delete so it does not really reduce code to refactor this.

src/taipy/core/scenario/_scenario_manager.py
toan-quach: raise an error will prevent you from running the rest so I'm not sure how it will affect it, and the else is better in meaning :/ but sure I'm ok with it :D

tests/core/scenario/test_scenario_manager.py
toan-quach: Ah! Saw it thanks!

src/taipy/core/scenario/_scenario_manager.py
toan-quach: Ahha you're right!

toan-quach (dismissed):

toan-quach (approved):

jrobinAV (approved):

PR #288, opened 2022-08-12
Fix - Read/write data node after set new path

khaitruong922:
The read/write method of data node which has `path` attribute (e.g: CSV, JSON, etc.) uses the `_path` value, which was not updated when setting `path`. The read/write method does not trigger data node reload, so the old path is still used. This change will ensure that the method will read/write from the set path.

Since this is a critical bug, it should be merged in the current release as well.


src/taipy/core/data/csv.py
```diff
@@ -109,6 +109,7 @@ def path(self):
 
     @path.setter
     def path(self, value):
+        self._path = value
         self.properties[self.__PATH_KEY] = value
```
jrobinAV, line 113: Indeed we have two ways of storing the path: 
- as self._path
- as self.properties[self.__PATH_KEY]
But only one way was updated. Well spotted!

The fix you proposed is to update both of them. Instead, I would only use a single way of storing it (self.properties[self.__PATH_KEY]). I would completely remove self._path.

DOes it make sense to you? Do I miss something here or do we just need one storage?

jrobinAV: That means we can have in the __init__ method : 

```
path = properties.get(self.__PATH_KEY, properties.get(self.__DEFAULT_PATH_KEY))
if path is None:
    raise MissingRequiredProperty("A default_path property is required in a CSV data node config.")
else:
    properties[self.__PATH_KEY] = path
```

jrobinAV: In the reload property we could have:

```
def path(self):
        return self.properties[self.__PATH_KEY]
```

jrobinAV (approved):
The Unit tests are great. They prove you fixed the bug. 
I believe we can improve the fix by removing one of the 2 ways to store the path. The Unit Tests should still work as today.

PR #291, opened 2022-08-17
Feature/#76 integrate new changes from taipy core

toan-quach:


src/taipy/core/_scheduler/_scheduler_factory.py
```diff
@@ -24,15 +24,18 @@
 class _SchedulerFactory:
 
     _TAIPY_ENTERPRISE_MODULE = "taipy.enterprise"
-    _TAIPY_ENTERPRISE_CORE_MODULE = _TAIPY_ENTERPRISE_MODULE + ".core._scheduler._scheduler"
+    _TAIPY_ENTERPRISE_CORE_SCHEDULER_MODULE = _TAIPY_ENTERPRISE_MODULE + ".core._scheduler._scheduler"
```
FabienLelaquais, line 27: Can't you use the __module__ name here?
Or would that be overkill..?

FabienLelaquais (dismissed):
👍

khaitruong922 (dismissed):

src/taipy/core/_scheduler/_scheduler_factory.py
toan-quach: I'm not sure what you mean by that :/

FabienLelaquais: I was thinking you might want to use the \_\_module\_\_ name instead of raw strings.
But because you are targeting Enterprise, this may not work.
The objective is to accelerate refactoring when it comes.

Now I don't really get how come you have a reference (even symbolic) to Enterprise from Core...

toan-quach: oh okay. So we can't use __module__ because we weren't sure if taipy-enterprise exists in the user's environment or not so we can't do something like`from taipy.enterprise import ...` . So we need something else that allows use to check if taipy-enterprise is installed in the user environment :D as for the reference to enterprise, we use the `importlib` module, provide it with the name of the package that we want to installed in the from of a string, and it does the rest of the work for us. Of course, when refactoring we will have to take the string part into account as well and have to solve it manually.

jrobinAV (dismissed):

src/taipy/core/_scheduler/_scheduler_factory.py
FabienLelaquais: Hum. Ok then.

jrobinAV (approved):

PR #292, opened 2022-08-17
Fix - Message for NonExistingCycle

khaitruong922:
taipy-rest uses `message` attribute of the exception to return the JSON response, and the NonExistingCycle does not have `message` yet.

### DIFF
--- a/src/taipy/core/exceptions/exceptions.py
+++ b/src/taipy/core/exceptions/exceptions.py
@@ -15,7 +15,7 @@ class CycleAlreadyExists(Exception):
 
 
 class NonExistingCycle(Exception):
-    """Raised if we request a cycle not known by the cycle manager."""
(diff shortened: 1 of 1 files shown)


src/taipy/core/exceptions/exceptions.py
```diff
@@ -17,6 +17,9 @@ class CycleAlreadyExists(Exception):
 class NonExistingCycle(Exception):
     """Raised if we request a cycle not known by the cycle manager."""
```
FabienLelaquais: "if we request a cycle"
->
"if a requested cycle"

FabienLelaquais (dismissed):
Minor rephrasing. No 'we'...

FabienLelaquais (approved):
👌

jrobinAV (approved):

Issue #276, opened 2022-08-01
Create a synchronous submit method in the scheduler


jrobinAV:
In development mode, the synchronous submit should do the same as the submit method.
In standalone mode, the synchronous submit should block until all the submitted jobs are finished (completed, failed, abandoned, or canceled)

toan-quach:
@jrobinAV I believe this ticket has already been implemented in the current version of scheduler and dispatcher. We can improve it by some further refactoring between development mode and standalone mode. But the general idea is already there

jrobinAV:
Not really, in standalone mode, the submit is always asynchronous. I would like to have a synchronous_submit method. It could simply delegate the submission to the usual asynchronous way, and wait for all jobs to be finished before to return.

toan-quach:
Ah I see what you mean. So we will have another synchronous_submit but for standalone mode where it will wait for all the submitted jobs to be finished before actually do the returning

jrobinAV:
Yes. absolutely

jrobinAV:
An alternative to the new method could also be to reuse the same submit method. It could take a new `sync` boolean parameter with a default value equals to False (asynchronous). When we set the parameter to True, it becomes synchronous.

PR #295, opened 2022-08-19
Feature/#276 synchronous submit method for async dispatcher

toan-quach:
#276

### DIFF
--- a/src/taipy/core/_scheduler/_scheduler.py
+++ b/src/taipy/core/_scheduler/_scheduler.py
@@ -140,12 +140,12 @@ class _Scheduler(_AbstractScheduler):
         start = datetime.now()
         jobs = jobs if isinstance(jobs, Iterable) else [jobs]
 
-        while __check_if_timeout(start, timeout):
+        while __check_if_timeout(start, timeout) and jobs:
             try:
-                if all([job._is_finished() for job in jobs]):
-                    return
-                sleep(0.1)  # Limit CPU usage
-            except:
+                if jobs[0]._is_finished():
+                    jobs.pop(0)
+                sleep(0.5)  # Limit CPU usage
+            except Exception:
                 pass
 
     @staticmethod
--- a/src/taipy/core/_scheduler/_scheduler.py
+++ b/src/taipy/core/_scheduler/_scheduler.py
@@ -52,7 +52,7 @@ class _Scheduler(_AbstractScheduler):
         callbacks: Optional[Iterable[Callable]] = None,
         force: bool = False,
         wait: bool = False,
-        timeout: Optional[int] = None,
+        timeout: Optional[Union[float, int]] = None,
     ) -> List[Job]:
         """Submit the given `Pipeline^` for an execution.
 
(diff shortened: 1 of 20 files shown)


src/taipy/core/_scheduler/_scheduler.py
```diff
@@ -108,6 +130,24 @@ def _schedule_job_to_run_or_block(cls, job: Job):
...
+            return True
+
+        start = datetime.now()
+        jobs = jobs if isinstance(jobs, Iterable) else [jobs]
+
+        while __check_if_timeout(start, timeout):
+            try:
+                if all([job._is_finished() for job in jobs]):
+                    return
+                sleep(0.1)  # Limit CPU usage
```
jrobinAV: Here you are testing all jobs all the time.
What you could do is: 
- test jobs one by one in a loop
- pop out a finished job
- break the loop once a job is not finished (without passing to the next ones) 
- restart the loop with only the remaining jobs.

Maybe this can reduce the collision probability with the execution of the job itself.

You can also sleep one second instead of 0.1 second.

This is a proposal for a temporary workaround.

jrobinAV (dismissed):
Don't forget to update the user man in taipy-doc and the release notes.

src/taipy/core/_scheduler/_scheduler.py
toan-quach: I think 1 second is long enough to make the user feels conscious about the wait time, I'll lower it down to 0.5s, what do you think :D
Also I'm not sure I understand your 3rd point. Did you mean once a job is not finished and is timed out?

jrobinAV: On the third point, I mean we should continue, not break. Sorry. if a job is not finished, we don't need to check other jobs. We can go to the next iteration.

src/taipy/core/_scheduler/_scheduler.py
```diff
@@ -108,6 +130,24 @@ def _schedule_job_to_run_or_block(cls, job: Job):
...
+            return True
+
+        start = datetime.now()
+        jobs = jobs if isinstance(jobs, Iterable) else [jobs]
+
+        while __check_if_timeout(start, timeout) and jobs:
+            try:
+                if jobs[0]._is_finished():
+                    jobs.pop(0)
+                sleep(0.5)  # Limit CPU usage
```
jrobinAV:
```suggestion
                if jobs[0]._is_finished():
                    jobs.pop(0)
                else:
                    sleep(0.5)  # Limit CPU usage
                
```
if the first job to check is finished, we don't need to sleep before trying the next job. (that s basically the third point of my previous comment)

jrobinAV (dismissed):

src/taipy/core/_scheduler/_scheduler.py
toan-quach: ah ha! yep that makes much more sense! :D

src/taipy/core/_scheduler/_scheduler.py
```diff
@@ -86,10 +105,13 @@ def submit_task(
         for dn in task.output.values():
             dn.lock_edit()
         job = _JobManagerFactory._build_manager()._create(
-            task, itertools.chain([cls._on_status_change], callbacks or []), submit_id
+            task, itertools.chain([cls._on_status_change], callbacks or []), submit_id, force=force
```
jrobinAV, line 130: Well spotted !

If we did not spot this bug before that means we don't have unit tests on it. Can we create one?

jrobinAV (dismissed):

src/taipy/core/_scheduler/_scheduler.py
toan-quach: sure I'll have a look :D

jrobinAV (approved):

Issue #247, opened 2022-06-20
handle blocked jobs by failing job


jrobinAV:
When submitting a pipeline, some jobs are blocked until the first job is complete. If the first job fails, the subsequent jobs remain blocked for an undefined time. 
The purpose of the ticket is to propose a method to abandon subsequent jobs that remain blocked by a failing job.

** Expected outcome:**
- [x] A design or presentation on how to implement the feature.
- [x] A first implementation
- [ ] Documentation on user manual and ref manual
- [ ] Add entry in 1.2 release notes
- [ ] A milestone demonstration

PR #297, opened 2022-08-24
feature#247 added abandon subsequent jobs of fail jobs

toan-quach:
#247

### DIFF
--- a/src/taipy/core/_scheduler/_scheduler.py
+++ b/src/taipy/core/_scheduler/_scheduler.py
@@ -18,6 +18,7 @@ from typing import Callable, Iterable, List, Optional, Set, Union
 from taipy.config.config import Config
 from taipy.logger._taipy_logger import _TaipyLogger
 
+from ..common.alias import JobId
 from ..data._data_manager_factory import _DataManagerFactory
 from ..job._job_manager_factory import _JobManagerFactory
 from ..job.job import Job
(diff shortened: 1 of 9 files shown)


src/taipy/core/_scheduler/_scheduler.py
```diff
@@ -200,11 +209,22 @@ def __remove_jobs_to_run(cls, jobs):
         cls.jobs_to_run = new_jobs_to_run
 
     @classmethod
-    def _cancel_jobs(cls, job_id_to_cancel, jobs):
+    def _fail_subsequent_jobs(cls, failed_job: Job):
+        with cls.lock:
+            to_fail_jobs = set()
```
jrobinAV: I would rename it to 'to_abandon_jobs'

src/taipy/core/_scheduler/_scheduler.py
```diff
@@ -131,11 +141,10 @@ def _unlock_edit_on_outputs(jobs: Union[Job, List[Job], Set[Job]]):
...
     @classmethod
     def _on_status_change(cls, job: Job):
-        if job.is_completed():
+        if job.is_completed() or job.is_skipped():
             cls.__unblock_jobs()
-            cls.__check_and_execute_jobs_if_development_mode()
-        elif job.is_canceled():
-            cls.__check_and_execute_jobs_if_development_mode()
+        elif job.is_failed():
+            cls._fail_subsequent_jobs(job)
```
jrobinAV, line 153: When the job fails, I don't think we unlock the outputs. Do we?
Do we have a test that check that asserts that after a job fails its outputs are unlocked?

src/taipy/core/_scheduler/_scheduler.py
toan-quach: I'll also add the failed_jobs to the set and rename it to to_fail_or_abandon_jobs :D

src/taipy/core/_scheduler/_scheduler.py
toan-quach: As we discussed earlier, we're only unlocking the abandoned jobs and not the failed job. I've added some manipulation to the set to_fail_or_cancel_jobs to make sure we also unlock the failed job

jrobinAV (approved):

PR #298, opened 2022-08-24
Fix - SQLDataNode write

khaitruong922:
Problems:
- Insert list/tuple: Incorrect SQL syntax
- Insert dict: Table columns are not automatically loaded so it cannot map the values to correct column, so null is inserted everywhere.

Fix:
- Table autoload: Now load all table columns after connected to database so insert_dicts will work.
- Insert list/tuple: Fix SQL syntax, remove insert_lists method and convert lists to tuples instead.

Improvements:
- Refactor _write logic, convert everything to a list first if it is not a list (except pd.DataFrame).

### DIFF
--- a/src/taipy/core/data/sql.py
+++ b/src/taipy/core/data/sql.py
@@ -15,6 +15,7 @@ import urllib.parse
 from datetime import datetime, timedelta
 from typing import Any, Dict, List, Optional, Sequence, Tuple
 
+import numpy as np
 import pandas as pd
 from sqlalchemy import MetaData, Table, create_engine, text
 
@@ -181,6 +182,9 @@ class SQLDataNode(DataNode):
             if isinstance(data, pd.DataFrame):
                 self._insert_dicts(data.to_dict(orient="records"), table, connection)
                 return
+
+            if isinstance(data, np.ndarray):
+                data = data.tolist()
 
             if not isinstance(data, list):
                 data = [data]
--- a/src/taipy/core/data/sql.py
+++ b/src/taipy/core/data/sql.py
@@ -164,29 +164,37 @@ class SQLDataNode(DataNode):
             return pd.read_sql_query(self.read_query, con=self.__engine())[columns]
         return pd.read_sql_query(self.read_query, con=self.__engine())
 
+    def _create_table(self, engine) -> Table:
+        return Table(
+            self.write_table,
+            MetaData(),
+            autoload=True,
+            autoload_with=engine,
+        )
+
     def _write(self, data) -> None:
         """Check data against a collection of types to handle insertion on the database."""
         engine = self.__engine()
         with engine.connect() as connection:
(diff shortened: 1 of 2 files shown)


khaitruong922:
The tests are failing because of mocking issues. The Table object cannot use the mock Engine object.

src/taipy/core/data/sql.py
```diff
@@ -157,53 +158,30 @@ def _read_as_pandas_dataframe(self, columns: Optional[List[str]] = None):
 
     def _write(self, data) -> None:
         """Check data against a collection of types to handle insertion on the database."""
-        with self.__engine().connect() as connection:
-            write_table = table(self.write_table)
+        engine = self.__engine()
+        with engine.connect() as connection:
+            write_table = Table(self.write_table, MetaData(), autoload=True, autoload_with=engine)
```
khaitruong922: The tests are failing because of mocking issues. The Table object cannot use the mock Engine object.

jrobinAV (commented):
Sorry, but I don't understand the fix. Can you write a unit test that shows the write methods work in all the cases (tuple, list, dict, dataframe, np.array) with zero, one or multiple entries?

src/taipy/core/data/sql.py
```diff
@@ -165,57 +165,41 @@ def _read_as_pandas_dataframe(self, columns: Optional[List[str]] = None):
...
-                raise
+        engine = self.__engine()
+
+        if isinstance(data, pd.DataFrame):
+            data = data.to_dict(orient="records")
+        elif isinstance(data, np.ndarray):
+            data = data.tolist()
+        if not isinstance(data, list):
+            data = [data]
+        if len(data) == 0:
```
joaoandre-avaiga, line 186: If data is None, this check will pass and we will write in the table. Is that the desired behavior?

jrobinAV (approved):
I believe it s correct.
I am just concerned about the data conversion : 
```
if isinstance(data, pd.DataFrame):
    data = data.to_dict(orient="records")
elif isinstance(data, np.ndarray):
    data = data.tolist()
if not isinstance(data, list):
    data = [data]
```
data could represent a large dataset. So any conversion could take time.
Is it necessary to convert a dataframe or a ndarray into a list or dict before inserting it or is there a direct insertion method?

Can you run your SQL performance evaluation script on it to measure the impact of such a change?

src/taipy/core/data/sql.py
khaitruong922: Yes, it is inserting 1 row of (null) into the table.

khaitruong922:
The dataframe conversion is already there before so nothing really changes. We can use the `to_sql` of dataframe https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.to_sql.html if it has better performance. 

Let's merge this to fix the current issue first.

PR #300, opened 2022-08-30
Fix -  KeyError in ExcelDataNode

khaitruong922:
The function __exposed_types_to_dict uses properties["path"], which is not set if the user provides "default_path". Changing the code execution order fixes this.


toan-quach (approved):
lgtm

jrobinAV (approved):
Well spotted !

PR #304, opened 2022-08-31
Feature - Overwrite table SQLDataNode

khaitruong922:


jrobinAV (changes requested):
With your fix, we solve the first consistency issue (drop the database before to insert entries).
But it does not fix the consistency issue on the reading and writing queries.

When reading the data node, we want to be able to retrieve the written data.

I commented the ticket https://app.zenhub.com/workspaces/taipy-613ef328affde6000ff77fb3/issues/avaiga/taipy-core/294 to provide more details on specifications.

jrobinAV (approved):

PR #305, opened 2022-08-31
Fix - ExcelDataNode exposed_type + read sheet names on demand.

khaitruong922:
Only read the sheet names when the data node is read.

Add serialization and deserialization for list as exposed_type .

Keep the exposed_type config value.
- If it is a custom class, then applied that exposed type to every sheet. Does not matter if the set of sheets are different (in terms of length or names) after we change the path.
- If it is a list, then apply each custom class in order. Does not matter if the sheet names are different after we change the path,  as long as its length and the number of sheets matches

Fix an issue where reading excel as dataframe when `sheet_names` is provided as a string, which returns a single dataframe instead of dictionary, and calling items() on that dataframe return inapproriate result.

Change default exposed_type to "pandas" instead of None.


src/taipy/core/config/checkers/_data_node_config_checker.py
```diff
@@ -103,3 +104,13 @@ def _check_generic_read_write_fct(self, data_node_config_id: str, data_node_conf
...
                     )
+
+    def _check_exposed_type(self, data_node_config_id: str, data_node_config: DataNodeConfig):
+        if not isinstance(data_node_config.exposed_type, str):
+            return
+        if data_node_config.exposed_type not in DataNodeConfig._ALL_EXPOSED_TYPES:
+            self._error(
+                data_node_config._EXPOSED_TYPE_KEY,
+                data_node_config.exposed_type,
+                f'`{data_node_config._EXPOSED_TYPE_KEY}` of DataNode `{data_node_config_id}` must be either "pandas", "numpy" or a type.',
```
jrobinAV:
```suggestion
                f'The `{data_node_config._EXPOSED_TYPE_KEY}` of the DataNodeConfig `{data_node_config_id}` must be either "pandas", "numpy" or a custom type.',
```

src/taipy/core/config/data_node_config.py
```diff
@@ -247,13 +257,21 @@ def _configure(id: str, storage_type: str = _DEFAULT_STORAGE_TYPE, scope: Scope
...
     @staticmethod
-    def _configure_csv(id: str, default_path: str = None, has_header: bool = True, scope=_DEFAULT_SCOPE, **properties):
+    def _configure_csv(
+        id: str,
+        default_path: str = None,
+        has_header: bool = True,
+        exposed_type=_EXPOSED_TYPE_PANDAS,
+        scope=_DEFAULT_SCOPE,
+        **properties,
+    ):
```
jrobinAV, line 267: I like it.

src/taipy/core/config/data_node_config.py
```diff
@@ -247,13 +257,21 @@ def _configure(id: str, storage_type: str = _DEFAULT_STORAGE_TYPE, scope: Scope
...
         """Configure a new CSV data node configuration.
 
         Parameters:
             id (str): The unique identifier of the new CSV data node configuration.
             default_path (str): The default path of the CSV file.
             has_header (bool): If True, indicates that the CSV file has a header.
+            exposed_type: The exposed type of the data read from CSV file.
```
```suggestion
            exposed_type: The exposed type of the data read from CSV file. The default value is `pandas`.
```

src/taipy/core/config/data_node_config.py
```diff
@@ -324,6 +344,7 @@ def _configure_excel(
             has_header (bool): If True, indicates that the Excel file has a header.
             sheet_name (Union[List[str], str]): The list of sheet names to be used. This
                 can be a unique name.
+            exposed_type: The exposed type of the data read from Excel file.
```
```suggestion
            exposed_type: The exposed type of the data read from Excel file. The default value is `pandas`.
```

src/taipy/core/config/data_node_config.py
```diff
@@ -459,6 +482,7 @@ def _configure_sql(
                 _"ODBC Driver 17 for SQL Server"_.
             db_extra_args (Dict[str, Any]): A dictionary of additional arguments to be passed into database
                 connection string.
+            exposed_type: The exposed type of the data read from SQL query.
```
```suggestion
            exposed_type: The exposed type of the data read from SQL query. The default value is `pandas`.
```

tests/core/data/test_excel_data_node.py
```diff
@@ -689,3 +693,170 @@ def test_read_write_after_modify_path(self):
         dn.write(read_data)
         for sheet, df in dn.read().items():
             assert np.array_equal(df.values, read_data[sheet].values)
+
+    def test_exposed_type_custom_class_after_modify_path(self):
```
jrobinAV, line 697: Please find a better name. We don't really test the path setting, but we test an NonExistingExcelSheet exception is raised when a wring sheet name is provided.

tests/core/data/test_excel_data_node.py
```diff
@@ -689,3 +693,170 @@ def test_read_write_after_modify_path(self):
...
+                "exposed_type": {
+                    "Sheet1": MyCustomObject1,
+                    "Sheet2": MyCustomObject1,
+                    "Sheet3": MyCustomObject1,
+                },
+            },
+        )
+        dn.read()
+        dn.path = new_path
+        dn.read()
```
jrobinAV: please add some assertions to show that it is working with path. 
Plus should we expect an exception with new_path since sheet3 does not exist?

tests/core/data/test_excel_data_node.py
```diff
@@ -689,3 +693,170 @@ def test_read_write_after_modify_path(self):
...
+
+        dn = ExcelDataNode(
+            "foo",
+            Scope.PIPELINE,
+            properties={"default_path": path, "exposed_type": MyCustomObject1, "sheet_name": ["Sheet4"]},
+        )
+        assert dn.exposed_type == MyCustomObject1
```
jrobinAV, line 713: Please add some code to show that it is working well when we provide an good existing sheet name.

tests/core/data/test_excel_data_node.py
```diff
@@ -689,3 +693,170 @@ def test_read_write_after_modify_path(self):
...
+                    "Sheet2": MyCustomObject1,
+                    "Sheet3": MyCustomObject1,
+                },
+            },
+        )
+        dn.read()
+        dn.path = new_path
+        dn.read()
+
+    def test_exposed_type_list_after_modify_path(self):
```
jrobinAV: Please find a better name. again we don't test path modification.

tests/core/data/test_excel_data_node.py
```diff
@@ -689,3 +693,170 @@ def test_read_write_after_modify_path(self):
...
+            pathlib.Path(__file__).parent.resolve(), "data_sample/example_3.xlsx"
+        )  # ["Sheet3", "Sheet4"]
+
+        dn = ExcelDataNode(
+            "foo", Scope.PIPELINE, properties={"default_path": path_1, "exposed_type": [MyCustomObject, MyCustomObject]}
+        )
+
+        dn.read()
+        dn.path = path_3
+        dn.read()
```
jrobinAV: I believe you don't need to test with 3 files. path_1 and path2 should be enough. (path_3 does not seem to be useful)

tests/core/data/test_excel_data_node.py
```diff
@@ -689,3 +693,170 @@ def test_read_write_after_modify_path(self):
...
+
+        dn.path = path_2
+        with pytest.raises(ExposedTypeLengthMismatch):
+            dn.read()
+
+    def test_not_trying_to_read_sheet_names_when_exposed_type_is_set(self):
+        ExcelDataNode(
+            "foo", Scope.PIPELINE, properties={"default_path": "notexistyet.xlsx", "exposed_type": MyCustomObject1}
+        )
+        ExcelDataNode(
```
jrobinAV: please add some asserts to check the various attributes of the datanode including an assertion on the path.

tests/core/data/test_excel_data_node.py
```diff
@@ -689,3 +693,170 @@ def test_read_write_after_modify_path(self):
...
+    def test_exposed_type_custom_class_after_modify_path(self):
+        path = os.path.join(pathlib.Path(__file__).parent.resolve(), "data_sample/example.xlsx")  # ["Sheet1", "Sheet2"]
+        new_path = os.path.join(
+            pathlib.Path(__file__).parent.resolve(), "data_sample/example_2.xlsx"
+        )  # ["Sheet1", "Sheet2", "Sheet3"]
+        dn = ExcelDataNode("foo", Scope.PIPELINE, properties={"default_path": path, "exposed_type": MyCustomObject1})
+        assert dn.exposed_type == MyCustomObject1
+        dn.read()
+        dn.path = new_path
+        dn.read()
```
jrobinAV, line 706: Please add some assertion to test that dn.read returned the expected value.
Plus I believe you don't need to test with a second path. Only one is enough.

tests/core/data/test_excel_data_node.py
```diff
@@ -689,3 +693,170 @@ def test_read_write_after_modify_path(self):
...
+        dn = ExcelDataNode(
+            "foo",
+            Scope.PIPELINE,
+            properties={"default_path": path, "exposed_type": MyCustomObject1, "sheet_name": ["Sheet4"]},
+        )
+        assert dn.exposed_type == MyCustomObject1
+        with pytest.raises(NonExistingExcelSheet):
+            dn.read()
+
+    def test_exposed_type_dict_after_modify_path(self):
```
jrobinAV: Please find a better name. You are not testing the path setting but the dict exposed type.

tests/core/data/test_sql_data_node.py
```diff
@@ -108,10 +109,10 @@ def test_read(self, mock_read_as, mock_read_as_pandas_dataframe, mock_read_as_nu
                 "db_engine": "mssql",
                 "read_query": "SELECT * from table_name",
```
jrobinAV, line 118: The comment on line 101 is not up to date anymore. Please update it, and maybe you can create a case to test with the custom expose_type

jrobinAV (commented):
Only minor remarks on the unit tests to make the test easier to read/maintain

tests/core/data/test_excel_data_node.py
khaitruong922: No because we only check the sheets in `sheet_names` attribute exist. This dictionary is just showing how to map each sheet to each type, and it does not matter if the sheet in the dictionary does not exist.

jrobinAV (approved):

PR #309, opened 2022-09-08
feature/#205 replace pandas with modin dataframe

toan-quach:
Issues:
- cannot load pickled modin dataframe 
- modin.read_excel() doesn't work with header=None (found a workaround, modin team also acknowledges this issue and is looking into it), many excel tests are also failing
- if the tests fail in test_complex.py, the test will also fail in test_csv_datanode.py, if we run without the test_complex.py, test_csv_dn is successful


src/taipy/core/data/data_node.py
```diff
@@ -16,8 +16,8 @@
 from functools import reduce
 from typing import Any, List, Optional, Set, Tuple, Union
 
+import modin.pandas as pd
```
jrobinAV: If we want to keep both exposed types (pandas and modin) I believe we should import modin as modin and pandas as pd. What do you think ?

jrobinAV (commented):
WE discussed the possibility of keeping the pandas implementation as well. Right? We can keep both exposed types: pandas and modin. So we could implement a new method read_as_modin_df in the csv and excel datanodes. 
Does it make sense to you?

src/taipy/core/_scheduler/_dispatcher/_development_job_dispatcher.py
```diff
@@ -44,4 +44,6 @@ def _dispatch(self, job: Job):
         config_as_string = _TomlSerializer()._serialize(Config._applied_config)
 
         rs = self._run_wrapped_function(Config.job_config.mode, config_as_string, job.id, job.task)
+        # breakpoint()
+        # TODO: the problem is here, trying to load the pickle of modin.dataframe fails
```
jrobinAV: To remove before merging

src/taipy/core/_scheduler/_dispatcher/_job_dispatcher.py
```diff
@@ -140,6 +140,7 @@ def _run_wrapped_function(cls, mode, config_as_string, job_id: JobId, task: Task
             outputs: List[DataNode] = list(task.output.values())
             fct = task.function
             results = fct(*cls.__read_inputs(inputs))
+            # breakpoint()
```
jrobinAV: todo remove before merge

jrobinAV (commented):
More remarks : 
1 - If we expose both types pandas and modin, we will need to duplicate most of the CSV, Excel, filter, Pickle unit tests.

2 - The pickle unit test should have detected the error. The test test_complex is an integration test that will be moved to another repo soon. This kind of bug must be detected by a simple unit test. (We just need to add a unit test to pickle a pandas dataframe and a modin data frame, now)
 
3 - After some investigation, it appears that dask modin dataframes are pickalizable. 
- Here is the [issue](https://github.com/modin-project/modin/issues/672) I found.
- here is the corresponding merged [PR](https://github.com/modin-project/modin/pull/2835), 
- and here is the [unit test](https://github.com/modin-project/modin/pull/2835/files#diff-d19d81ac1899c4ddc0a13a36c3b0c8acd507bd7ee369eba1031475e8ab6cc764) showing how to use it.
It seems we just need to set up an environment variable, but I could not make it work yet.

4 - More info on [experimental API for sql](https://modin.readthedocs.io/en/stable/flow/modin/experimental/pandas.html#modin.experimental.pandas.read_sql)

src/taipy/core/data/data_node.py
toan-quach: agree! I'll put it as something like:
```
import modin.pandas as modin_pd
import pandas as pd
```

jrobinAV: great

setup.py
```diff
@@ -22,6 +22,7 @@
     "networkx>=2.6,<3.0",
     "openpyxl>=3.0.7,<4.0",
     "modin[dask]>=0.16.2,<1.0",
+    "bokeh==2.4.3",
```
jrobinAV: hum?

Pipfile
```diff
@@ -11,6 +11,7 @@ toml = "==0.10"
 taipy-config = {git = "https://git@github.com/avaiga/taipy-config.git@develop"}
 pymongo = "==4.2.0"
 modin = {extras = ["dask"], version = "==0.16.2"}
+bokeh = "*"
```
jrobinAV: hum?

Pipfile
```diff
@@ -5,12 +5,13 @@ name = "pypi"
 
 [packages]
 networkx = "==2.6"
-openpyxl = "==3.0.3"
-pandas = "==1.3.4"
```
jrobinAV, line 9: We need to ensure changing the version of pandas is ok for taipy-gui as well. The pandas version pulled by modin 0.16.2 is pandas 1.5.1

@FabienLelaquais Can someone from your team check that it is OK?

setup.py
```diff
@@ -20,8 +20,10 @@
 
 requirements = [
     "networkx>=2.6,<3.0",
-    "openpyxl>=3.0.3,<4.0",
-    "pandas>=1.3.4,<2.0",
+    "openpyxl>=3.0.7,<4.0",
+    "modin[dask]>=0.16.2,<1.0",
+    "bokeh==2.4.3",
```
jrobinAV: why do we need bokeh?

tests/conftest.py
```diff
@@ -54,6 +55,7 @@
 from taipy.config.common.scope import Scope
 from taipy.config.config import Config
 
+client = Client()
```
jrobinAV: WHy do we need that in tests, but not in the "real life"?

tests/core/data/test_csv_data_node.py
```diff
@@ -90,13 +91,20 @@ def test_read_with_header(self):
             not_existing_csv.read_or_raise()
 
         path = os.path.join(pathlib.Path(__file__).parent.resolve(), "data_sample/example.csv")
-        # Create CSVDataNode without exposed_type (Default is pandas.DataFrame)
+        # # Create CSVDataNode without exposed_type (Default is pandas.DataFrame)
```
jrobinAV, line 94:
```suggestion
        # Create CSVDataNode without exposed_type (Default is pandas.DataFrame)
```

tests/core/data/test_data_node.py
```diff
@@ -287,6 +287,7 @@ def test_pandas_filter(self, default_data_frame):
         df_dn = FakeDataframeDataNode("fake_dataframe_dn", default_data_frame)
         COLUMN_NAME_1 = "a"
         COLUMN_NAME_2 = "b"
+        # breakpoint()
```
jrobinAV:
```suggestion
```

tests/core/data/test_csv_data_node.py
```diff
@@ -179,6 +196,29 @@ def test_write(self, csv_file, default_data_frame, content, columns):
...
+            ([[11, 22, 33], [44, 55, 66]], ["e", "f", "g"]),
+        ],
+    )
+    def test_write_modin(self, csv_file, default_data_frame, content, columns):
+        default_data_frame = modin_pd.DataFrame(default_data_frame)
+        csv_dn = CSVDataNode("foo", Scope.PIPELINE, properties={"path": csv_file, "exposed_type": "modin"})
+        assert np.array_equal(csv_dn.read().values, default_data_frame.values)
+        if not columns:
+            csv_dn.write(content)
+            df = pd.DataFrame(content)
```
jrobinAV, line 213: For consistency matter, it should use modin_pd:
```suggestion
            df = modin_pd.DataFrame(content)
```

tests/core/data/test_csv_data_node.py
```diff
@@ -179,6 +196,29 @@ def test_write(self, csv_file, default_data_frame, content, columns):
...
+        else:
+            csv_dn.write_with_column_names(content, columns)
+            df = pd.DataFrame(content, columns=columns)
```
jrobinAV, line 216:
```suggestion
            df = modin_pd.DataFrame(content, columns=columns)
```

tests/core/data/test_sql_data_node.py
```diff
@@ -12,7 +12,7 @@
 from importlib import util
 from unittest import mock
 
-import pandas as pd
+import modin.pandas as pd
```
jrobinAV: We are missing some tests here

tests/core/data/test_sql_table_data_node.py
```diff
@@ -56,7 +56,6 @@ class TestSQLTableDataNode:
                 },
```
jrobinAV, line 83: We are missing some tests in test_sql_table as well

tests/performance_testing/scenario-creation/config.py
```diff
@@ -11,8 +11,8 @@
 
 import datetime as dt
 
+import modin.pandas as pd
```
jrobinAV: I believe we should keep pandas as the default usage in this config.py file.

Other wise we should update the config to have both in parallel modin and pandas

tests/test_complex_application.py
```diff
@@ -13,7 +13,7 @@
 import pathlib
 from time import sleep
 
-import pandas as pd
+import modin.pandas as pd
```
jrobinAV: Same here, we should keep tests in pandas, and eventually add more tests in modin.

tests/test_complex_application.py
```diff
@@ -113,6 +113,7 @@ def test_complex():
     # |      |
     # |      |
     # t4     d4
+    os.environ["MODIN_PERSISTENT_PICKLE"] = "True"
```
jrobinAV: This should be set in conftest to be generalized. Same for unsettling the env variable.

jrobinAV (changes requested):
The Pickle data node should ensure the env variable is set before to read or write a modin df.

setup.py
toan-quach: I found out somehow we need bokeh for the distributed package used by modin, it's something to investigate next :D

setup.py
toan-quach: the distributed package used by modin needs it, I'm adding just to test it to see that it actually works and will investigate further

tests/conftest.py
toan-quach: Ah yes agree, at the moment I'm just using it to fix some failing tests, then will integrate it into the main code

tests/test_complex_application.py
toan-quach: I think this should be set in the pickle file, since the main code will potentially run into trouble without it, but there's some uncertainties at the moment so I'm currently taking it slow in making such move

tests/conftest.py
toan-quach: @jrobinAV regarding initializing the Client() object, I have found some references that can help us determine if we need to initialize it
https://stackoverflow.com/questions/61424359/do-i-need-to-initialize-the-client-when-using-dask-locally
https://docs.dask.org/en/latest/deploying-python.html
https://docs.dask.org/en/latest/scheduling.html
In my opinion, just remove this line. Previously this was added to remove the problem in the tests, but thanks to the current iteration of the PR, that's no longer the case. So I'd propose removing it and stick to the default setting of modin to keep things simpler for us. What do you think?

setup.py
toan-quach: @jrobinAV apparently after version 2022.10.1 dask and distributed packages needed bokeh to do some visualizations from within as a required dependency, but their setup.py file didn't specify such requirements, resulting in errors when trying to use modin with dask engine. I have digged into this and I believe it was a mistake on the dask team as they are making changes to keep bokeh as an optional dependency again (it was optional previously). More info can be found here: https://github.com/dask/distributed/commit/aa2cd65d84dad9ce58aab78c4f648d65c52ccc9b

toan-quach: apparently this has been fixed in 2022.10.2 (their latest version), I have removed bokeh from our dependencies list :D

jrobinAV: great !

setup.py
jrobinAV: You removed it from PipFile, but not setup.py

src/taipy/core/data/pickle.py
```diff
@@ -125,6 +127,8 @@ def _read(self):
         return pickle.load(open(self._path, "rb"))
 
     def _write(self, data):
+        if isinstance(data, (pd.DataFrame, pd.Series)):
```
jrobinAV, line 131: Should we set the env var when reading as well?

tests/test_complex_application.py
jrobinAV: Absolutely !

tests/conftest.py
jrobinAV: yes that s sounds great!

setup.py
toan-quach: Thanks! That was very careless of me 😅

src/taipy/core/data/pickle.py
toan-quach: I was thinking we don't have to since we don't what data is in the pickle file and if it is needed. But it does make sense that in the scenario where the data is a modin dataframe, we can safely unpickle it. I'll add the env in the read function 😄

jrobinAV (approved):

PR #310, opened 2022-09-08
Feature - SQLDataNode write query builder

khaitruong922:
See milestone 21 for implementation details: https://github.com/Avaiga/milestones/pull/82


src/taipy/core/config/data_node_config.py
```diff
@@ -456,8 +456,9 @@ def _configure_sql(
         db_password: str,
         db_name: str,
         db_engine: str,
-        read_query: str,
-        write_table: str = None,
+        table: str = None,
+        read_query: str = None,
+        write_query_factory: Callable = None,
```
jrobinAV: I don't like it. This makes the simple case too complex for the user.
As explained in the ticket [308](https://app.zenhub.com/workspaces/taipy-613ef328affde6000ff77fb3/issues/avaiga/taipy-core/308), I would like to have 2 methods : 
- `Config.configure_sql_table_data_node` that only requires an existing table_name.
- `Config.configure_sql_db_data_node` that requires the necessary parameters to build the two queries.
I strongly believe it is easier for the user to separate the 2 use cases. Either the data is stored in a table, or it is stored in a database.

Let me propose a way to have 2 different storage types without code duplication.

src/taipy/core/config/data_node_config.py
```diff
@@ -501,7 +502,8 @@ def _configure_sql(
             db_engine=db_engine,
             db_driver=db_driver,
             read_query=read_query,
-            write_table=write_table,
+            table=table,
+            write_query_factory=write_query_factory,
```
jrobinAV: I would rename the "factory" to "builder". In the end, it is not a real factory. It is just a function to build the queries.

jrobinAV (changes requested):
**Data node config:** I am going to push some changes to illustrate what I am expecting regarding the 2 use cases to differentiate for the user.

*_*unit tests_:** They are failing. I won't update them. Please update them so they are successful again.

**_config_ checkers:** If you have any idea about a new specific checker to check the `write_query_builder` function, do not hesitate to propose something.

I did not run the pre-commit hooks. Please run them.

Thank you,

src/taipy/core/config/data_node_config.py
```diff
@@ -468,20 +479,79 @@ def _configure_sql(
     ):
         """Configure a new SQL data node configuration.
```
toan-quach: Please update the description for both `_configure_sql_table` and `_configure_sql_table`, the docstring should at least gives a hint what's the difference between the 2 functions, what they're based on for example

src/taipy/core/config/data_node_config.py
toan-quach: @jrobinAV If I remember correctly, `configure_sql_table_data_node` is used for simple case where the user only needs to provide the table name to retrieve the data right while `configure_sql_db_data_node` allows user to have their own query. In that case, I'd propose we change the name of the function `configure_sql_db_data_node` to something else, as user might understand it that they only need to provide the database name (which is not correct and not a good idea as well). Something like `configure_sql_query_data_node` is better in my opinion

src/taipy/core/data/sql_db.py
```diff
@@ -0,0 +1,97 @@
...
+
+from taipy.config.common.scope import Scope
+
+from ..common.alias import DataNodeId, JobId
+from ..exceptions.exceptions import MissingRequiredProperty
+from .sql import SQLDataNode
+
+
+class SQLDbDataNode(SQLDataNode):
+    """Data Node stored in an SQL database.
```
toan-quach: "Data Node stored in an SQL database." => "Data Node stored as an SQL database." or something else that you see fit. It would also be great if you can point out the main difference between SQLDbDN and SQLTableDataNode.

src/taipy/core/data/sql_table.py
```diff
@@ -0,0 +1,152 @@
...
+
+from taipy.config.common.scope import Scope
+
+from ..common.alias import DataNodeId, JobId
+from ..exceptions.exceptions import MissingRequiredProperty
+from .sql import SQLDataNode
+
+
+class SQLTableDataNode(SQLDataNode):
+    """Data Node stored in an SQL table.
```
toan-quach: "Data Node stored in an SQL database." => "Data Node stored as an SQL database." or something else that you see fit. It would also be great if you can point out the main difference between SQLDbDN and SQLTableDataNode.

src/taipy/core/exceptions/exceptions.py
```diff
@@ -43,7 +43,7 @@ class NoData(Exception):
 
 
 class UnknownDatabaseEngine(Exception):
-    """Raised if the database engine is not known when creating a connection with a SQLDataNode."""
+    """Raised if the database engine is not known when creating a connection with an SQLDataNode."""
```
toan-quach: I noticed that we have some inconsistency of "a SQL..." or "an SQL..." throughout not just taipy-core, but also in taipy-doc (I haven't checked other repos). Please fix it while you're on the topic. Thanks!

src/taipy/core/config/data_node_config.py
```diff
@@ -468,20 +479,79 @@ def _configure_sql(
...
+        db_host: str = "localhost",
+        db_driver: str = "ODBC Driver 17 for SQL Server",
+        db_extra_args: Dict[str, Any] = None,
+        read_query: str = None,
+        write_query_builder: Callable = None,
+        exposed_type=_EXPOSED_TYPE_PANDAS,
+        scope: Scope = _DEFAULT_SCOPE,
+        **properties,
+    ):
+        """Configure a new SQL data node configuration.
```
toan-quach: same as above

toan-quach (changes requested):

src/taipy/core/config/data_node_config.py
```diff
@@ -449,39 +461,97 @@ def _configure_pickle(id: str, default_data: Optional[Any] = None, scope: Scope
...
         db_driver: str = "ODBC Driver 17 for SQL Server",
         db_extra_args: Dict[str, Any] = None,
+        read_query: str = None,
+        write_query_builder: Callable = None,
         exposed_type=_EXPOSED_TYPE_PANDAS,
         scope: Scope = _DEFAULT_SCOPE,
         **properties,
     ):
-        """Configure a new SQL data node configuration.
+        """Configure a new SQL database data node configuration.
```
khaitruong922:
```suggestion
        """Configure a new SQL data node configuration.
```

src/taipy/core/config/data_node_config.py
```diff
@@ -449,39 +461,97 @@ def _configure_pickle(id: str, default_data: Optional[Any] = None, scope: Scope
...
-            read_query (str): The SQL query string used to read the data from the database.
-            write_table (str): The name of the table in the database to write the data to.
             db_port (int): The database port. The default value is 1433.
             db_host (str): The database host. The default value is _"localhost"_.
             db_driver (str): The database driver. The default value is
                 _"ODBC Driver 17 for SQL Server"_.
             db_extra_args (Dict[str, Any]): A dictionary of additional arguments to be passed into database
                 connection string.
+            read_query (str): The SQL query string used to read the data from the database.
+            write_query_builder (Callable): Function called to build the SQL query from data.
```
khaitruong922:
```suggestion
            write_query_builder (Callable): A callback function which takes in the data and returns a list of queries.
```

src/taipy/core/config/data_node_config.py
khaitruong922:
```suggestion
        """Configure a new SQL data node configuration.
```

src/taipy/core/config/data_node_config.py
```diff
@@ -474,14 +544,14 @@ def _configure_sql(
...
-            read_query (str): The SQL query string used to read the data from the database.
-            write_table (str): The name of the table in the database to write the data to.
             db_port (int): The database port. The default value is 1433.
             db_host (str): The database host. The default value is _"localhost"_.
             db_driver (str): The database driver. The default value is
                 _"ODBC Driver 17 for SQL Server"_.
             db_extra_args (Dict[str, Any]): A dictionary of additional arguments to be passed into database
                 connection string.
+            read_query (str): The SQL query string used to read the data from the database.
+            write_query_builder (Callable): A callback function which takes in the data and returns a list of queries.
```
```suggestion
            write_query_builder (Callable): A callback function that takes the data as an input parameter and returns a list of SQL queries.
```

jrobinAV (dismissed):

jrobinAV (approved):

toan-quach (approved):

Issue #312, opened 2022-09-12
Design and implement a MongoDBDataNode


jrobinAV:
- [ ] Design the Config and the behavior of a DataNode corresponding to a MongoDB Collection.
The read method should be able to get all the documents of a MongoDB Collection provided by the config.
The write method should be able to replace all the documents of the same MongoDB Collection.
The exposed type should be a python iterable of Custom objects provided in the config.
- [ ] Implement a first version
- [ ] implement a milestone demo
- [ ] update release notes
- [ ] update user man and ref man

PR #316, opened 2022-09-15
BLOCKED until 2.0 is released - Feature/#312 implement mongodb datanode

trgiangdo:
Implement MongoCollectionDataNode https://github.com/Avaiga/taipy-core/issues/312


src/taipy/core/data/mongo.py
```diff
@@ -0,0 +1,225 @@
...
+            return pd.DataFrame(list(cursor))[columns]
+        return pd.DataFrame(list(cursor))
+
+    def _read_by_query(self):
+        """Query from MongoDB, exclude the _id field"""
+
+        read_collection = self.mongo_db[self.properties.get("read_collection")]
+        query_result = list(read_collection.find(self.read_query, {"_id": 0}))
+
+        encoded_json = json.dumps(query_result)
```
toan-quach: I'm not sure why we need to convert the return from MongoDB to json? Is there a reason for it?

src/taipy/core/data/mongo.py
```diff
@@ -0,0 +1,225 @@
...
+        self._insert_dicts(self._encode_json(data))
+
+    def _insert_dicts(self, data: List[Dict]) -> None:
+        """
+        :param data: a list of dictionaries
+
+        This method will overwrite the data contained in a list of dictionaries into a collection.
+        """
+        write_collection = self.mongo_db[self.properties.get("write_collection")]
+        write_collection.delete_many({})
```
toan-quach: Why do we have to delete_many before we insert new data?

src/taipy/core/data/mongo.py
```diff
@@ -0,0 +1,225 @@
...
+        if isinstance(data, pd.DataFrame):
+            self._insert_dataframe(data)
+            return
+
+        if not isinstance(data, list):
+            data = [data]
+
+        if len(data) == 0:
+            read_collection = self.mongo_db[self.properties.get("read_collection")]
+            read_collection.delete_many({})
```
toan-quach: not sure why we're deleting from the read_collection here?

toan-quach: and why are we deleting the data when len(data) == 0?

src/taipy/core/data/mongo.py
```diff
@@ -0,0 +1,225 @@
...
+            return
+
+        self._insert_dicts(self._encode_json(data))
```
toan-quach: why do we need to convert to json before writing data? :D

tests/core/data/test_mongo_data_node.py
```diff
@@ -0,0 +1,344 @@
...
+        super().__init__(object_hook=self.object_hook, *args, **kwargs)
+
+    def object_hook(self, o):
+        if o.get("__type__") == "AnotherCustomObject":
+            return AnotherCustomObject(o["id"], o["integer"], o["text"])
+        else:
+            return o
+
+
+class TestMongoDataNode:
```
toan-quach: Can you add tests that will test on actual MongoDB database? We can do a condition for such tests like saying if we detect a MongoDB database running, we will run those tests (tests on actual mongodb database), if not then we will skip them. You can refer to file `tests/core/data/test_sql_table_data_node.py` right after the import for the condition 😃

Pipfile
```diff
@@ -10,6 +10,7 @@ pandas = "==1.3.4"
 sqlalchemy = "==1.4"
 toml = "==0.10"
 taipy-config = {git = "https://git@github.com/avaiga/taipy-config.git@develop"}
+pymongo = "*"
```
jrobinAV: We should define the exact version to depend on.

src/taipy/core/data/mongo.py
trgiangdo: This is to create JSON string for the JSONDecoder in `json.load` (line 175) to decode custom Object, or datetime, or Enum from the return of MongoDB.

I'm not sure if there is a better method to decode custom data tyle directly.

src/taipy/core/data/mongo.py
trgiangdo: Currently, when writing, the content of data node will be overwritten. So writing nothing to the data node means delete everything.

src/taipy/core/data/mongo.py
trgiangdo: The `_encode_json` function will encode custom data type into a dictionary that MongoDB can understand.
The function name seems to be misleading though. I'll need to reconsider this.

src/taipy/core/data/mongo.py
trgiangdo: This is to overwrite old data.

tests/core/data/test_mongo_data_node.py
trgiangdo: As far as I can understand, the test cases for SQL only check for `pyodbc` package. The test are using mocking to test the database, not on actual SQL database.

Do you mean we'll skip the test by checking for actual MongoDB like this? Or removing mongomock?
```suggestion
try:
    pymongo.MongoClient(
        host=db_host,
        port=db_port,
        username=db_username,
        password=db_password,
        serverSelectionTimeoutMS = 2000
    )
    client.server_info()
except:
    pytest.skip("skipping tests because there is no MongoDB database.", allow_module_level=True)
    
class TestMongoDataNode:
```

jrobinAV (commented):
Your implementation is very similar to the SQL Datanode, but it does not really match the requirements. 

We will probably keep many things, but we also need to change a few things.

Let me better explain what I have in mind. 

**What do we want?** 
The MongoDBDataNode represents a data set stored in a single Mongo Collection.
We want the read method to get all the documents from the Collection.
And we want the write method to replace all the documents from the Collection.
We want the `Config.configure_mongo_data_node` method to provide the necessary  parameters to the MongoDataNode so we are able to read and write it.

**So what do we need for that ?**
- the database information (db_name, db_credentials, ...) so we can connect to the DB. (I believe your implementation is good on this part)
- the collection name so we can build the reading and/or the writing query.
- A Python class, representing a Mongo document in Python. Let's call it CustomDocumentClass. (In your implementation this is called exposed_type) 
This CustomDocumentClass must be able to encode the python document into the mongo format and decode the mongo document into Python format. For example, we could require the CustomDocumentClass to implement two methods decode and encode.

So we could expose the following method : 
`Config.configure_mongo_data_node(db_name, user_name, password, collection_name, CustomClass)`

Since we want to read all the collection documents, the data returned by the read method should be an Iterable of CustomDocumentClass objects. And since we want to write the collection documents, the data passed as a parameter of the write method should be an Iterable of CustomDocumentClass objects.

What do you think ?

Let s discuss this on Monday.

src/taipy/core/data/mongo.py
jrobinAV: I believe the user should provide the CustomClass representing the document, and the methods to both convert (encode) from the CustomClass to the MongoDB format, and convert (decode) from the MongoDB format to the CustomClass.

src/taipy/core/data/mongo.py
```diff
@@ -110,7 +109,7 @@ def __init__(
             db_username=properties.get("db_username"),
             db_password=properties.get("db_password"),
         )
-        self.mongo_db = mongo_client[properties.get("db_name")]
+        self.collection = mongo_client[properties.get("db_name")][properties.get("collection_name")]
```
toan-quach, line 102: Hmm then in this case, it should be a `MongoCollectionDataNode` if we want to be precise. What do you think @jrobinAV?

src/taipy/core/data/mongo.py
toan-quach: Why don't we ask the user to provide the dictionary directly to us so we don't need to create a JSON object as the middle-man?

src/taipy/core/data/mongo.py
```diff
@@ -197,9 +194,8 @@ def _insert_dicts(self, data: List[Dict]) -> None:
 
         This method will overwrite the data contained in a list of dictionaries into a collection.
         """
-        write_collection = self.mongo_db[self.properties.get("write_collection")]
-        write_collection.delete_many({})
-        write_collection.insert_many(data)
+        self.collection.delete_many({})
```
toan-quach: @jrobinAV I'm having mixed feelings as to why we would want to delete everything before we want to write new data to the collection. My point of view as a user would be I want to record my historical results, deleting everything would prevent me from doing any comparison with historical data. And a collection that keeps deleting the document sounds weird. My proposal would be to not delete anything when inserting new things, or add another field to the config for user to specify if they want to overwrite everything when writing or not (by default I strongly think it should be no)

src/taipy/core/data/mongo.py
toan-quach: Ah I understand what you were trying to do now, the thing that I'm a bit confused is that, if we provide either a pymongo type output, or a json type output, the user will still need to provide an encoding and decoding method right? If yes, then do we need json as the middle-man when we can provide the pymongo output directly? Either way, whether it's JSON or pymongo output, I know for certain we have to document this carefully so the user knows what to expect and provide the CustomClass and its encoder/decoder methods (if any) better :D

src/taipy/core/data/mongo.py
trgiangdo: The JSON encoder is to ensure we insert the right format to MongoDB, and for now, the method is to use JSON string as the middle-man. I'll try a custom object and we can discuss some use cases.

tests/core/data/test_mongo_data_node.py
```diff
@@ -9,18 +9,24 @@
...
 import json
 from dataclasses import dataclass
 from enum import Enum
+from importlib import util
 from unittest import mock
 
+import pytest
+
+if not util.find_spec("pymongo"):
+    pytest.skip("skipping tests because PyODBC is not installed", allow_module_level=True)
```
toan-quach:
```suggestion
    pytest.skip("skipping tests because PyMongo is not installed", allow_module_level=True)
```

tox.ini
```diff
@@ -1,7 +1,7 @@
 [tox]
 skipsdist = true
 isolated_build = true
-envlist = clean, lint, without-pyodbc
+envlist = clean, lint, without-extras
```
toan-quach: Minor remark only, I would say without-pyodbc-and-pymongo is better, since we aren't sure what extra packages we might use in the future, I'll leave the judgement call to you

trgiangdo:
```suggestion
envlist = clean, lint, without-pyodbc-and-pymongo
```

tox.ini
```diff
@@ -26,15 +26,28 @@ commands =
     black taipy tests
     flake8 taipy tests
 
-[testenv:without-pyodbc]
+[testenv:without-extras]
```
trgiangdo:
```suggestion
[testenv:without-pyodbc-and-pymongo]
```

src/taipy/core/config/data_node_config.py
```diff
@@ -96,7 +95,13 @@ class DataNodeConfig(Section):
...
     _REQUIRED_WRITE_TABLE_SQL_PROPERTY = "write_table"
     _OPTIONAL_DB_EXTRA_ARGS_SQL_PROPERTY = "db_extra_args"
-
+    # MONGO
+    _OPTIONAL_EXPOSED_TYPE_MONGO_PROPERTY = "exposed_type"
+    _REQUIRED_DB_USERNAME_MONGO_PROPERTY = "db_username"
+    _REQUIRED_DB_PASSWORD_MONGO_PROPERTY = "db_password"
+    _REQUIRED_DB_NAME_MONGO_PROPERTY = "db_name"
+    _REQUIRED_COLLECTION_NAME_MONGO_PROPERTY = "collection_name"
+    _REQUIRED_READ_QUERY_MONGO_PROPERTY = "read_query"
```
```suggestion
```

src/taipy/core/config/data_node_config.py
```diff
@@ -115,6 +120,13 @@ class DataNodeConfig(Section):
             _REQUIRED_READ_QUERY_SQL_PROPERTY,
             _REQUIRED_WRITE_TABLE_SQL_PROPERTY,
         ],
+        _STORAGE_TYPE_VALUE_MONGO_COLLECTION: [
+            _REQUIRED_DB_USERNAME_MONGO_PROPERTY,
+            _REQUIRED_DB_PASSWORD_MONGO_PROPERTY,
+            _REQUIRED_DB_NAME_MONGO_PROPERTY,
+            _REQUIRED_COLLECTION_NAME_MONGO_PROPERTY,
+            _REQUIRED_READ_QUERY_MONGO_PROPERTY,
```
```suggestion
```

src/taipy/core/config/data_node_config.py
```diff
@@ -509,3 +522,53 @@ def _configure_sql(
...
+    def _configure_mongo_collection(
+        id: str,
+        db_username: str,
+        db_password: str,
+        db_name: str,
+        read_query: str,
+        collection_name: str = None,
+        db_port: int = 27017,
+        db_host: str = "localhost",
+        exposed_type=_EXPOSED_TYPE_PANDAS,
```
jrobinAV: We don't want to have pandas as the default exposed type. I believe the exposed_type should be required, as a DataClass provided by the user.

```suggestion
        exposed_type: Any,
```

src/taipy/core/config/data_node_config.py
```diff
@@ -509,3 +522,53 @@ def _configure_sql(
...
+        Parameters:
+            id (str): The unique identifier of the new Mongo collection data node configuration.
+            db_username (str): The database username.
+            db_password (str): The database password.
+            db_name (str): The database name.
+            collection_name (str): The collection in the database to read from and to write the data to.
+            read_query (str): The Mongo query string used to read the data from the database.
+            db_port (int): The database port. The default value is 27017.
+            db_host (str): The database host. The default value is _"localhost"_.
+            exposed_type: The exposed type of the data read from Mongo query. The default value is `pandas`.
```
```suggestion
            exposed_type: The custom class representing the exposed type of the data read from the Mongo collection.
```

src/taipy/core/data/mongo.py
```diff
@@ -0,0 +1,205 @@
...
+        last_edit_date (datetime): The date and time of the last modification.
+        job_ids (List[str]): The ordered list of jobs that have written this data node.
+        validity_period (Optional[timedelta]): The validity period of a cacheable data node.
+            Implemented as a timedelta. If _validity_period_ is set to None, the data_node is
+            always up-to-date.
+        edit_in_progress (bool): True if a task computing the data node has been submitted
+            and not completed yet. False otherwise.
+        properties (dict[str, Any]): A dictionary of additional properties. Note that the
+            _properties_ parameter must at least contain an entry for _"db_username"_,
+            _"db_password"_, _"db_name"_, _"collection_name"_, and _"read_query"_.
```
jrobinAV:
```suggestion
            _"db_password"_, _"db_name"_, _"collection_name"_, and _"exposed_type"_.
```

src/taipy/core/data/mongo.py
```diff
@@ -0,0 +1,205 @@
...
+    """
+
+    __STORAGE_TYPE = "mongo_collection"
+    __EXPOSED_TYPE_PROPERTY = "exposed_type"
+    _REQUIRED_PROPERTIES: List[str] = [
+        "db_username",
+        "db_password",
+        "db_name",
+        "collection_name",
+        "read_query",
```
jrobinAV:
```suggestion
```

src/taipy/core/data/mongo.py
```diff
@@ -0,0 +1,205 @@
...
+    def storage_type(cls) -> str:
+        return cls.__STORAGE_TYPE
+
+    def _read(self):
+        cursor = self._read_by_query()
+
+        return list(map(lambda row: self._decoder(row), cursor))
+
+    def _read_by_query(self):
+        """Query from a Mongo collection, exclude the _id field"""
```
jrobinAV, line 135: I don't think we should exclude the id. We should let the responsibility to the decoder for excluding it or not.

jrobinAV (changes requested):

src/taipy/core/config/data_node_config.py
```diff
@@ -642,3 +656,55 @@ def _configure_sql(
...
+        id: str,
+        db_username: str,
+        db_password: str,
+        db_name: str,
+        custom_document: Any,
+        read_query: dict = {},
+        collection_name: str = None,
+        db_port: int = 27017,
+        db_host: str = "localhost",
+        scope: Scope = _DEFAULT_SCOPE,
```
toan-quach, line 673: please help me add cacheable param here as well. Thanks!

src/taipy/core/config/data_node_config.py
```diff
@@ -102,6 +101,13 @@ class DataNodeConfig(Section):
     # SQL
     _REQUIRED_READ_QUERY_SQL_PROPERTY = "read_query"
     _REQUIRED_WRITE_QUERY_BUILDER_SQL_PROPERTY = "write_query_builder"
+    # MONGO
+    _OPTIONAL_READ_QUERY_MONGO_PROPERTY = "read_query"
```
```suggestion
```

src/taipy/core/config/data_node_config.py
```diff
@@ -156,6 +169,7 @@ class DataNodeConfig(Section):
         _STORAGE_TYPE_VALUE_IN_MEMORY: [_OPTIONAL_DEFAULT_DATA_IN_MEMORY_PROPERTY],
         _STORAGE_TYPE_VALUE_SQL_TABLE: [_OPTIONAL_EXPOSED_TYPE_SQL_PROPERTY, _OPTIONAL_DB_EXTRA_ARGS_SQL_PROPERTY],
         _STORAGE_TYPE_VALUE_SQL: [_OPTIONAL_EXPOSED_TYPE_SQL_PROPERTY, _OPTIONAL_DB_EXTRA_ARGS_SQL_PROPERTY],
+        _STORAGE_TYPE_VALUE_MONGO_COLLECTION: [_OPTIONAL_READ_QUERY_MONGO_PROPERTY],
```
```suggestion
```

src/taipy/core/config/data_node_config.py
```diff
@@ -642,3 +656,58 @@ def _configure_sql(
...
         return Config.sections[DataNodeConfig.name][id]
+
+    @staticmethod
+    def _configure_mongo_collection(
+        id: str,
+        db_username: str,
+        db_password: str,
+        db_name: str,
+        custom_document: Any,
+        read_query: dict = {},
```
```suggestion
```

src/taipy/core/config/data_node_config.py
```diff
@@ -642,3 +656,58 @@ def _configure_sql(
...
+        section = DataNodeConfig(
+            id,
+            DataNodeConfig._STORAGE_TYPE_VALUE_MONGO_COLLECTION,
+            scope=scope,
+            cacheable=cacheable,
+            db_username=db_username,
+            db_password=db_password,
+            db_name=db_name,
+            collection_name=collection_name,
+            read_query=read_query,
```
```suggestion
```

src/taipy/core/config/data_node_config.py
```diff
@@ -642,3 +656,58 @@ def _configure_sql(
...
+    ):
+        """Configure a new Mongo collection data node configuration.
+
+        Parameters:
+            id (str): The unique identifier of the new Mongo collection data node configuration.
+            db_username (str): The database username.
+            db_password (str): The database password.
+            db_name (str): The database name.
+            collection_name (str): The collection in the database to read from and to write the data to.
+            read_query (dict): The Mongo query dictionary used to read the data from the database.
```
```suggestion
```

src/taipy/core/config/data_node_config.py
```diff
@@ -642,3 +656,58 @@ def _configure_sql(
...
+            id,
+            DataNodeConfig._STORAGE_TYPE_VALUE_MONGO_COLLECTION,
+            scope=scope,
+            cacheable=cacheable,
+            db_username=db_username,
+            db_password=db_password,
+            db_name=db_name,
+            collection_name=collection_name,
+            read_query=read_query,
+            exposed_type=custom_document,
```
```suggestion
            custom_document=custom_document,
```

src/taipy/core/config/data_node_config.py
```diff
@@ -102,6 +101,13 @@ class DataNodeConfig(Section):
     # SQL
     _REQUIRED_READ_QUERY_SQL_PROPERTY = "read_query"
     _REQUIRED_WRITE_QUERY_BUILDER_SQL_PROPERTY = "write_query_builder"
+    # MONGO
+    _OPTIONAL_READ_QUERY_MONGO_PROPERTY = "read_query"
+    _REQUIRED_EXPOSED_TYPE_MONGO_PROPERTY = "exposed_type"
```
```suggestion
    _REQUIRED_CUSTOM_DOCUMENT_MONGO_PROPERTY = "custom_document"
```

src/taipy/core/config/data_node_config.py
```diff
@@ -127,6 +133,13 @@ class DataNodeConfig(Section):
             _REQUIRED_READ_QUERY_SQL_PROPERTY,
             _REQUIRED_WRITE_QUERY_BUILDER_SQL_PROPERTY,
         ],
+        _STORAGE_TYPE_VALUE_MONGO_COLLECTION: [
+            _REQUIRED_DB_USERNAME_MONGO_PROPERTY,
+            _REQUIRED_DB_PASSWORD_MONGO_PROPERTY,
+            _REQUIRED_DB_NAME_MONGO_PROPERTY,
+            _REQUIRED_COLLECTION_NAME_MONGO_PROPERTY,
+            _REQUIRED_EXPOSED_TYPE_MONGO_PROPERTY,
```
```suggestion
            _REQUIRED_CUSTOM_DOCUMENT_MONGO_PROPERTY,
```

src/taipy/core/data/mongo.py
```diff
@@ -0,0 +1,207 @@
...
+            always up-to-date.
+        edit_in_progress (bool): True if a task computing the data node has been submitted
+            and not completed yet. False otherwise.
+        properties (dict[str, Any]): A dictionary of additional properties. Note that the
+            _properties_ parameter must at least contain an entry for _"db_username"_,
+            _"db_password"_, _"db_name"_, _"collection_name"_, and _"custom_document"_.
+    """
+
+    __STORAGE_TYPE = "mongo_collection"
+    __CUSTOM_DOCUMENT_PROPERTY = "exposed_type"
```
jrobinAV:
```suggestion
    __CUSTOM_DOCUMENT_PROPERTY = "custom_document"
```

src/taipy/core/data/mongo.py
```diff
@@ -0,0 +1,207 @@
...
+    ):
+        if properties is None:
+            properties = {}
+        required = self._REQUIRED_PROPERTIES
+        if missing := set(required) - set(properties.keys()):
+            raise MissingRequiredProperty(
+                f"The following properties " f"{', '.join(x for x in missing)} were not informed and are required"
+            )
+
+        self._check_exposed_type(properties[self.__CUSTOM_DOCUMENT_PROPERTY])
```
jrobinAV: can be renamed to `_check_custom_document()`

src/taipy/core/data/mongo.py
```diff
@@ -0,0 +1,207 @@
...
+            self._encoder = custom_encoder
+        else:
+            self._encoder = self._default_encoder
+
+        if not self._last_edit_date:
+            self.unlock_edit()
+
+    def _check_exposed_type(self, exposed_type):
+        if not isclass(exposed_type):
+            raise InvalidExposedType(f"Invalid exposed type of {exposed_type}. Only custom object class are supported.")
```
jrobinAV: We should create a dedicated exception. Something like `InvalidCustomDocument`

src/taipy/core/data/mongo.py
```diff
@@ -0,0 +1,207 @@
...
+
+    def _read(self):
+        cursor = self._read_by_query()
+
+        return list(map(lambda row: self._decoder(row), cursor))
+
+    def _read_by_query(self):
+        """Query from a Mongo collection, exclude the _id field"""
+
+        return self.collection.find(self.read_query)
```
jrobinAV:
```suggestion
        return self.collection.find()
```

src/taipy/core/data/mongo.py
```diff
@@ -0,0 +1,207 @@
...
+
+    def _write(self, data) -> None:
+        """Check data against a collection of types to handle insertion on the database."""
+
+        if not isinstance(data, list):
+            data = [data]
+
+        if len(data) == 0:
+            self.collection.delete_many({})
```
jrobinAV: I guess using the drop command is better in terms of performance. 
```suggestion
            self.collection.drop()
```
However, I wonder if we need to recreate the collection after dropping it.
`self.collection = mongo_client[properties.get("db_name")][properties.get("collection_name")]`

src/taipy/core/data/mongo.py
jrobinAV: For consistency reasons with other data nodes, we should keep the deletion. It is also a way to ensure the consistency of the collection itself. For example, let s imagine we write twice the data with 2 different datasets that have some overlapping. What would be the result?

However, the use case you mentioned is completely valid. Sometimes we want to keep history. But I believe this use case is more generic, and not specific to MongoDB. We should handle this on a different ticket.
For SQL data node we can handle the use case by using SQLDataNode with custom queries and not SQLTableDataNode. We could imagine something similar for Mongo, but I am not sure that is the right way. 
A way to do it for all kinds of data nodes would be to have an overwrite attribute (as you proposed). If it's False, we systematically add a `writer_id` field in the data (mongo document, SQL schema, CSV column, Excel column, ...) that would correspond to the identifier of the writer (basically the JobId). FYI, the JobId is already stored in the data node when we write it.

src/taipy/core/data/mongo.py
trgiangdo: You are right on `.drop()` method.

The recreation of a collection is actually not necessary, since if a collection is not exist, the `find()` method will return a empty cursor, which can be converted to an empty list.

src/taipy/core/data/_data_repository.py
```diff
@@ -96,6 +97,11 @@ def _to_model(self, data_node: DataNode):
                         self._EXPOSED_TYPE_KEY
                     ] = f"{properties[self._EXPOSED_TYPE_KEY].__module__}.{properties[self._EXPOSED_TYPE_KEY].__qualname__}"
 
+        if self._CUSTOM_DOCUMENT_KEY in properties.keys():
+            properties[
+                self._CUSTOM_DOCUMENT_KEY
+            ] = f"{properties[self._CUSTOM_DOCUMENT_KEY].__module__}.{properties[self._CUSTOM_DOCUMENT_KEY].__qualname__}"
+
```
jrobinAV, line 104: Great!

src/taipy/core/exceptions/exceptions.py
```diff
@@ -75,6 +75,10 @@ class InvalidExposedType(Exception):
     """Raised if an invalid exposed type is provided."""
 
 
+class InvalidCustomDocument(Exception):
+    """Raised if an invalid custom document class is provided."""
```
jrobinAV:
```suggestion
    """Raised if an invalid custom document class is provided to a `MongoCollectionDataNode`."""
```

jrobinAV (dismissed):
Looks very good to me.

jrobinAV (approved):
Great !

Issue #307, opened 2022-09-02
Ability to export the local files of a single scenario


FlorianJacta:
**What would that feature address**
This feature would enable the user to export all the files needed for a scenario without exporting all the .data. Sometimes, just one or a small group of scenarios need to be exported.

***Description of the ideal solution***
A Python line of code and/or an API Rest command line should be able to give all the files of the local system concerning a scenario

***Proposal***
Having a Python command to export a scenario : `taipy.export("folder_name", "scenario_id")`

The result would be the following folder hierarchy:
```
folder_name
   - cycles
        - cycle_id.json
   - datanodes
        - datanode_id1.json
        - datanode_id2.json
        ...
   - jobs
        - job_id1.json
        - job_id2.json
        - job_id3.json
        ...
   - pipelines
        - pipeline_id1.json
        - pipeline_id2.json
   - scenarios
        scenario_id.json
   - tasks
        - task_id1.json
        - task_id2.json
```

When the folder_name already exists, the call of `taipy.export()` should delete the folder before recreating it.

jrobinAV:
@FlorianJacta Do we want to export pickle files as well ? I would say no! Do you agree?

Answer is no

jrobinAV:
If a MongoRepository or an SQLRepository is used, we still want to export the scenario as JSON files. This should be part of another ticket.

PR #317, opened 2022-09-19
Feature - Export scenario

khaitruong922:
#307

### DIFF
--- a/src/taipy/core/_manager/_manager.py
+++ b/src/taipy/core/_manager/_manager.py
@@ -99,8 +99,8 @@ class _Manager(Generic[EntityType]):
         _DataManagerFactory._build_manager()._delete_many(_entity_ids.data_node_ids)
 
     @classmethod
-    def _export(cls, id, **kwargs):
+    def _export(cls, id: str, folder_path: str):
         """
         Export an entity.
         """
-        return cls._repository._export(id, **kwargs)
+        return cls._repository._export(id, folder_path)
--- a/src/taipy/core/_manager/_manager.py
+++ b/src/taipy/core/_manager/_manager.py
@@ -97,3 +97,10 @@ class _Manager(Generic[EntityType]):
         _TaskManagerFactory._build_manager()._delete_many(_entity_ids.task_ids)
         _JobManagerFactory._build_manager()._delete_many(_entity_ids.job_ids)
         _DataManagerFactory._build_manager()._delete_many(_entity_ids.data_node_ids)
+
+    @classmethod
+    def _export(cls, id, **kwargs):
+        """
+        Export an entity.
+        """
+        return cls._repository._export(id, **kwargs)
--- a/src/taipy/core/_repository/_repository.py
+++ b/src/taipy/core/_repository/_repository.py
@@ -116,13 +116,13 @@ class _AbstractRepository(Generic[ModelType, Entity]):
         raise NotImplementedError
 
     @abstractmethod
-    def _export(self, entity_id: str, **kwargs):
+    def _export(self, entity_id: str, folder_path: str):
         """
         Export an entity from the repository.
 
         Args:
             entity_id: The id of the entity to be exported.
-            **kwargs: Additional arguments to be passed to the export function.
+            folder_path (str): The folder path to export the entity to.
 
         """
         raise NotImplementedError
--- a/src/taipy/core/_repository/_repository.py
+++ b/src/taipy/core/_repository/_repository.py
@@ -115,6 +115,18 @@ class _AbstractRepository(Generic[ModelType, Entity]):
         """
         raise NotImplementedError
 
+    @abstractmethod
+    def _export(self, entity_id: str, **kwargs):
+        """
+        Export an entity from the repository.
+
+        Args:
+            entity_id: The id of the entity to be exported.
+            **kwargs: Additional arguments to be passed to the export function.
+
+        """
+        raise NotImplementedError
+
 
 class _CustomEncoder(json.JSONEncoder):
     def default(self, o: Any) -> Json:
--- a/src/taipy/core/common/_entity_ids.py
+++ b/src/taipy/core/common/_entity_ids.py
@@ -10,6 +10,7 @@
 # specific language governing permissions and limitations under the License.
 
 from __future__ import annotations
+from typing import Set
 
 
 class _EntityIds:
@@ -35,7 +36,7 @@ class _EntityIds:
         return self
 
     @property
-    def _union_all(self) -> set:
+    def _union_all(self) -> Set:
         return set().union(
             self.data_node_ids, self.task_ids, self.scenario_ids, self.pipeline_ids, self.job_ids, self.cycle_ids
         )
--- a/src/taipy/core/common/_entity_ids.py
+++ b/src/taipy/core/common/_entity_ids.py
@@ -10,6 +10,7 @@
 # specific language governing permissions and limitations under the License.
 
 from __future__ import annotations
+
 from typing import Set
 
 
@@ -34,9 +35,3 @@ class _EntityIds:
     def __iadd__(self, other: _EntityIds):
         self.__add__(other)
         return self
-
-    @property
-    def _union_all(self) -> Set:
-        return set().union(
-            self.data_node_ids, self.task_ids, self.scenario_ids, self.pipeline_ids, self.job_ids, self.cycle_ids
-        )
--- a/src/taipy/core/common/_entity_ids.py
+++ b/src/taipy/core/common/_entity_ids.py
@@ -33,3 +33,9 @@ class _EntityIds:
     def __iadd__(self, other: _EntityIds):
         self.__add__(other)
         return self
+
+    @property
+    def _union_all(self) -> set:
+        return set().union(
+            self.data_node_ids, self.task_ids, self.scenario_ids, self.pipeline_ids, self.job_ids, self.cycle_ids
+        )
--- a/src/taipy/core/cycle/_cycle_manager.py
+++ b/src/taipy/core/cycle/_cycle_manager.py
@@ -86,12 +86,12 @@ class _CycleManager(_Manager[Cycle]):
     @classmethod
     def _hard_delete(cls, cycle_id: CycleId):
         cycle = cls._get(cycle_id)
-        entity_ids_to_delete = cls._get_owned_entity_ids(cycle)
+        entity_ids_to_delete = cls._get_children_entity_ids(cycle)
         entity_ids_to_delete.cycle_ids.add(cycle.id)
         cls._delete_entities_of_multiple_types(entity_ids_to_delete)
 
     @classmethod
-    def _get_owned_entity_ids(cls, cycle: Cycle) -> _EntityIds:
+    def _get_children_entity_ids(cls, cycle: Cycle) -> _EntityIds:
         from ..scenario._scenario_manager_factory import _ScenarioManagerFactory
 
         entity_ids = _EntityIds()
--- a/src/taipy/core/cycle/_cycle_manager.py
+++ b/src/taipy/core/cycle/_cycle_manager.py
@@ -47,7 +47,9 @@ class _CycleManager(_Manager[Cycle]):
     ) -> Cycle:
         creation_date = creation_date if creation_date else datetime.now()
         start_date = _CycleManager._get_start_date_of_cycle(frequency, creation_date)
-        cycles = cls._repository.get_cycles_by_frequency_and_start_date(frequency=frequency, start_date=start_date)  # type: ignore
+        cycles = cls._repository.get_cycles_by_frequency_and_start_date(
+            frequency=frequency, start_date=start_date
+        )  # type: ignore
         if len(cycles) > 0:
             return cycles[0]
         else:
@@ -84,12 +86,12 @@ class _CycleManager(_Manager[Cycle]):
     @classmethod
     def _hard_delete(cls, cycle_id: CycleId):
         cycle = cls._get(cycle_id)
-        entity_ids_to_delete = cls._get_owned_entities(cycle)
+        entity_ids_to_delete = cls._get_owned_entity_ids(cycle)
         entity_ids_to_delete.cycle_ids.add(cycle.id)
         cls._delete_entities_of_multiple_types(entity_ids_to_delete)
 
     @classmethod
-    def _get_owned_entities(cls, cycle: Cycle) -> _EntityIds:
+    def _get_owned_entity_ids(cls, cycle: Cycle) -> _EntityIds:
         from ..scenario._scenario_manager_factory import _ScenarioManagerFactory
 
         entity_ids = _EntityIds()
--- a/src/taipy/core/pipeline/_pipeline_manager.py
+++ b/src/taipy/core/pipeline/_pipeline_manager.py
@@ -118,12 +118,12 @@ class _PipelineManager(_Manager[Pipeline]):
     @classmethod
     def _hard_delete(cls, pipeline_id: PipelineId):
         pipeline = cls._get(pipeline_id)
-        entity_ids_to_delete = cls._get_owned_entity_ids(pipeline)
+        entity_ids_to_delete = cls._get_children_entity_ids(pipeline)
         entity_ids_to_delete.pipeline_ids.add(pipeline.id)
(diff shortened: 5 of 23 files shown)


src/taipy/core/taipy.py
```diff
@@ -463,3 +466,34 @@ def clean_all_entities() -> bool:
...
+    scenario_id: ScenarioId,
+    folder: str,
+):
+    """Export all related entities of a scenario to a folder.
+
+    Parameters:
+        scenario_id (ScenarioId): The id of the scenario to export.
+        folder (str): The folder to export. This folder must not be the storage folder.
+
+    """
```
```suggestion
    """Export all related entities of a scenario to a folder.

    Parameters:
        scenario_id (ScenarioId): The id of the scenario to export.
        folder (str): The folder to export. This folder must not be the storage folder.
    
    Raises:
        InvalidExportPath^: If the folder path is invalid or reserved by Taipy.
    """
```

src/taipy/core/pipeline/_pipeline_manager.py
```diff
@@ -124,6 +126,8 @@ def _hard_delete(cls, pipeline_id: PipelineId):
     def _get_owned_entity_ids(cls, pipeline: Pipeline) -> _EntityIds:
         entity_ids = _EntityIds()
         for task in pipeline.tasks.values():
+            if not isinstance(task, Task):
+                task = _TaskManagerFactory._build_manager()._get(task)
```
jrobinAV, line 130: How can task variable can be a task ID here?

src/taipy/core/scenario/_scenario_manager.py
```diff
@@ -275,6 +277,8 @@ def _get_owned_entity_ids(cls, scenario: Scenario) -> _EntityIds:
             if pipeline.parent_id in (pipeline.id, scenario.id):
                 entity_ids.pipeline_ids.add(pipeline.id)
             for task in pipeline.tasks.values():
+                if not isinstance(task, Task):
+                    task = _TaskManagerFactory._build_manager()._get(task)
```
jrobinAV, line 281: Same question

src/taipy/core/taipy.py
```diff
@@ -463,3 +466,34 @@ def clean_all_entities() -> bool:
...
+
+    # Copy storage folder to the export folder
+    if Path(folder).exists():
+        shutil.rmtree(folder)
+    shutil.copytree(Config.global_config.storage_folder, folder, dirs_exist_ok=True)
+
+    # Remove all entities that are not related to the scenario
+    for f in Path(folder).rglob("*"):
+        if f.is_file() and f.stem not in entity_ids:
+            f.unlink()
```
jrobinAV: 1/ This only works with Filesystem storage. The logic of getting all the entity ids is good. But we should move the logic of extracting the entities (from the ids) to the repositories. The purpose is to implement the same for other storage systems (mongo, SQL). 

2/ I believe this implementation will be problematic when the user has many scenarios and entities. It will copy everything which can take a while and remove the major part of the files. 
Since you already have the list of entities, it should be straightforward to directly copy the files of the entities related to the scenario.

jrobinAV (changes requested):

src/taipy/core/pipeline/_pipeline_manager.py
khaitruong922: Not sure but I check the type of `pipeline.tasks` and it returns a list of task ids or tasks.

```python
    @tasks.setter  # type: ignore
    @_self_setter(_MANAGER_NAME)
    def tasks(self, tasks: Union[List[TaskId], List[Task]]):
        self._tasks = tasks
```

src/taipy/core/taipy.py
khaitruong922: Yes, so I am thinking of adding a method in manager class to retrieve the path of an entity by id. About the Mongo and SQL repo, we should export them into another database, am I correct?

src/taipy/core/pipeline/_pipeline_manager.py
jrobinAV: Ah yes. I remember now. We did it to optimize performance. OK thx.

src/taipy/core/taipy.py
jrobinAV: I would put this method in the FileSystemRepository. not in the manager.
The mongo and SQL repo should also export the data into the same folder.

src/taipy/core/taipy.py
```diff
@@ -463,3 +466,34 @@ def clean_all_entities() -> bool:
...
+        scenario_id (ScenarioId): The id of the scenario to export.
+        **kwargs: Additional keyword arguments.
+    """
+
+    manager = _ScenarioManagerFactory._build_manager()
+    scenario = manager._get(scenario_id)
+    entity_ids = manager._get_owned_entity_ids(scenario)  # type: ignore
+    entity_ids.scenario_ids = {scenario_id}
+    entity_ids.cycle_ids = {scenario.cycle.id}
+
```
jrobinAV, line 485: Here you can remove the folder (kwargs["folder"]) if it exists already

src/taipy/core/taipy.py
```diff
@@ -463,3 +466,34 @@ def clean_all_entities() -> bool:
...
+    for task_id in entity_ids.task_ids:
+        _TaskManagerFactory._build_manager()._export(task_id, **kwargs)
+    for pipeline_id in entity_ids.pipeline_ids:
+        _PipelineManagerFactory._build_manager()._export(pipeline_id, **kwargs)
+    for cycle_id in entity_ids.cycle_ids:
+        _CycleManagerFactory._build_manager()._export(cycle_id, **kwargs)
+    for scenario_id in entity_ids.scenario_ids:
+        _ScenarioManagerFactory._build_manager()._export(scenario_id, **kwargs)
+    for job_id in entity_ids.job_ids:
+        _JobManagerFactory._build_manager()._export(job_id, **kwargs)
```
```suggestion
    for data_node_id in entity_ids.data_node_ids:
        _DataManagerFactory._build_manager()._export(data_node_id, folder=kwargs.pop("folder")+"/datanodes", **kwargs)
    for task_id in entity_ids.task_ids:
        _TaskManagerFactory._build_manager()._export(task_id, folder=kwargs.pop("folder")+"/tasks", **kwargs)
    for pipeline_id in entity_ids.pipeline_ids:
        _PipelineManagerFactory._build_manager()._export(pipeline_id, folder=kwargs.pop("folder")+"/pipelines", **kwargs)
    for cycle_id in entity_ids.cycle_ids:
        _CycleManagerFactory._build_manager()._export(cycle_id, folder=kwargs.pop("folder")+"/cycles", **kwargs)
    for scenario_id in entity_ids.scenario_ids:
        _ScenarioManagerFactory._build_manager()._export(scenario_id, folder=kwargs.pop("folder")+"/scenarios", **kwargs)
    for job_id in entity_ids.job_ids:
        _JobManagerFactory._build_manager()._export(job_id, folder=kwargs.pop("folder")+"/jobs", **kwargs)
```

src/taipy/core/pipeline/_pipeline_manager.py
```diff
@@ -124,6 +126,8 @@ def _hard_delete(cls, pipeline_id: PipelineId):
     def _get_owned_entity_ids(cls, pipeline: Pipeline) -> _EntityIds:
         entity_ids = _EntityIds()
         for task in pipeline.tasks.values():
+            if not isinstance(task, Task):
```
toan-quach, line 129: I remember pipeline.tasks being task ids instead of task entity, if so can we remove the if condition here and go straight to _get?

src/taipy/core/scenario/_scenario_manager.py
```diff
@@ -275,6 +277,8 @@ def _get_owned_entity_ids(cls, scenario: Scenario) -> _EntityIds:
             if pipeline.parent_id in (pipeline.id, scenario.id):
                 entity_ids.pipeline_ids.add(pipeline.id)
             for task in pipeline.tasks.values():
+                if not isinstance(task, Task):
```
toan-quach, line 280: same point as above

toan-quach (commented):
I thing it'd be great if we can also have scenario.export(path) also, meaning to call the function directly from the entity

src/taipy/core/pipeline/_pipeline_manager.py
khaitruong922:
```python   
 @tasks.setter  # type: ignore
    @_self_setter(_MANAGER_NAME)
    def tasks(self, tasks: Union[List[TaskId], List[Task]]):
        self._tasks = tasks
```

This is the signature of pipeline.tasks so we should expect an item can be a task id or a task.

khaitruong922:
```python
        tasks: Union[List[TaskId], List[Task], List[Union[TaskId, Task]]],
```

This is the type annotation of pipeline.tasks so we should expect an item to be either a task id or task.

src/taipy/core/_manager/_manager.py
```diff
@@ -97,3 +97,10 @@ def _delete_entities_of_multiple_types(cls, _entity_ids: _EntityIds):
         _TaskManagerFactory._build_manager()._delete_many(_entity_ids.task_ids)
         _JobManagerFactory._build_manager()._delete_many(_entity_ids.job_ids)
         _DataManagerFactory._build_manager()._delete_many(_entity_ids.data_node_ids)
+
+    @classmethod
+    def _export(cls, id, **kwargs):
```
toan-quach, line 102: instead of just the id, maybe we can expand this further and accept the entity object itself too

toan-quach: and why do we use kwargs here instead of something like folder_path?

src/taipy/core/common/_entity_ids.py
```diff
@@ -33,3 +34,9 @@ def __add__(self, other: _EntityIds):
     def __iadd__(self, other: _EntityIds):
         self.__add__(other)
         return self
+
+    @property
+    def _union_all(self) -> Set:
```
toan-quach, line 39: Not sure what this function is for?

src/taipy/core/_manager/_manager.py
khaitruong922: 1. We can always get the id with `entity.id` so what is the use of having the option to pass the entity?
2. kwargs is used to allow the export method to be fully generic at repository level. For example, if the scenario will be exported into a SQLite file if the SQLite repo is used. Similarly, the scenario will be exported into a collection if the MongoDB repo is used. Either way the paramater cannot be `folder_path`, and we may need more than one parameter for other repos.

src/taipy/core/common/_entity_ids.py
khaitruong922: This is to return a set of all the entity ids in the `EntityIds` object itself. It is not used after I have made some changes with the export implementation, but you can keep it if it can be useful in the future.

src/taipy/core/cycle/_cycle_manager.py
```diff
@@ -84,12 +86,12 @@ def _get_end_date_of_cycle(frequency: Frequency, start_date: datetime):
...
     def _hard_delete(cls, cycle_id: CycleId):
         cycle = cls._get(cycle_id)
-        entity_ids_to_delete = cls._get_owned_entities(cycle)
+        entity_ids_to_delete = cls._get_owned_entity_ids(cycle)
         entity_ids_to_delete.cycle_ids.add(cycle.id)
         cls._delete_entities_of_multiple_types(entity_ids_to_delete)
 
     @classmethod
-    def _get_owned_entities(cls, cycle: Cycle) -> _EntityIds:
+    def _get_owned_entity_ids(cls, cycle: Cycle) -> _EntityIds:
```
toan-quach:
```suggestion
    def _get_children_entity_ids(cls, cycle: Cycle) -> _EntityIds:
```

src/taipy/core/common/_entity_ids.py
toan-quach: Not sure what this function is for, find no usage

src/taipy/core/pipeline/_pipeline_manager.py
```diff
@@ -124,6 +126,8 @@ def _hard_delete(cls, pipeline_id: PipelineId):
     def _get_owned_entity_ids(cls, pipeline: Pipeline) -> _EntityIds:
```
toan-quach:
```suggestion
    def _get_owned_children_ids(cls, pipeline: Pipeline) -> _EntityIds:
```

src/taipy/core/scenario/_scenario_manager.py
```diff
@@ -275,6 +277,8 @@ def _get_owned_entity_ids(cls, scenario: Scenario) -> _EntityIds:
             if pipeline.parent_id in (pipeline.id, scenario.id):
```
toan-quach, line 277: this function should be called _get_children_entity_ids

src/taipy/core/_manager/_manager.py
toan-quach:
```suggestion
    def _export(cls, entity: Union[Entity, str], folder_path: str):
```

Is there another format for exporting entities or we will only export them as JSON? If we'll only be exporting them as JSON, I think folder_path is better choice than kwargs

src/taipy/core/taipy.py
```diff
@@ -463,3 +464,40 @@ def clean_all_entities() -> bool:
     _CycleManagerFactory._build_manager()._delete_all()
     _JobManagerFactory._build_manager()._delete_all()
     return True
+
+
+def export_scenario(
```
toan-quach, line 469:
```suggestion
def export_scenario(scenario: Union[Scenario, ScenarioId], **kwargs):
```

Instead of just the id, we can expand this further and accept the entity object itself too. What kind of formats we're expecting to export, if we're only expecting to export them as Json, wouldn't `folder_path` be better than `kwargs`?

Do we also want to expand this feature to other entities like Cycle, Pipeline, Task?

tests/core/common/test_entity_ids.py
```diff
@@ -44,3 +44,27 @@ def test_add_two_entity_ids(self):
         assert entity_ids_1.pipeline_ids == {"pipeline_id_1", "pipeline_id_2", "pipeline_id_3"}
         assert entity_ids_1.scenario_ids == {"scenario_id_1", "scenario_id_2", "scenario_id_3"}
         assert entity_ids_1.cycle_ids == {"cycle_id_1", "cycle_id_2", "cycle_id_3"}
+
+    def test_union_all(self):
```
toan-quach: Not sure what the union_all is for

toan-quach (changes requested):
We can consider calling export directly from the entity: `scenario.export(folder_path)` too!

src/taipy/core/_manager/_manager.py
jrobinAV: The export format will always be JSON. I do agree that the folder_path is a better implementation. It makes the function usage self explanatory.

src/taipy/core/_repository/_repository.py
```diff
@@ -115,6 +115,18 @@ def _search(self, attribute: str, value: Any) -> Optional[Entity]:
         """
         raise NotImplementedError
 
+    @abstractmethod
+    def _export(self, entity_id: str, **kwargs):
```
jrobinAV: same remark on the signature. the folder_path should make it easier to understand.

src/taipy/core/cycle/_cycle_manager.py
jrobinAV: Yes, I agree. The 'children' naming is more consistent with the new naming convention you proposed, @toan-quach

src/taipy/core/taipy.py
jrobinAV: We only plan to export as JSON files. 

In this ticket, we are only considering exporting scenarios. We may want to expand the feature to other entities in the future. But not right now.

jrobinAV (dismissed):

jrobinAV (approved):

src/taipy/core/common/_entity_ids.py
```diff
@@ -11,6 +11,8 @@
 
 from __future__ import annotations
 
+from typing import Set
```
trgiangdo, line 14: This line should be discarded since `_union_all` is removed.

toan-quach (approved):

PR #329, opened 2022-10-02
Use ubuntu-latest instead of 18.04

joaoandre-avaiga:
We were using an old version of ubuntu. There is a option to update the version to Ubuntu 20.04 which is already 2 years old. What do you think about just using `ubuntu-latest` as we already use the latest version for windows and mac.

In case you prefer using a fixed version, let me know and I can update the PR.


jrobinAV (approved):
Neat

PR #332, opened 2022-10-04
Feature/ Ability to export the local files of a single scenario from SQLite Repositories

trgiangdo:


src/taipy/core/_repository/_sql_repository.py
```diff
@@ -187,3 +188,19 @@ def __model_to_entity(self, file_content):
         data = json.loads(file_content, cls=_CustomDecoder)
         model = self.model.from_dict(data)  # type: ignore
         return self._from_model(model)
+
+    def _export(self, entity_id: str, folder_path: str):
+        folder: pathlib.Path = pathlib.Path(folder_path)
```
jrobinAV:
```suggestion
    def _export(self, entity_id: str, folder_path: Union[str, Path]):
        if isinstance(folder_path, str):
            folder_path = pathlib.Path(folder_path)
```

folder_path could be a Union[str, Path].

jrobinAV (dismissed):

jrobinAV (approved):

Issue #152, opened 2022-04-10
Expose all parent entities


jrobinAV:
**What would that feature address**
- Propose an API to expose all the parent entities of a data node (tasks, pipelines, scenarios, and cycles) i.e. all the entities referencing directly or indirectly the data node.
- Propose an API to expose all the parent entities of a task (pipelines, scenarios, and cycles) i.e. all the entities referencing directly or indirectly the data node.
- Propose an API to expose all the parent entities of a pipeline(scenarios and cycles) i.e. all the entities referencing directly or indirectly the data node.

PR #333, opened 2022-10-06
Feature/#152 expose get_parents function

toan-quach:
#152 
This ticket includes:
- Adding auto-reloading for parent_ids
- exposing the get_parents() API in taipy.py
- fix bugs in parent_ids assigning:
  - the proper flow of assigning parent_ids should be only to assign the parent_ids to the child entity when the parent entity is created (getting the children entities -> creating the parent entity -> assigning the parent_ids of the children entity) 
  - this helps prevent assigning new parent id to the children parent_ids when the parent entity is not created, for reference: the pipeline_id is generated new all the time but it doesn't guarantee that the pipeline entity will be created, therefore an unknown pipeline_id could be assigned to the child task entity by mistake
  - although assigning scenario parent to pipeline children does not have this issue as the scenario is guaranteed to be created new every time but for the purpose of being consistent, it will also follow this flow

### DIFF
--- a/src/taipy/core/pipeline/_pipeline_manager.py
+++ b/src/taipy/core/pipeline/_pipeline_manager.py
@@ -93,8 +93,17 @@ class _PipelineManager(_Manager[Pipeline]):
             owner_id,
             {scenario_id} if scenario_id else None,
         )
+        for task in tasks:
+            task._parent_ids.update([pipeline_id])
+        cls.__save_tasks(tasks)
         cls._set(pipeline)
         return pipeline
+
+    @classmethod
+    def __save_tasks(cls, tasks):
+        task_manager = _TaskManagerFactory._build_manager()
+        for i in tasks:
+            task_manager._set(i)
 
     @classmethod
     def _submit(
--- a/src/taipy/core/taipy.py
+++ b/src/taipy/core/taipy.py
@@ -12,11 +12,12 @@
 import pathlib
 import shutil
 from datetime import datetime
-from typing import Any, Callable, Dict, List, Optional, Union
+from typing import Any, Callable, Dict, List, Optional, Set, Union
 
 from taipy.config.config import Config
 from taipy.logger._taipy_logger import _TaipyLogger
 
+from .common._entity import _Entity
 from .common.alias import CycleId, DataNodeId, JobId, PipelineId, ScenarioId, TaskId
 from .config.pipeline_config import PipelineConfig
 from .config.scenario_config import ScenarioConfig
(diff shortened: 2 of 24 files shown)


src/taipy/core/pipeline/_pipeline_manager.py
```diff
@@ -93,9 +93,18 @@ def _get_or_create(cls, pipeline_config: PipelineConfig, scenario_id: Optional[S
             owner_id,
             {scenario_id} if scenario_id else None,
         )
+        for task in tasks:
```
toan-quach, line 96: I think we can refactor this and turn it into a shared function in the _Manager class, something like ```__set_parent_ids_for_children_entities(child_entities: List[Entity]): ...``` and set it to NotImplemented for Scenario, Cycle Manager. What do you think? @jrobinAV

src/taipy/core/taipy.py
```diff
@@ -497,3 +498,51 @@ def export_scenario(
         _ScenarioManagerFactory._build_manager()._export(scenario_id, folder_path)
     for job_id in entity_ids.job_ids:
         _JobManagerFactory._build_manager()._export(job_id, folder_path)
+
+
+def get_parents(
+    entity: Union[TaskId, DataNodeId, PipelineId, Task, DataNode, Pipeline], parent_dict=None
+) -> Dict[str, Set[_Entity]]:
+    """Get parents of the entity from the entity or its identifier.
```
```suggestion
    """Get parents of an entity from the entity or its identifier.
```

src/taipy/core/taipy.py
```diff
@@ -497,3 +498,51 @@ def export_scenario(
...
+
+    Parameters:
+        entity (Union[TaskId^, DataNodeId^, PipelineId^, Task, DataNode, Pipeline]): The entity of its identifier
+            to get the parent.<br/>
+    Returns:
+        Dict[str, Set[_Entity]]: The dictionary of parents
+        matching the corresponding entity. An empty dictionary if the entity does not have parents.
```
jrobinAV: We should describe the structure of the dictionary. What are the keys?

jrobinAV (dismissed):
minor remarks on the reference manual documentation

src/taipy/core/pipeline/_pipeline_manager.py
jrobinAV: I don't think we will benefit from this. We will lose some readability. No? What would we gain?

toan-quach: Ah agree! Initially I thought that since the code is repeated, maybe we can group them together in the parent Manager class, but I'm not entirely convinced on the idea. But yes, we will lose readability without gaining anything!

src/taipy/core/taipy.py
toan-quach: @jrobinAV 
I've changed it to `Get the parents of an entity from itself or its identifier.`, what do you think?

src/taipy/core/taipy.py
toan-quach: I have changed it to `The dictionary of parents matching the corresponding entity with the key is the level of the parent for example 'scenarios', 'pipelines', 'tasks' and the value is a list of the parent entities.`. What do you think?

src/taipy/core/taipy.py
jrobinAV: Yes, it s better.

src/taipy/core/taipy.py
```diff
@@ -497,3 +498,52 @@ def export_scenario(
...
+) -> Dict[str, Set[_Entity]]:
+    """Get the parents of an entity from itself or its identifier.
+
+    Parameters:
+        entity (Union[TaskId^, DataNodeId^, PipelineId^, Task, DataNode, Pipeline]): The entity or its identifier
+            to get the parents.<br/>
+    Returns:
+        Dict[str, Set[_Entity]]: The dictionary of parents matching the corresponding entity with the key
+        is the level of the parent for example 'scenarios', 'pipelines', 'tasks' and the value is a list
+        of the parent entities. An empty dictionary if the entity does not have parents.
```
jrobinAV: I find it very complex and difficult to understand. 
Tell me what you think about this proposal.
```suggestion
        Dict[str, Set[_Entity]]: The dictionary of all parent entities.
            They are grouped by their type (`Scenario^`, `Pipelines^`, or `tasks^`) so each key
            corresponds to a level of the parents and is among the following list: 
            'scenarios', 'pipelines', or 'tasks'. 
            An empty dictionary is returned if the entity does not have parents.<br/>
            Example: The following instruction returns all the pipelines that include the
            datanode identified by "my_datanode_id".
            `taipy.get_parents("id_of_my_datanode")["pipelines"]`
```

toan-quach: @jrobinAV 
I think it's better! But we didn't include the set of parent entities as the value of the dictionary, how about `They are grouped by their type (`Scenario^`, `Pipelines^`, or `tasks^`) so each key corresponds to a level of the parents and the value is a set of the parent entities.`

jrobinAV: Yes that s better.

jrobinAV (approved):

Issue #337, opened 2022-10-10
Update Datanode cache when file is written (or update)


jrobinAV:
- [x] For Datanodes associated with a file, the cache can also use the last edit timestamp of the file to check if the data node is up-to-date or not. The concerned data nodes are Excel, CSV, JSON, and Pickle, (maybe SQLlite?, SQLTable?, MongoCollection?)
- [x] Unit tests
- [x] Documentation
- [x] end2end demo

toan-quach:
@jrobinAV 
It seems like SQLite and Mongodb don't support storing "last modified date" at the moment. If we want to use something like that, we will need to implement it ourselves. :D

jrobinAV:
> @jrobinAV It seems like SQLite and Mongodb don't support storing "last modified date" at the moment. If we want to use something like that, we will need to implement it ourselves. :D

OK let s postpone that for now. Thx!

PR #339, opened 2022-10-11
Feature/#337 dn cache with file modified date

toan-quach:
#337

### DIFF
--- a/tests/core/data/test_sql_table_data_node.py
+++ b/tests/core/data/test_sql_table_data_node.py
@@ -56,6 +56,7 @@ class TestSQLTableDataNode:
                 },
             },
         )
+
     if util.find_spec("pymysql"):
         __properties.append(
             {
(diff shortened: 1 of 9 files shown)


jrobinAV (dismissed):

tests/core/data/test_sql_table_data_node.py
```diff
@@ -127,40 +132,27 @@ def test_read(self, mock_read_as, mock_read_as_pandas_dataframe, mock_read_as_nu
             properties=properties,
         )
 
-        assert sql_data_node_as_pandas.read() == "pandas"
+        assert sql_data_node_as_pandas._read() == "pandas"
```
jrobinAV: Why don't you use the read method anymore?
Maybe we need to test both methods, no?

toan-quach: Hi, something went wrong with my git and pre-commit, I have changed it back to read(). Can you have a look again?

jrobinAV (approved):

tests/core/data/test_sql_table_data_node.py
jrobinAV: Good to go!

Issue #334, opened 2022-10-07
Make DefaultJSONEncoder and DefaultJSONDecoder consistent


jrobinAV:
Today the encoder can handle enums and datetimes but the decoder can't.

The decoder and the encoder should handle the same types. 

- [ ] Add a unit test for each type
- [ ] Ensure the documentation is up-to-date

PR #340, opened 2022-10-11
Feature/#334 Make DefaultJSONEncoder and DefaultJSONDecoder consistent 

trgiangdo:
https://github.com/Avaiga/taipy-core/issues/334

### DIFF
--- a/tests/core/data/test_json_data_node.py
+++ b/tests/core/data/test_json_data_node.py
@@ -181,12 +181,7 @@ class TestJSONDataNode:
         data = [MyEnum.A, MyEnum.B, MyEnum.C]
         json_dn.write(data)
         read_data = json_dn.read()
-        assert read_data[0].value == 1
-        assert read_data[0].name == "A"
-        assert read_data[1].value == 2
-        assert read_data[1].name == "B"
-        assert read_data[2].value == 3
-        assert read_data[2].name == "C"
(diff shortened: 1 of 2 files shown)


jrobinAV (dismissed):
great !

tests/core/data/test_json_data_node.py
```diff
@@ -159,35 +171,24 @@ def test_write_non_serializable(self, json_file):
...
-            A = 1
-            B = 2
-            C = 3
-
         json_dn = JSONDataNode("foo", Scope.PIPELINE, properties={"default_path": json_file})
         data = [MyEnum.A, MyEnum.B, MyEnum.C]
         json_dn.write(data)
         read_data = json_dn.read()
-        assert read_data == [1, 2, 3]
+        assert read_data == [MyEnum.A, MyEnum.B, MyEnum.C]
```
toan-quach, line 187: Great! Do you think we can also add the dictionary to make sure we cover as much data type in the test as possible?

toan-quach (dismissed):

jrobinAV (approved):

tests/core/data/test_json_data_node.py
trgiangdo: The `test_write()` test the write method with a dictionary, so I think we're good on this

Issue #335, opened 2022-10-07
make mypy happy again 


jrobinAV:
Mypy is showing some errors in particular on checkers.

Mypy should be happy again

PR #342, opened 2022-10-12
Feature/#335 make mypy happy

toan-quach:
#335

### DIFF
--- a/src/taipy/core/config/checkers/_data_node_config_checker.py
+++ b/src/taipy/core/config/checkers/_data_node_config_checker.py
@@ -58,24 +58,26 @@ class _DataNodeConfigChecker(_ConfigChecker):
             if storage_type in DataNodeConfig._REQUIRED_PROPERTIES:
                 required_properties = DataNodeConfig._REQUIRED_PROPERTIES[storage_type]
                 if storage_type == DataNodeConfig._STORAGE_TYPE_VALUE_SQL:
-                    if engine := data_node_config.properties.get(DataNodeConfig._REQUIRED_DB_ENGINE_SQL_PROPERTY):
-                        if engine == DataNodeConfig._REQUIRED_DB_ENGINE_SQLITE:
-                            required_properties = [
-                                DataNodeConfig._REQUIRED_DB_NAME_SQL_PROPERTY,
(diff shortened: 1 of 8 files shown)


src/taipy/core/config/checkers/_data_node_config_checker.py
```diff
@@ -22,7 +24,7 @@ def __init__(self, config: _Config, collector: IssueCollector):
         super().__init__(config, collector)
 
     def _check(self) -> IssueCollector:
-        data_node_configs = self._config._sections[DataNodeConfig.name]
+        data_node_configs: Dict[str, DataNodeConfig] = self._config._sections[DataNodeConfig.name]
         for data_node_config_id, data_node_config in data_node_configs.items():
```
toan-quach, line 28: @jrobinAV 
This file has several failed mypy tests in line 61, 70, 78, 93, 94, 110 with errors related to Optional[Any] (I believe it's related to the variable data_node_config used in each function), can't explain the problem though. Can you have a look?

jrobinAV (approved):

src/taipy/core/config/checkers/_data_node_config_checker.py
jrobinAV: Apparently, mypy detects that data_node_config.properties can be None. It is built from **properties in the Section class in the taipy-config repo. when no kwargs are passed, then properties is not an empty dict but None. And so mypy tells us that some functions like "in" or "indexation" are incompatible with None type.

I just pushed a fix in the taipy-config repo. It should be solved.

jrobinAV (dismissed):

src/taipy/core/config/checkers/_data_node_config_checker.py
toan-quach: @jrobinAV 
I have tried running again but the problem is still there. I also tried adding type Dict to the @property properties in Section with no help at all!

jrobinAV: I don't have it anymore. Are you sure you pulled the latest develop version of taipy-config?

jrobinAV: OK, I finally fixed the issue in the current PR as well. Can you try again?

jrobinAV (approved):

Issue #343, opened 2022-10-12
DefaultCustomDocument should be documented.


jrobinAV:
In `taipy.core.common.default_custom_document.py` the class `DefaultCustomDocument` is public and not documented (ref man).

If the class `DefaultCustomDocument` is made to be manipulated by the developer user, then it must be documented (and probably referenced in the `MongoDataNode` documentation (ref man and user man)).
Otherwise, the class DefaultCustomDocument should be renamed to `_DefaultCustomDocument` (as well as the module `default_custom_document` to `_default_custom_document.py`) in order to hide it in the ref manual.

Unrelated question: Why is `DefaultCustomDocument` in the common package while it is only used with json data nodes?

trgiangdo:
The `DefaultCustomDocument` is actually for `MongoCollectionDataNode`.

About if `DefaultCustomDocument` is made to be manipulated by the developer user, the answer is no.
However, I think it's good to reference it somehow in the document, so that when a user defines a mongo_cfg without `custom_document`, he/she can see how the data is actually being handled.

Also, I put it in the common package because I planned to add a `custom_document` to JSONDataNode latter, but now I think it will overlap with DefaultJSONEncoder and DefaultJSONDecoder.

jrobinAV:
After discussion, it seems that `DefaultCustomDocument` will be manipulated by the developer when he/she does a `my_mongo_datanode.read()`.
The `DefaultCustomDocument` should have a docstring documentation for the refmanual.

We also proposed to ensure to have: 
- A sentence in the user manual (in the mongo datanode config section) to explain the default document's behavior.
- more specific examples in the datanode entity page (the read section).

PR #344, opened 2022-10-14
Add documentation for the DefaultCustomDocument class

trgiangdo:
https://github.com/Avaiga/taipy-core/issues/343

### DIFF
--- a/src/taipy/core/data/mongo.py
+++ b/src/taipy/core/data/mongo.py
@@ -43,7 +43,16 @@ class MongoCollectionDataNode(DataNode):
         edit_in_progress (bool): True if a task computing the data node has been submitted
             and not completed yet. False otherwise.
         properties (dict[str, Any]): A dictionary of additional properties. Note that the
-            _properties_ parameter must at least contain an entry for _"db_name"_ and _"collection_name"_.
+            _properties_ parameter must at least contain an entry for _"db_name"_ and _"collection_name"_:
+
+            - _"db_name"_ `(str)`: The database name.\n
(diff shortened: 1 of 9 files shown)


src/taipy/core/data/mongo.py
```diff
@@ -43,7 +43,7 @@ class MongoCollectionDataNode(DataNode):
         edit_in_progress (bool): True if a task computing the data node has been submitted
             and not completed yet. False otherwise.
         properties (dict[str, Any]): A dictionary of additional properties. Note that the
-            _properties_ parameter must at least contain an entry for _"db_name"_, _"collection_name"_, and _"custom_document"_.
+            _properties_ parameter must at least contain an entry for _"db_name"_ and _"collection_name"_.
```
jrobinAV: custom_document is still used even if it is not mandatory. Can you add a small comment saying it is optional, and is a class representing a mongo document to read/write from the mongo collection.

trgiangdo: Since there is other optional parameters, should we also explain those? Like the hostname, port, or the extra_args?

jrobinAV: Indeed. Even if the user should not realy use this API since he/she will build datanodes through the config API. Still, it would be better.

trgiangdo: The config API is imported dynamically in taipy-core, so currently there is no ref manual for that (I think). It would be better if we can include those. But for now I think this is sufficient.

jrobinAV (approved):

PR #345, opened 2022-10-14
rename parent_id to owner_id for sqlrepo, added test for sqlrepo

toan-quach:
- [x] **Too many open files**: the occurs when running pytest (probably the behavior is inconsistent as Giang said it happens from time to time but not always). Can try to reproduce by run `pytest` on a single file then rerun `pytest` on all tests, but, as mentioned previously, the behavior is inconsistent so it's not guaranteed that it will happen again. This causes all the remaining test to fail. _(This probably is a problem in macos (potentially also Linux) where the OS limited a number of files to open at the same time (256). Still needs to investigate further the reason for a sudden surge of files opened)_]
- [x] **Got multiple values for keyword argument 'scope'**: to reproduce, same as above. This errors showed up in multiple tests
- [x] **Function is not an attribute of task**: in test_raise_when_trying_to_delete_unfinished_job() of `tests/core/job/test_job_manager_with_sql_repo.py`. I think we had this issues several times for other entities, I'm certain we did have that for DataNode. How to fix it, the last time this happened, the problem was in another place. Not sure if it's the same this time but we should take that into account as well.
- [x] **Job status being completed while still in lock**: `test_raise_when_trying_to_delete_unfinished_job` in `tests/core/job/test_job_manager_with_sql_repo.py`
- [x] **sqlite3.OperationalError: database is locked**: `test_hard_delete` in `tests/core/task/test_task_manager_with_sql_repo.py`. When run alone, it returns ModelNotFound for the task, the task_id in the job model was different from the task_id of the task that was submitted

## Previous notes
Current failed tests include:

- [x] recreating existing pipeline with the same config and owner id (it shouldn't create new pipeline) -> also applies for tasks, this is due to the function `__get_entities_by_config_and_owner` in `_SQLRepository` couldn't detect the already created pipeline/task
- [x] cannot detect raise `NoModelFound` error (in `_TaskManager` and `_JobManager`) -> the delete function in `_SQLRepository` doesn't raise the error

To look at later:
- [x] some more failed tests with the `data` folder (not yet carefully evaluated)

Warning messages:
- [x] When running `pytest tests/core/job`: it produces a warning `sqlite3.ProgrammingError: SQLite objects created in a thread can only be used in that same thread. The object was created in thread id 6185201664 and this is thread id 4300113280.`


jrobinAV (commented):
Great. 
I cannot approve yet because of the failing tests, but you know the pb.

.gitignore
```diff
@@ -82,6 +82,7 @@ dataset
 # Filesystem default local storage
 .data/
 .my_data/
+None
```
trgiangdo, line 85: Is this line added by mistake?

tests/core/pipeline/test_pipeline_manager_with_sql_repo.py
```diff
@@ -0,0 +1,525 @@
...
+    dn_output_config_scope_scenario = Config.configure_data_node("my_output", "in_memory", scope=Scope.SCENARIO)
+    task_config = Config.configure_task(
+        "task_config", print, dn_input_config_scope_scenario, dn_output_config_scope_scenario
+    )
+    pipeline_config = Config.configure_pipeline("pipeline_config_1", [task_config])
+
+    # Scope is scenario
+    pipeline_1 = _PipelineManager._get_or_create(pipeline_config)
+    assert len(_PipelineManager._get_all()) == 1
+    # TODO: failed because couldn't detect the already created pipeline
```
trgiangdo: This should have been fixed already, I believe.

.gitignore
toan-quach: Ah nope, when running pytest, a None file is created (holds the code for SQLite since I didn't specify the path for it)

tests/core/pipeline/test_pipeline_manager_with_sql_repo.py
toan-quach: agree! Please leave your comment here for now so I can remember to remove the line later :D

tests/conftest.py
```diff
@@ -288,6 +288,8 @@ def init_managers():
     _TaskManagerFactory._build_manager()._delete_all()
     _JobManagerFactory._build_manager()._delete_all()
     _CycleManagerFactory._build_manager()._delete_all()
+    if os.path.exists("./None"):
+        os.remove("./None")
```
trgiangdo: This line is causing failing test on SQL repo tests. @toan-quach Can you check again?

Error message:
> PermissionError: [WinError 32] The process cannot access the file because it is being used by another process: './None'

src/taipy/core/_repository/_sql_repository.py
```diff
@@ -58,7 +59,9 @@ def __init__(
 
         try:
             # More sql databases can be easily added in the future
-            self.engine = create_engine(f"sqlite:///{properties.get('db_location')}")
+            self.engine = create_engine(
+                f"sqlite:///{properties.get('db_location')}?check_same_thread=False", poolclass=StaticPool
```
jrobinAV, line 63: What are these arguments for?

src/taipy/core/_repository/_sql_repository.py
```diff
@@ -182,7 +196,7 @@ def __update_entry(self, entry, model):
         self.session.commit()
 
     def __to_entity(self, entry: _TaipyModel) -> Entity:
-        return self.__model_to_entity(entry.document)
+        return self.__model_to_entity(entry.document) if entry else None
```
jrobinAV, line 197: It should be useless, no?

src/taipy/core/data/_data_manager.py
```diff
@@ -10,11 +10,12 @@
 # specific language governing permissions and limitations under the License.
 
 import os
-from typing import Dict, Iterable, Optional, Set, Union
+from typing import Dict, Iterable, List, Optional, Set, Union
 
 from taipy.config.common.scope import Scope
 
 from .._manager._manager import _Manager
+from ..common._entity_ids import _EntityIds
```
jrobinAV: ?

src/taipy/core/data/_data_manager.py
```diff
@@ -10,11 +10,12 @@
 # specific language governing permissions and limitations under the License.
 
 import os
-from typing import Dict, Iterable, Optional, Set, Union
+from typing import Dict, Iterable, List, Optional, Set, Union
```
jrobinAV: ?

src/taipy/core/_repository/_sql_repository.py
```diff
@@ -102,7 +105,11 @@ def _save(self, entity: Entity):
         self.__insert_model(model)
 
     def _delete(self, model_id: str):
-        self.session.query(_TaipyModel).filter_by(model_id=model_id).delete()
+        number_of_deleted_entries = self.session.query(_TaipyModel).filter_by(model_id=model_id).delete()
+
+        if not number_of_deleted_entries:
+            raise ModelNotFound(self.model, model_id)
```
jrobinAV: why?

tests/conftest.py
```diff
@@ -226,6 +227,7 @@ def setup():
 @pytest.fixture(scope="function", autouse=True)
 def teardown():
     init_scheduler()
+    init_config()
```
jrobinAV, line 240: Why did you add this init_config ?

I suspect this could be a problem. Let me summarize what happens when a test runs: 
- First, conftest.setup runs: 
    - We initialize the config
    - We initialize the scheduler using the **default config**
    - We initialize managers using the **default config** and ensure default repos are empty
- Then, the test runs: 
    - We modify the config and use the SQL repos
    - We re-build schedulers and managers
    - We create some entities in SQL repo
- Finally, conftest.teardown runs:
    - We re init scheduler using a **modified config**
    - We re-initialize the config
    - We re-initialize the managers using the **default config** !!! and ensure default repos are empty. Not the SQL repos !!!
    - We re-initialize the config
    
Depending on the order of all the tests to run, this could lead to some unexpected entities in the SQL repository. Am I correct ?

src/taipy/core/_repository/_sql_repository.py
trgiangdo: By default, when SQLite is created in one thread, it can not be use elsewhere.

The `poolclass=StaticPool` makes sure that there is only one connection among all threads.
And `?check_same_thread=False` allows the connection to be used by any thread.

src/taipy/core/_repository/_sql_repository.py
trgiangdo: Is it? I'm not sure.
Can the `entry` be `None`?

src/taipy/core/_repository/_sql_repository.py
trgiangdo: This is to raise an exception if the model you want to delete is not exist.

The same goes for FS and Mongo repo.

tests/conftest.py
trgiangdo: Yes, we are worrying that the problem lies on how we init/re-init the config and repository. The additional `init_config()` doesn't seem to be the problem though. We need to test this more thoroughly.

src/taipy/core/_repository/_sql_repository.py
jrobinAV: OK thanks.

src/taipy/core/_repository/_sql_repository.py
jrobinAV: you are right it can be None!

src/taipy/core/_repository/_sql_repository.py
jrobinAV: Can't we just log a warning ? if we try to delete something that has already been deleted, it is not a major error. No ?

toan-quach: I think this behavior is consistent with the filesystem repo
image

src/taipy/core/data/_data_manager.py
toan-quach: I have removed the unnecessary imports

src/taipy/core/data/_data_manager.py
toan-quach: I have removed the unnecessary imports

src/taipy/core/_repository/_sql_repository.py
jrobinAV: Great!

jrobinAV: OK thanks.

trgiangdo (dismissed):
I'm still skeptical about the fact that the `db_location` is not required, which produced a `None` file.

Anyway, for now, this look good to me. Great work 🎉

tests/core/job/test_job_manager.py
```diff
@@ -306,7 +306,7 @@ def test_cancel_single_running_job():
         assert_true_after_1_minute_max(lambda: _SchedulerFactory._dispatcher._nb_available_workers == 1)
         assert_true_after_1_minute_max(job.is_running)
         _JobManager._cancel(job.id)
-        assert_true_after_1_minute_max(job.is_running)
+        assert_true_after_1_minute_max(job.is_running)  # TODO: probably the job was succeeded and this test failed
```
jrobinAV: Should we remove the TODO comment?

jrobinAV (dismissed):
OK for me!

tests/core/job/test_job_manager.py
toan-quach: right! my bad I forgot!

trgiangdo (dismissed):

trgiangdo (approved):

jrobinAV (approved):

PR #346, opened 2022-10-17
bug/ fix expiration_date not accessing modified date

toan-quach:


src/taipy/core/data/data_node.py
```diff
@@ -181,10 +181,13 @@ def validity_period(self, val):
     @property  # type: ignore
     @_self_reload(_MANAGER_NAME)
     def expiration_date(self) -> datetime:
-        if not self._last_edit_date:
+        last_edit_date = self.last_edit_date
```
toan-quach, line 184: I want to improve this part since we're reloading the dn twice here, any ideas would help :D

src/taipy/core/data/data_node.py
```diff
@@ -181,10 +181,13 @@ def validity_period(self, val):
     @property  # type: ignore
     @_self_reload(_MANAGER_NAME)
     def expiration_date(self) -> datetime:
-        if not self._last_edit_date:
```
trgiangdo, line 184: Is there any specific reason why we didn't call the getter for `last_edit_date` here before?

src/taipy/core/data/data_node.py
```diff
@@ -181,10 +181,13 @@ def validity_period(self, val):
...
+        last_edit_date = self.last_edit_date
+        validity_period = self._validity_period
+
+        if not last_edit_date:
             raise NoData(f"Data node {self.id} has not been written yet.")
 
-        return self._last_edit_date + self.validity_period if self.validity_period else self._last_edit_date
```
trgiangdo, line 187: Here, we called the getter for `validity_period` but not `last_edit_date`.

trgiangdo (commented):
I just have a few questions about the old code.

src/taipy/core/data/data_node.py
toan-quach: The expiration_date is already reloading the entity, so by calling _last_edit_date, we are already getting the latest data and calling last_edit_date will only trigger the self reload function again without much use. But now the last_edit_date also checks for the system file modification date and not only the date we're storing in the entity. So we need to call last_edit_date instead of _last_edit_date

src/taipy/core/data/data_node.py
toan-quach: I don't have an answer for this as I think _validity_period should have been enough :D

src/taipy/core/data/data_node.py
jrobinAV: I don't have any idea at this point.

jrobinAV (approved):

src/taipy/core/data/data_node.py
trgiangdo: Can we remove the `@_self_reload(_MANAGER_NAME)` wrapper here? Since `last_edit_date = self.last_edit_date` will reload the data node.

toan-quach: we certainly can, but it doesn't help improve the readability of the code. By that I mean we are calling self.last_edit_date while also calling self._validity_period. Can be confusing for others to understand

PR #359, opened 2022-11-10
Feature/#353 Add version number property to entities

trgiangdo:
https://github.com/Avaiga/taipy-core/issues/353

In this PR:
- `version` property was added to all entity class
- All entity managers and repositories are updated with `version`
- Update the tests to test the `version` property.

For now, the `version` is saved to `Config.global_config.version`.

The method to set a specific version is out of this PR scope.

PR #360, opened 2022-11-15
feat: versioning module

joaoandre-avaiga:


src/taipy/core/_version/_version_fs_repository.py
```diff
@@ -0,0 +1,26 @@
...
+from ._version import _Version
+from ._version_model import _VersionModel
+
+
+class _VersionFSRepository(_FileSystemRepository):
+    def __init__(self):
+        super().__init__(_VersionModel, "version", self._to_model, self._from_model)
+
+    def _to_model(self, version: _Version):
+        return _VersionModel(id=version.id, config=version.config)
```
jrobinAV: the type of version.config is Config while the expected type of _VersionModel.config is Dict.

I don't get how the conversion is done.

src/taipy/core/_version/_version_manager.py
```diff
@@ -0,0 +1,32 @@
...
+
+class _VersionManager(_Manager[_Version]):
+    _repository = _VersionRepositoryFactory._build_repository()  # type: ignore
+
+    @classmethod
+    def _create(cls, id: str, config: Config) -> _Version:
+        if cls._get(id) is not None:
+            raise VersionAlreadyExists(f"Version {id} already exists")
+
+        # TODO: replace with config export to new format
```
jrobinAV: Should the conversion to the new format be done in the repository and not in the manager?

src/taipy/core/_version/_version_fs_repository.py
joaoandre-avaiga: This is where we would use the new config export method. I put some TODOs in the code to change that when the config export is done,

jrobinAV: Great!

jrobinAV (dismissed):
I would like to wait for the export method to be implemented before merging the PR.

By the way, I believe `export` is not a good name for such a method. The purpose is to convert the Config singleton into a serializable format. Not really to export it. I believe the method should be internal.

joaoandre-avaiga:
> I would like to wait for the export method to be implemented before merging the PR.
> 
> By the way, I believe `export` is not a good name for such a method. The purpose is to convert the Config singleton into a serializable format. Not really to export it. I believe the method should be internal.

I agree, thinking about it now, we should have started with the config serialization. This PR should be merged after the config is done.

joaoandre-avaiga:
I've updated the PR to be compliance with the new serializer on taipy config. I have tested localy and it works for creating, loading and listing versions, for the FS Repositoy. It shouldn't be too hard to expand to other repositories, however I think it's best to first merge all the PRs and do the integration needed for a test end to end. 

This PR will fail because the new serializer is not merged yet on taipy config. I wish we had the time to do this before I went on vacation, but I think the code is very simple and Giang will do a great job integrating and improving anything if necessary. 

Feel free to do any adjustments to the code as you see fit.

See you guys soon. Cheers!

trgiangdo:
This PR is merged from https://github.com/Avaiga/taipy-core/pull/359 so it's pretty large.

To summarize the PR:
- The `"version"` property is added to all entities except for Cycle, due to Cycle is attach to Scenario.
- The internal `_Version` class represents the Version entity, with `version.config` is a JSON serialized string of Config singleton.
- The `_VersionManager` is responsible for create, manage the versions.
- The `version_cli.py` manages how the user will interact with the versioning system via console.

The current use-case:
- When the user start Core service, which will block Config update, the version_cli is call (on `core/_core.py`).
- With FS repo, the current version and the development version is save to a file named "version/version.json". This is so that the Manager can quickly access the current or dev version without the need for checking the created/modified date, or check for a flag.
- Dev mode: All entities with the development version is deleted. The development version is set to the current version.
- Experiment mode: A new version number will be set as the current version.
- If switch from experiment mode back to development mode, the old development version will be clean, which keep the experiment version safe.

src/taipy/core/_version/_version_cli.py
```diff
@@ -0,0 +1,101 @@
...
+    _JobManagerFactory._build_manager()._delete_by_version(version_number)
+    _ScenarioManagerFactory._build_manager()._delete_by_version(version_number)
+    _PipelineManagerFactory._build_manager()._delete_by_version(version_number)
+    _TaskManagerFactory._build_manager()._delete_by_version(version_number)
+    _DataManagerFactory._build_manager()._delete_by_version(version_number)
+
+
+@click.command()
+@click.option(
+    "--development",
```
jrobinAV, line 29: I would go for --dev instead of --development

```suggestion
    "--dev",
```

src/taipy/core/_version/_version_cli.py
```diff
@@ -0,0 +1,101 @@
...
+    "-d",
+    "mode",
+    flag_value="development",
+    default=True,
+    help="Execute Taipy application in DEVELOPMENT mode. This is the default behavior.",
```
jrobinAV:
```suggestion
    help="Execute Taipy application in `dev` mode. This is the default behavior.",
```

src/taipy/core/_version/_version_cli.py
```diff
@@ -0,0 +1,101 @@
...
+    _TaskManagerFactory._build_manager()._delete_by_version(version_number)
+    _DataManagerFactory._build_manager()._delete_by_version(version_number)
+
+
+@click.command()
+@click.option(
+    "--development",
+    "-d",
+    "mode",
+    flag_value="development",
```
jrobinAV, line 33:
```suggestion
    flag_value="dev",
```

src/taipy/core/_version/_version_cli.py
```diff
@@ -0,0 +1,101 @@
...
+    default=True,
+    help="Execute Taipy application in DEVELOPMENT mode. This is the default behavior.",
+)
+@click.option(
+    "--experiment", "-e", "mode", flag_value="experiment", help="Execute Taipy application in EXPERIMENTAL mode."
```
jrobinAV:
```suggestion
    "--experiment", "-e", "mode", flag_value="experiment", help="Execute Taipy application in `experiment` mode."
```

src/taipy/core/_version/_version_cli.py
```diff
@@ -0,0 +1,101 @@
...
+)
+@click.option(
+    "--version-number",
+    "-v",
+    "_version_number",
+    type=str,
+    default=None,
+    help="The version number when execute and save in experimental mode. If not provided, a random version name is used.",
```
jrobinAV:
```suggestion
    help="The version number when execute and save in `experiment` mode. If not provided, a random version name is used.",
```

src/taipy/core/_version/_version_cli.py
```diff
@@ -0,0 +1,101 @@
...
+    flag_value="development",
+    default=True,
+    help="Execute Taipy application in DEVELOPMENT mode. This is the default behavior.",
+)
+@click.option(
+    "--experiment", "-e", "mode", flag_value="experiment", help="Execute Taipy application in EXPERIMENTAL mode."
+)
+@click.option(
+    "--version-number",
+    "-v",
```
jrobinAV: Usually `-v` is used as a shortcut for `--version` option which returns the version of the tool. IT is confusing for me to use it as a setter for a version number.

src/taipy/core/_version/_version_cli.py
```diff
@@ -0,0 +1,101 @@
...
+    default=None,
+    help="The version number when execute and save in experimental mode. If not provided, a random version name is used.",
+)
+@click.option(
+    "--override",
+    "-o",
+    "_override",
+    is_flag=True,
+    help='Override the version specified by "--version-number" if existed. Default to False.',
+)
```
jrobinAV: Do you have a use case for that? It seems error-prone to me.

src/taipy/core/_version/_version_cli.py
```diff
@@ -0,0 +1,101 @@
...
+
+    elif mode == "development":
+        curren_version_number = _VersionManager.get_development_version()
+        _VersionManager.set_development_version(curren_version_number)
+        override = True
+
+        clean_all_entities_by_version(curren_version_number)
+        _TaipyLogger._get_logger().info(f"Development mode: Clean all entities with version {curren_version_number}")
+
+    else:
```
jrobinAV: What about having by default a `--production` mode?

src/taipy/core/_core.py
```diff
@@ -34,5 +35,7 @@ def run(self, force_restart=False):
         """
         Start a Core service. This method is blocking.
         """
+        version_cli(standalone_mode=False)
```
jrobinAV: What is `standalone_mode=False` for?

src/taipy/core/_manager/_manager.py
```diff
@@ -40,6 +40,13 @@ def _delete_many(cls, ids: Iterable):
         """
         cls._repository._delete_many(ids)
 
+    @classmethod
+    def _delete_by_version(cls, version_number: str):
+        """
+        Deletes entities by a list of ids.
+        """
```
jrobinAV: Docstring to update

src/taipy/core/config/data_node_config.py
```diff
@@ -30,13 +31,14 @@ class DataNodeConfig(Section):
...
     Attributes:
-        id (str):  Unique identifier of the data node config. It must be a valid Python variable name.
+        id (str): Unique identifier of the data node config. It must be a valid Python variable name.
         storage_type (str): Storage type of the data nodes created from the data node config. The possible values
             are : "csv", "excel", "pickle", "sql_table", "sql", "mongo_collection", "generic", "json", "parquet" and "in_memory".
             The default value is "pickle".
             Note that the "in_memory" value can only be used when `JobConfig^`.mode is "standalone".
         scope (Scope^):  The `Scope^` of the data nodes instantiated from the data node config. The default value is
             SCENARIO.
+        cacheable (bool): If True, indicates that the CSV data node is cacheable. The default value is _False_.
```
jrobinAV: It s not for CSV only.

src/taipy/core/data/csv.py
```diff
@@ -38,6 +39,7 @@ class CSVDataNode(DataNode):
         parent_ids (Optional[Set[str]]): The identifiers of the parent tasks or `None`.
         last_edit_date (datetime): The date and time of the last modification.
         job_ids (List[str]): The ordered list of jobs that have written this data node.
+        version (str): The string indicates the version number of the config. The default version is "latest".
```
jrobinAV:
```suggestion
        version (str): The string indicates the application version of the data node to instantiate. If not provided, the current version is used.

```

src/taipy/core/_version/_version_cli.py
trgiangdo: This will require a systematic update.

How about `--experiment` mode, should we set it to `--exp` for consistency?
It's a bit weird though.

src/taipy/core/_version/_version_cli.py
trgiangdo: Let say you've already had a version "2.1" saved, and working on "2.2". However, in the process, you found out that there is an error on "2.1" and need to update that.

By default, we don't want the user to directly override an old version to prevent data loss.
However, in this case, we can allow that.

src/taipy/core/_version/_version_cli.py
trgiangdo: What do you mean by "--production" mode?
I'm setting "--dev" mode by default.

src/taipy/core/_core.py
trgiangdo: `click` library by default will create a separate process then end it right after the command is finish.
`standalone_mode=False` will prevent this behavior.

src/taipy/core/_version/_version_cli.py
joaoandre-avaiga: +1 for `--dev`, I think is more consistent with other tools with similar functionality, But I would suggest keeping both alias: `--development` and `--dev`

src/taipy/core/_version/_version_cli.py
joaoandre-avaiga: Maybe providing a little more context on what development mode is would be helpfull.

src/taipy/core/_version/_version_cli.py
joaoandre-avaiga: Same comment for experimental mode. I know it will be eventually on the official doc, but I think it would be nice for the user to have it hand on the cli.

src/taipy/core/_version/_version_cli.py
trgiangdo: "--production" mode would be the current state of Taipy, which mean we will not re-create the entities, but more like cache it and add new one if the config changed.

src/taipy/core/_core.py
jrobinAV: OK thx

jrobinAV (approved):

PR #361, opened 2022-11-15
Feature/ Block Config update when Taipy is running

trgiangdo:
When running the core service, a JobDispatcher is created in a separated thread.

Without `core.run()`, each time `tp.submit()` is called, a different JobDispatcher is created and deleted when the job is done.

My proposal is to call `Config.block_update()` whenever `build_dispatcher()` is called.
Then, we only unblock it when taipy application is terminated. 

This is not ideal though.

### DIFF
--- a/src/taipy/core/_scheduler/_dispatcher/_job_dispatcher.py
+++ b/src/taipy/core/_scheduler/_dispatcher/_job_dispatcher.py
@@ -42,7 +42,6 @@ class _JobDispatcher(threading.Thread):
         threading.Thread.__init__(self, name="Thread-Taipy-JobDispatcher")
         self.daemon = True
         self.scheduler = scheduler
-        print("dispatcher is created")
         Config.block_update()
 
     def start(self):
--- a/src/taipy/core/_scheduler/_dispatcher/_job_dispatcher.py
+++ b/src/taipy/core/_scheduler/_dispatcher/_job_dispatcher.py
@@ -8,6 +8,7 @@
 # Unless required by applicable law or agreed to in writing, software distributed under the License is distributed on
 # an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the
 # specific language governing permissions and limitations under the License.
+
 import threading
 from abc import abstractmethod
 from multiprocessing import Lock
@@ -41,6 +42,10 @@ class _JobDispatcher(threading.Thread):
         threading.Thread.__init__(self, name="Thread-Taipy-JobDispatcher")
         self.daemon = True
         self.scheduler = scheduler
+        Config.block_update()
+
+    def __del__(self):
+        Config.unblock_update()
 
     def start(self):
         """Start the dispatcher"""
--- a/src/taipy/core/_scheduler/_dispatcher/_job_dispatcher.py
+++ b/src/taipy/core/_scheduler/_dispatcher/_job_dispatcher.py
@@ -138,6 +138,7 @@ class _JobDispatcher(threading.Thread):
         try:
             if mode != JobConfig._DEVELOPMENT_MODE:
                 Config._applied_config = _TomlSerializer()._deserialize(config_as_string)
+                Config.block_update()
             inputs: List[DataNode] = list(task.input.values())
             outputs: List[DataNode] = list(task.output.values())
             fct = task.function
--- a/src/taipy/core/_scheduler/_dispatcher/_job_dispatcher.py
+++ b/src/taipy/core/_scheduler/_dispatcher/_job_dispatcher.py
@@ -42,10 +42,8 @@ class _JobDispatcher(threading.Thread):
         threading.Thread.__init__(self, name="Thread-Taipy-JobDispatcher")
         self.daemon = True
         self.scheduler = scheduler
+        print("dispatcher is created")
         Config.block_update()
-
-    def __del__(self):
-        Config.unblock_update()
 
     def start(self):
         """Start the dispatcher"""
--- a/tests/core/_scheduler/test_scheduler.py
+++ b/tests/core/_scheduler/test_scheduler.py
@@ -40,6 +40,7 @@ from tests.core.utils import assert_true_after_1_minute_max
 @pytest.fixture(scope="function", autouse=True)
 def reset_configuration_singleton():
     yield
+    Config.unblock_update()
     Config._python_config = _Config()
     Config._file_config = None
     Config._env_file_config = None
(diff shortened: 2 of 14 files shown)


trgiangdo:
The test is failing due to https://github.com/Avaiga/taipy-config/pull/17 is not merged yet.

src/taipy/core/_scheduler/_dispatcher/_job_dispatcher.py
```diff
@@ -41,6 +42,7 @@ def __init__(self, scheduler: _AbstractScheduler):
         threading.Thread.__init__(self, name="Thread-Taipy-JobDispatcher")
         self.daemon = True
         self.scheduler = scheduler
+        Config.block_update()
```
jrobinAV, line 45: This is a minor nonblocking remark.
I don't see any use case for now, but it feels more logical to me to:
- call `Config.block_update()` in the start method instead of the init.
- call `Config.unblock_update()` at the end of the run method, just before returning.

What do you think?

jrobinAV (dismissed):

src/taipy/core/_scheduler/_dispatcher/_job_dispatcher.py
trgiangdo: On Development mode, there will be no call to the `start` or `run` method though.

I did implemented this in some of the previous version, which make much more sense on standalone mode.

jrobinAV: ok thx.

tests/core/_scheduler/test_scheduler.py
```diff
@@ -472,7 +482,6 @@ def modified_config_task(n):
 
 def test_can_exec_task_with_modified_config():
```
toan-quach, line 484: @trgiangdo @jrobinAV 
Do you think this test is still relevant in this PR? Maybe what we should test here is can't modify during task execution?

tests/core/job/test_job_manager.py
```diff
@@ -60,9 +60,10 @@ def lock_multiply(lock, nb1: float, nb2: float):
 
 def test_create_jobs():
     Config.configure_job_executions(mode=JobConfig._DEVELOPMENT_MODE)
-    _SchedulerFactory._build_dispatcher()
     task = _create_task(multiply, name="get_job")
 
+    _SchedulerFactory._build_dispatcher()
```
toan-quach, line 65: @trgiangdo 
why do we need to move the build dispatcher call here? It doesn't seem like it makes any difference right? If it doesn't make any difference then I have no problem with it, if it does, can you briefly explain it to me?

toan-quach: Same with many other places apparently

trgiangdo: The use-case for block update config, for now, is to block all configuration update after running the core service (or submit a task), which equivalent to start a JobDispatcher.

That's the reason behind moving all `_build_dispatcher()` after all of the configuration.

tests/core/_scheduler/test_scheduler.py
trgiangdo: I don't get your idea, but I think it's not relevant to this PR anyway.
The test is only modified so that we only build the dispatcher after all configurations.

toan-quach: oh I think the idea of this test is to see if we can modify the config during job execution (meaning a dispatcher has already been built and I'm assuming it's something we don't want to do). So the purpose of this test is contradicting with the PR itself

tests/core/job/test_job_manager.py
toan-quach: Ah my bad, I forgot the create_task function has some config in it, nvm :D

tests/core/_scheduler/test_scheduler.py
trgiangdo: Oh, it's passing, which is not good.

So the problem is that we cannot modify the config by `Config.config_global_app()`, but by directly assign using `Config.global_config.property = ...` is still allowed. I need to fix this.

toan-quach: Exactly! :D

trgiangdo: Actually, the idea of this test is to check in stand-alone mode, if we modify the `global_config` before submitting the task, will the `global_config` update? There is no setter here.

I'll create a new test for this purpose.

jrobinAV: In standalone mode, task execution is performed in a dedicated subprocess that does not share the memory with the main process. That means the Config singleton is not shared. Each subprocess builds another instance of the Config. For that purpose, we serialize the config as a string in the main process and pass it as a parameter to the subprocess. Then at the very beginning of the subprocess task execution, we rebuild a new Config using the string parameter.

The purpose of this test is to ensure that each subprocess takes into account the Config modifications done in the main Process when it builds a new instance of the Config.

With your new implementation, the _ConfigBlocker is a new Singleton that is not passed to the subprocesses. That means in each subprocess, the config will not be blocked. After rebuilding a new Config in each subprocess, we should block the Config. [See line 138 of file _job_dispatcher.py](https://github.com/Avaiga/taipy-core/blob/develop/src/taipy/core/_scheduler/_dispatcher/_job_dispatcher.py#L138)

jrobinAV: May be this explanation should be added as a comment in the test to explain the purpose.

trgiangdo: This is exactly the problem that I'm facing right now. Thank you 😆

trgiangdo: So for now, every time we dispatch a job on stand-alone mode, we block the config update as @jrobinAV suggested.
I've tried to add `Config.block_update()` on `JobDispatcher.start()` with the idea that we only run the block function when we create a new thread. However, this doesn't work. @toan-quach If you have spare time, pls take a look and see if we can improve this in any possible way.

I've also added a test for this.
Since the exception when running task is caught by the logger, I cannot check for the exception, so I add a `pytest.raises` directly inside the task function.

Again, the test is failing due to https://github.com/Avaiga/taipy-config/pull/20.

jrobinAV (dismissed):
Go to go for me!

trgiangdo:
@jrobinAV sorry, 2 last commits on block data node config setter. This should be completely done now.

toan-quach (approved):

jrobinAV (approved):
Great!

PR #363, opened 2022-11-16
fix excel dn bugs from integration tests

toan-quach:


src/taipy/core/data/excel.py
```diff
@@ -286,11 +287,18 @@ def _read_as_modin_dataframe(
             return modin_pd.DataFrame()
 
     def _write(self, data: Any):
-        if isinstance(data, Dict) and all([isinstance(x, (pd.DataFrame, modin_pd.DataFrame)) for x in data.values()]):
+        if isinstance(data, Dict) and all(
+            [isinstance(x, (pd.DataFrame, modin_pd.DataFrame, np.ndarray)) for x in data.values()]
```
trgiangdo, line 291: I remember you said that a dictionary of custom object also causes the same problem as `np.array`, right?

toan-quach: Yep yep correct, but I dont have a solution for custom objects as it can be anything! Can even be a string or an object of a Row class or an int or float! Numpy is a bit more straightforward although it also depends on the typing of its elements but overall it's manageable to some extend

trgiangdo: If that's the case, there should be an exception imo.

jrobinAV (commented):
Could you write a unit test that is failing without the fix and is successful with it?

src/taipy/core/data/excel.py
toan-quach: @jrobinAV 
What do you think? I think Giang has a good point. It's quite tricky, but maybe we can check to see if the custom object is of type "primitive" or not (int, float, str and the likes), if yes, we can convert them into pd dataframe and save it to excel file. If it's a "truly" custom class object, we will throw an exception or log a warning that user is trying to write a custom class to a csv file.

jrobinAV (approved):

Issue #351, opened 2022-10-31
Implement a ParquetDataNode


jrobinAV:
The purpose of the ticket is to create a ParquetDataNode as a data node able to use parquet format to store the data.

[Parquet format documentation](https://parquet.apache.org/docs/)

The new class should inherit from the abstract class DataNode, and should implement all the necessary methods, in particular the _write, _read, storage_type. 
The storage type key used in the config should be 'parquet'.
There should be no expected required properties.
The optional properties should be:
- `path`: corresponding to the path of the parquet storage. If no path is provided, an optional `default_path` property could be used to populate the `path` attribute.
- the `exposed_type` property should accept `pandas` (the default value) or numpy.

Note that the CSVDataNode has a very similar behavior. This could be used as a reference class to implement the ParquetDataNode.

Issue #350, opened 2022-10-31
Implement TabularDataNode


jrobinAV:
Implement an abstract TabularDataNode class to factorize common methods and attributes for tabular storage formats.

The actual implementations would be CSVDataNode, AbstractSQLDataNode, ExcelDataNode, and ParquetDataNode.

PR #364, opened 2022-11-17
Feature/#351 Implement ParquetDataNode

arcanaxion:
#351 

Hello Taipy team! This is my attempt at implementing ParquetDataNode. Some things to note:

- This wraps the convenient `pandas.read_parquet` and `pandas.to_parquet` methods
- pyarrow (or fastparquet) was not added as a dependency since pandas doesn't do so either. The aforementioned pandas methods will raise an error informing the user to install either of those packages

Also, for the moment, ParquetDataNode only accepts two new optional parameters. Note that:
  - `compression`: The pandas default is `"snappy"`. I did not re-declare this default. `None` is meaningful, and means no compression. This implementation just excludes passing the `compression` parameter to `pandas.to_parquet` if not explicitly passed to the data node constructor -- thus using the pandas default.
  - `columns` : The `columns` parameter only applies to the read. The reason for this is to support the use case where a user may only want to read a subset of columns (parquet does this efficiently), but write the entire DataFrame (they may be accessing the parquet file externally).

Please let me know how I can improve this!


src/taipy/core/data/parquet.py
```diff
@@ -0,0 +1,166 @@
...
+            )
+
+    def _read(self):
+        if self.properties[self.__EXPOSED_TYPE_PROPERTY] == self.__EXPOSED_TYPE_PANDAS:
+            return self._read_as_pandas_dataframe()
+        if self.properties[self.__EXPOSED_TYPE_PROPERTY] == self.__EXPOSED_TYPE_MODIN:
+            return self._read_as_modin_dataframe()
+        if self.properties[self.__EXPOSED_TYPE_PROPERTY] == self.__EXPOSED_TYPE_NUMPY:
+            return self._read_as_numpy()
+
```
trgiangdo, line 165: In Taipy, most of the data node also allows the user to set the `exposed_type` as a custom (data) class.
Can you add a `read_as()` method similar to [`read_as()` method in CSVDataNode](https://github.com/Avaiga/taipy-core/blob/develop/src/taipy/core/data/csv.py#L146) for this purpose?

src/taipy/core/config/data_node_config.py
```diff
@@ -383,6 +397,47 @@ def _configure_json(
...
         return Config.sections[DataNodeConfig.name][id]
 
+    @staticmethod
+    def _configure_parquet(
+        id: str,
+        default_path: str = None,
+        exposed_type=_DEFAULT_EXPOSED_TYPE,
+        columns: Optional[List[str]] = None,
+        scope=_DEFAULT_SCOPE,
+        cacheable: bool = False,
```
trgiangdo, line 410:
```suggestion
        cacheable: bool = False,
        compression: str = "snappy",
```
The compression property and its default value should be specifically defined.
This will require an update on "parquet.py" too.

src/taipy/core/config/data_node_config.py
```diff
@@ -383,6 +397,47 @@ def _configure_json(
...
+        Returns:
+            `DataNodeConfig^`: The new Parquet data node configuration.
+        """
+        section = DataNodeConfig(
+            id,
+            DataNodeConfig._STORAGE_TYPE_VALUE_PARQUET,
+            scope=scope,
+            cacheable=cacheable,
+            default_path=default_path,
+            columns=columns,
```
trgiangdo:
```suggestion
            columns=columns,
            compression=compression,
```

src/taipy/core/data/parquet.py
```diff
@@ -0,0 +1,166 @@
...
+        self._path = value
+        self.properties[self.__PATH_KEY] = value
+
+    def _check_exposed_type(self, exposed_type):
+        if exposed_type not in self.__VALID_STRING_EXPOSED_TYPES:
+            raise InvalidExposedType(
+                f"Invalid exposed type {exposed_type}. Supported values are {', '.join(self.__VALID_STRING_EXPOSED_TYPES)}"
+            )
+
+    def _read(self):
```
trgiangdo, line 158: With Parquet format, I believe that we can read from multiple .parquet file from a folder.
Can you help us add that functionality here?

src/taipy/core/data/parquet.py
```diff
@@ -0,0 +1,166 @@
...
+    def _read_as_numpy(self):
+        return self._read_as_pandas_dataframe().to_numpy()
+
+    def _read_as_pandas_dataframe(self) -> pd.DataFrame:
+        return pd.read_parquet(self._path, columns=self.properties[self.__COLUMNS_PROPERTY])
+
+    def _read_as_modin_dataframe(self) -> modin_pd.DataFrame:
+        return modin_pd.read_parquet(self._path, columns=self.properties[self.__COLUMNS_PROPERTY])
+
+    def _write(self, data: Any):
```
trgiangdo, line 175: Again, I believe that we can write a DataFrame to multiple small .parquet file in a folder.
Can you help us add that functionality here?

Pipfile
```diff
@@ -26,6 +26,7 @@ tox = ">=3.24"
 types-toml = ">=0.10.0"
 autopep8 = "*"
 mongomock = ">=4.1.2"
+pyarrow = "*"
```
trgiangdo, line 29: Just want to make sure, without `pyarrow` or `fastparquet`, the user can not use ParquetDatanode, right?
If so, it should be a required package.

src/taipy/core/data/parquet.py
trgiangdo, line 175: Can you add a `write_with_column_names()` like [this method in CSVDataNode](https://github.com/Avaiga/taipy-core/blob/develop/src/taipy/core/data/csv.py#L203), so that the user can write with different column name?

src/taipy/core/config/data_node_config.py
```diff
@@ -383,6 +397,47 @@ def _configure_json(
...
+        cacheable: bool = False,
+        **properties,
+    ):
+        """Configure a new Parquet data node configuration.
+
+        Parameters:
+            id (str): The unique identifier of the new Parquet data node configuration.
+            default_path (str): The default path of the Parquet file.
+            exposed_type: The exposed type of the data read from Parquet file. The default value is `pandas`.
+            columns (Optional[List[str]]): If not None, only these columns will be read from the file.
```
trgiangdo: I'm just wondering, if this is only applied when reading, is it better to have this `columns` properties on `read()` method?
Since the user might want to read multiple times, each time with different set of columns, and we don't want the user to directly update the Config.

Let me know what you think.

src/taipy/core/data/parquet.py
```diff
@@ -0,0 +1,166 @@
...
+            Implemented as a timedelta. If _validity_period_ is set to None, the data_node is
+            always up-to-date.
+        edit_in_progress (bool): True if a task computing the data node has been submitted
+            and not completed yet. False otherwise.
+        path (str): The path to the Parquet file.
+        properties (dict[str, Any]): A dictionary of additional properties. The _properties_
+            must have a _"default_path"_ or _"path"_ entry with the path of the Parquet file:
+
+            - _"default_path"_ `(str)`: The default path of the Parquet file.\n
+            - _"columns"_ `(Optional[List[str]])`: If not None, only these columns will be read from the file.
```
trgiangdo: Here, we document the required properties of a ParquetDataNode that the user must provide.
`columns` and `compression` are optional though.

trgiangdo (commented):
Thank you for your help. Overall, this is a great first PR.

There are some remarks that I think will help improve this  a lot.

src/taipy/core/data/parquet.py
arcanaxion: Ah, I only implemented pandas (and modin) and numpy per #351. I guess the custom class implementation would then be something like `[fn(*row) for row in self._read_as_numpy()]`?

src/taipy/core/config/data_node_config.py
arcanaxion: Since I think we don't have a preference on what compression algorithm to use, I deliberately omitted defining a default value here so as to use whatever pandas has set (`"snappy"` in this case). But I agree with your feedback, I'll write a test then to check that the pandas default is `"snappy"`, so that we can change the default if/when they do.

src/taipy/core/data/parquet.py
arcanaxion: Gotcha, will do that

Pipfile
arcanaxion: That's right. Since this is quite an optional feature, I think we needn't make it a required package. Pandas also has it as an optional package. Perhaps we should have a discussion on this, if it hasn't already been had.

src/taipy/core/data/parquet.py
trgiangdo: Something like that 😄 You can check out [read_as() method in CSVDataNod](https://github.com/Avaiga/taipy-core/blob/develop/src/taipy/core/data/csv.py#L146)

src/taipy/core/data/parquet.py
arcanaxion: Sure.

https://github.com/Avaiga/taipy-core/blob/7381d79e0dbd5324aaefb4b5f2f4bc3666bd8054/src/taipy/core/data/csv.py#L216-L218

I'm not sure what these lines are for though -- do I copy this as well?

Pipfile
trgiangdo: I'm not sure, adding another required package might raise some problems.
@jrobinAV Any thought on this? `pyarrow` is a small package, so I think it should be fine.

src/taipy/core/config/data_node_config.py
arcanaxion: Ah yes that didn't occur to me! That's so much better. I guess I will need to override the `DataNode.read` and `DataNode.read_or_raise` methods in `ParquetDataNode` I think

src/taipy/core/data/parquet.py
trgiangdo: Line 216 is to update the last edit time made by Taipy.
Line 217-218 is to make sure that the list of jobs which are working on this data node is updated.

And yes, these line should be on ParquetDataNode too. Thank you 😁

src/taipy/core/data/parquet.py
arcanaxion: Oh, so the optional parameters should only be in the docstring for `DataNodeConfig._configure_parquet`? I included those optional parameters in the `ParquetDataNode` docstring as well because [CSVDataNode](https://github.com/Avaiga/taipy-core/blob/7381d79e0dbd5324aaefb4b5f2f4bc3666bd8054/src/taipy/core/data/csv.py#L52-L53) listed the `has_header` and `exposed_type` optional parameters there.

src/taipy/core/config/data_node_config.py
trgiangdo: Probably. But overriding the abstract class with new property is not the best idea. We should keep it consistent between all DataNode, or keep it, or change it all.

Anyway, in the case of `pandas` expose type, the user can just select the columns from the DataFrame.
With `numpy`, it will be indexing.
And with custom class, it will be defined by the properties of that class.

@arcanaxion What do you think?
And @jrobinAV @toan-quach, let me know what you think about this.

src/taipy/core/data/parquet.py
trgiangdo: Oh, yes, this is just a list of properties, and `"path"` is required. My bad, you are actually correct on this 😄
We need to rethink about `"columns"` and update this accordingly. Thank you.

src/taipy/core/config/data_node_config.py
trgiangdo: I mean, on the documentation for example, we don't want to define:
> "The default value for `compression` is the same as the default value defined by `pandas`".

We should explicitly state what is the default value.

The same goes for the configure method and the ParquetDataNode.

Pipfile
jrobinAV: The Default Pandas engine is `pyarrow`, maybe we can stick to `pyarrow` as well, and add it as a required package.
What would be the purpose of using `fastparquet`?

src/taipy/core/config/data_node_config.py
jrobinAV: The read method signature cannot change. It is used by taipy scheduler when executing a task. 
So the read method cannot use runtime parameters. Config properties can be used though.

If the user wants to read data providing specific parameters manually, we need to expose another method. It could be a method specific to ParquetDatanode, or we could override the filter method, or both.

jrobinAV (commented):
The few minor remarks we made could make the PR even better, but it is already very good. Thank you. 

What you proposed here works well, so the remarks could be implemented on a separate PR if you prefer. It s up to you.
Here is a summary of the remarks: 
- [x] Add custom exposed type.
- [x] Add the possibility to read/write from a folder instead of a file.
- [x] Expose new methods to read/write data with column and filter parameters.

Pipfile
arcanaxion: I think we could choose between these 3 approaches:

1. Allow users to specify the `engine` through the Pandas Parquet interface (`pd.DataFrame.to_parquet` and `pd.read_parquet`). Don't add either package as a required package, and preserve the Pandas error if either package isn't installed.
2. Add `pyarrow` (the Pandas default) as a required package. Continue using the Pandas Parquet interface.
3. Add `pyarrow` as a required package. Implement a new wrapper using `pyarrow` for reading/writing Parquet files; i.e., don't use the Pandas interface.

I think the third option will be considerably more work, and it seems a shame to do so when the Pandas interface is already robust and convenient. The second would differ from the first only in that `pyarrow` is added as a required package. We could also maintain the `engine` parameter, which would allow the user to optionally install `fastparquet` and pass `engine="fastparquet"` if they so desired.

What do you think?

trgiangdo: Does `fastparquet` have any advantage compare to `pyarrow` in processing parquet file?

src/taipy/core/data/parquet.py
arcanaxion: I think this functionality is quite specific and easily achievable in the Task // function rather than having a separate method for it. I've expressed this to @jrobinAV and I think he agrees, but do let me know if I should add it anyway.

Pipfile
arcanaxion: I don't know enough to evaluate it myself, and the sentiment online just [seems to depend on who you ask](https://stackoverflow.com/questions/51361356/a-comparison-between-fastparquet-and-pyarrow). I'd personally prefer preserving Pandas behaviour and allow the user to use fastparquet if they choose. Adding pyarrow as a dependency (if that's what we want to do) won't conflict with that.

arcanaxion:
Hello again! The latest changes should address all the suggestions. Now, parameters can be passed to the Pandas Parquet IO functions (`pd.read_parquet` and `pd.DataFrame.to_parquet`) **in increasing order of precedence** by:

1. "Top-level" ParquetDataNode config properties (`"engine"` and `"compression"`);
2. `"read_kwargs"` and `"write_kwargs"` dictionaries in config properties; and
3. `ParquetDataNode.read_with_kwargs` and `ParquetDataNode.write_with_kwargs` methods.

Writing to [several small Parquet files](https://arrow.apache.org/docs/python/generated/pyarrow.dataset.partitioning.html) can be done by passing `write_kwargs={"partition_cols": ["col1", "col2"]}`.

src/taipy/core/data/parquet.py
```diff
@@ -0,0 +1,219 @@
...
+    def read_with_kwargs(self, **kwargs):
+        """Read data node, with keyword arguments passed to `pandas.read_parquet`.
+
+        Parameters:
+            **kwargs: The keyword arguments are passed to `pandas.read_parquet`.
+        """
+
+        # return None if data was never written
+        if not self.last_edit_date:
+            return None
```
trgiangdo, line 215: I believe we should also log the warning as `DataNode.read()`.

The duplication is not good. However, this problem should be resolved in a different refactoring PR for https://github.com/Avaiga/taipy-core/issues/350.

```suggestion
            self.__logger.warning(
                f"Data node {self.id} from config {self.config_id} is being read but has never been " f"written."
            )
            return None
```

src/taipy/core/data/parquet.py
```diff
@@ -0,0 +1,219 @@
...
+        self._path = value
+        self.properties[self.__PATH_KEY] = value
+
+    def _check_exposed_type(self, exposed_type):
+        if isinstance(exposed_type, str) and exposed_type not in self.__VALID_STRING_EXPOSED_TYPES:
+            raise InvalidExposedType(
+                f"Invalid string exposed type {exposed_type}. Supported values are {', '.join(self.__VALID_STRING_EXPOSED_TYPES)}"
+            )
+
+    def _read(self, read_kwargs: Dict = dict()):
```
trgiangdo: The read method signature cannot be changed since Taipy scheduler uses it when executing a task. This is the same for `._read()` method though, since it is internal and used by `.read()`.

Therefore, `read_kargs` here is not a good idea.

The `self.properties[self.__EXPOSED_TYPE_PROPERTY]` should be sufficient.

trgiangdo (changes requested):
Hello Zac, thank you again for your great work.

With your new update, the feature is almost ready to merge to core.
To summarize my review:
 
- [x] Add custom exposed type. This is great.
- [x] Add the possibility to read/write from a folder. This is done by setting the path to a folder path, and customize with the new `write_kargs`, which is more flexible.
- [x] Expose new methods to read/write data with column and filter parameters. This is done via the new `read_kargs`.

There are some changes that need to be done before merging:
- [x] The `_read()` and `_write()` method should not accept `read_kargs` or `write_kargs` since they are internal and used by job scheduler.
- [ ] We must decide to use `pyarrow` or `fastparquet` as the default. If use `pyarrow`, we need to define it as a requirements in `setup.py` and `Pipfile`. The other should be on `extra_require` as an optional package.

Additionally:
- [ ] There are some inconsistency in the documentation. However, we can amend this in other PRs, where we discuss how to document for the new ParquetDataNode.

Thank you very much.

src/taipy/core/data/parquet.py
arcanaxion: Yup I understand that `DataNode.read`'s signature cannot be changed // overriden. 

Having the `read_kwargs` optional param in the `ParquetDataNode._read*` methods is to enable the `.read_with_kwargs` method (the 3rd way of passing arguments to `pandas.read_parquet`) to not have to repeat code by simply passing `read_kwargs` to `._read`. I don't know much about the internal workings, but I currently don't see why `._read` can't have optional params which `.read` need not care about.

If `._read(self)` should be maintained, I'm thinking it might then be better to create `.__read(self, read_kwargs: Dict = dict())` instead. Should I make this change?

arcanaxion:
Thanks for the review @trgiangdo! Are we sure about adding pyarrow or fastparquet as required packages? If so, I'm not sure which one either. Pandas has pyarrow as the default and it does seem to be more popular, but I've seen people suggesting fastparquet in certain cases.

I currently have pyarrow==10.0.0 and fastparquet==2022.11.0 taking up about 116 MB and 6.7 MB on disk respectively.

Perhaps @jrobinAV could weigh in on which package we should go with.

src/taipy/core/data/parquet.py
jrobinAV: - The problem here is that the `read` method must stay unchanged since it is in the parent class.
So the parameter passed to the `_read` method is always the default param (empty dict).
That means the `read_kwargs` will not be used when using `read` method. This is a bug.
However, the `read_kwargs` and `write_kwargs` are accessible through the `properties` variable. In particular, you do ensure in the `__init__` method, it is automatically populated even when the user does not specify it. This is good. That means the `_read` method can directly access the `properties` variable. It does not require getting an extra parameter. 

```suggestion
    def _read(self):
        read_kwargs = properties[self.__READ_KWARGS_PROPERTY]
```

At this stage, I don't see the purpose of having another method with the signature:
`.__read(self, read_kwargs: Dict = dict())`
only `_read` will use it.

The same for `_write` method.

jrobinAV (dismissed):

arcanaxion:
@jrobinAV The `read_kwargs` parameter in `._read` is different from the one in `self.properties[self.__READ_KWARGS_PROPERTY]`. Currently, `read_kwargs` is passed to `._read` by the `.read_with_kwargs` method. This is so that a user can pass a parameter like `columns=["name", "age"]` to `pandas.read_parquet` at runtime without modifying the config.

Kwargs passed to `.read_with_kwargs` override the ones in the config (properties).

jrobinAV:
> @jrobinAV The `read_kwargs` parameter in `._read` is different from the one in `self.properties[self.__READ_KWARGS_PROPERTY]`. Currently, `read_kwargs` is passed to `._read` by the `.read_with_kwargs` method. This is so that a user can pass a parameter like `columns=["name", "age"]` to `pandas.read_parquet` at runtime without modifying the config.
> 
> Kwargs passed to `.read_with_kwargs` override the ones in the config (properties).

Ah, OK, sorry. I did not get that. 
But still, the `read` method will always call the `_read` with default value `dict()`. No? So it will always override the config. Am I missing something else here? 
If I am right, I would do the opposite. I would implement `_read` so it calls _read_with_params with the config values. And I would expose `read_with_kwargs` with an extra parameter so the user can call it manually. Tell me if it makes sense to you ?

To summarize : 
- `read` can be used by task execution. It will call `_read` without extra params. '_read' will get params from config and call `_read_with_kwargs`.
- Users can manually call `read` method without extra params. It will run the same as above.
- Users can manually call `read_with_kwargs` with runtime parameters to override config params.

arcanaxion:
@jrobinAV I have made some changes, is this what you mean?

src/taipy/core/data/parquet.py
```diff
@@ -0,0 +1,222 @@
...
+    def read_with_kwargs(self, **read_kwargs):
+        """Read data node, with keyword arguments passed to `pandas.read_parquet`.
+
+        Parameters:
+            **read_kwargs: The keyword arguments are passed to `pandas.read_parquet`.
+        """
+
+        # return None if data was never written
+        if not self.last_edit_date:
+            self._DataNode__logger.warning(
```
trgiangdo, line 212:
```suggestion
            self.__logger.warning(
```
Is this by mistake?

src/taipy/core/data/parquet.py
```diff
@@ -0,0 +1,222 @@
...
+                f"Data node {self.id} from config {self.config_id} is being read but has never been " f"written."
```
trgiangdo:
```suggestion
                f"Data node {self.id} from config {self.config_id} is being read but has never been written."
```

src/taipy/core/data/parquet.py
```diff
@@ -0,0 +1,222 @@
...
+            )
+            return None
+
```
trgiangdo, line 216:
```suggestion

        kwargs = self.properties[self.__READ_KWARGS_PROPERTY]
        kwargs.update(read_kwargs)

```
The update is performed here so that we don't have duplication on `. _read_as_pandas_dataframe()` and `._read_as_modin_dataframe()`.

src/taipy/core/data/parquet.py
```diff
@@ -0,0 +1,222 @@
...
+    def _read_as_pandas_dataframe(self, read_kwargs: Dict) -> pd.DataFrame:
+        kwargs = self.properties[self.__READ_KWARGS_PROPERTY]
+        kwargs.update(read_kwargs)
+        return pd.read_parquet(self._path, **kwargs)
+
+    def _read_as_modin_dataframe(self, read_kwargs: Dict) -> modin_pd.DataFrame:
+        kwargs = self.properties[self.__READ_KWARGS_PROPERTY]
+        kwargs.update(read_kwargs)
+        return modin_pd.read_parquet(self._path, **kwargs)
```
trgiangdo:
```suggestion
        return modin_pd.read_parquet(self._path, **read_kwargs)
```

src/taipy/core/data/parquet.py
```diff
@@ -0,0 +1,222 @@
...
+        list_of_dicts = self._read_as_pandas_dataframe(read_kwargs).to_dict(orient="records")
+        return [custom_class(**dct) for dct in list_of_dicts]
+
+    def _read_as_numpy(self, read_kwargs: Dict):
+        return self._read_as_pandas_dataframe(read_kwargs).to_numpy()
+
+    def _read_as_pandas_dataframe(self, read_kwargs: Dict) -> pd.DataFrame:
+        kwargs = self.properties[self.__READ_KWARGS_PROPERTY]
+        kwargs.update(read_kwargs)
+        return pd.read_parquet(self._path, **kwargs)
```
trgiangdo:
```suggestion
        return pd.read_parquet(self._path, **read_kwargs)
```

src/taipy/core/data/parquet.py
```diff
@@ -0,0 +1,222 @@
...
+
+        # return None if data was never written
+        if not self.last_edit_date:
+            self._DataNode__logger.warning(
+                f"Data node {self.id} from config {self.config_id} is being read but has never been " f"written."
+            )
+            return None
+
+        if self.properties[self.__EXPOSED_TYPE_PROPERTY] == self.__EXPOSED_TYPE_PANDAS:
+            return self._read_as_pandas_dataframe(read_kwargs)
```
trgiangdo:
```suggestion
            return self._read_as_pandas_dataframe(kwargs)
```

src/taipy/core/data/parquet.py
```diff
@@ -0,0 +1,222 @@
...
+        if self.properties[self.__EXPOSED_TYPE_PROPERTY] == self.__EXPOSED_TYPE_MODIN:
+            return self._read_as_modin_dataframe(read_kwargs)
```
trgiangdo:
```suggestion
            return self._read_as_modin_dataframe(kwargs)
```

src/taipy/core/data/parquet.py
```diff
@@ -0,0 +1,222 @@
...
+        if self.properties[self.__EXPOSED_TYPE_PROPERTY] == self.__EXPOSED_TYPE_NUMPY:
+            return self._read_as_numpy(read_kwargs)
+        return self._read_as(read_kwargs)
```
trgiangdo:
```suggestion
            return self._read_as_numpy(kwargs)
        return self._read_as(kwargs)
```

trgiangdo (commented):
Just some typo and a small refactor.

I believe we can do the same for `._write()` method to call `.write_with_kwargs()`

src/taipy/core/data/parquet.py
arcanaxion: Haha this typo is in `DataNode.read` as well

src/taipy/core/data/parquet.py
arcanaxion: No, the `__logger` is defined in DataNode, so it's mangled to `_DataNode__logger` outside of that class. I too only noticed because a test was failing

src/taipy/core/data/parquet.py
arcanaxion: Made the changes accordingly, thanks!

arcanaxion:
> I believe we can do the same for `._write()` method to call `.write_with_kwargs()`

Done!

src/taipy/core/data/parquet.py
```diff
@@ -0,0 +1,222 @@
...
+        self.properties[self.__PATH_KEY] = value
+
+    def _check_exposed_type(self, exposed_type):
+        if isinstance(exposed_type, str) and exposed_type not in self.__VALID_STRING_EXPOSED_TYPES:
+            raise InvalidExposedType(
+                f"Invalid string exposed type {exposed_type}. Supported values are {', '.join(self.__VALID_STRING_EXPOSED_TYPES)}"
+            )
+
+    def _read(self):
+        return self.read_with_kwargs()
```
jrobinAV, line 159:
```suggestion
    def _read(self):
        read_kwargs = properties[self.__READ_KWARGS_PROPERTY]
        return self.read_with_kwargs(read_kwargs)
```

jrobinAV (commented):
I made a suggestion on _read method so it calls read_with_kwargs passing the config kwargs.

src/taipy/core/data/parquet.py
arcanaxion: I think `read_with_kwargs` needs to be the one that gets the config kwargs because it can be manually called at runtime. Even if we loaded the config kwargs in `_read`, we'd still need to load it in `read_with_kwargs` to address the manual case.

jrobinAV: We can use the `read` method for manual cases too.

arcanaxion: Yes but `DataNode.read` calls `ParquetDataNode._read` which calls `ParquetDataNode.read_with_kwargs` (which loads the config kwargs). The config kwargs are used in all exposed read methods (which I believe is the behaviour we want). The `ParquetDataNode.read_with_kwargs` method (compared to `DataNode.read`) only brings the added behaviour that new kwargs may be specified at runtime (such as for column names or filters)

jrobinAV: Yes OK. I understand!

jrobinAV (approved):

trgiangdo (approved):
Great

jrobinAV:
Thanks a lot @arcanaxion!

PR #367, opened 2022-11-23
fix #366 update last_edit_date when job fails or is abandoned

### LINKED ISSUE #366: BUG-Do not update `last_edit_date` for abandoned and failed jobs
When a job fails, the method `Job._unlock_edit_on_outputs` is called on the failed job and on all jobs to abandon. This method triggers the execution of the method `DataNode.unlock_edit` on all the outputs, which updates the `last_edit_date`. 

Consequently, the 'last_edit_date' is wrong on the output data nodes. 

This is a blocking issue in the specific case where a data node has its 'last_edit_date' updated while it has never been written. In this case, Taipy can try to read it, leading to an error thrown.

Requires a hotfix on release 2.0!
Requires a release note update on 2.0.4.

jrobinAV:


jrobinAV:
@toan-quach little reminder to approve (or not) this PR.

toan-quach (approved):
lgtm :D

PR #368, opened 2022-11-23
Fix/#366 hot fix

jrobinAV:
Please review it carefully since it must be merged directly in release/2.0

### DIFF
--- a/src/taipy/core/data/data_node.py
+++ b/src/taipy/core/data/data_node.py
@@ -28,7 +28,6 @@ from ..common._properties import _Properties
 from ..common._reload import _reload, _self_reload, _self_setter
 from ..common._warnings import _warn_deprecated
 from ..common.alias import DataNodeId, JobId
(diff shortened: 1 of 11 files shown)


src/taipy/core/data/data_node.py
```diff
@@ -291,26 +293,20 @@ def lock_edition(self):
     def unlock_edit(self, at: datetime = None, job_id: JobId = None):
```
toan-quach, line 293: since we will separate the last_edit_date from unlock_edit, we wouldn't need these 2 params anymore

toan-quach (approved):
lgtm! only a minor remark

src/taipy/core/data/data_node.py
jrobinAV: Absolutely, but for compatibility reasons, I kept the same signature, and I deprecated the two params.

PR #369, opened 2022-11-30
Fix/ Setup library for ParquetDataNode

trgiangdo:
Finishing https://github.com/Avaiga/taipy-core/issues/351.

In this PR:
- `fastparquet` is used as the default engine instead of `pyarrow`.
- Add test for both `fastparquet` and `pyarrow`.

Some existing problems with `fastparquet`:
1. Modin does support both for read and write Parquet. However, their internal code still check if arrow is installed or not. This is weird, and I’ve open an Issue on Modin (https://github.com/modin-project/modin/issues/5285).
2. The filter argument does not work with `fastparquet`, only work with `pyarrow`. It is not that `fastparquet` doesn’t support filter, just not though pandas interface.

@FabienLelaquais @jrobinAV @florian-vuillemot @toan-quach What do you think about these problems?

### DIFF
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -19,7 +19,7 @@ jobs:
         with:
           python-version: ${{ matrix.python-versions }}
 
-      - name: Pyodbc and Pymysql and Psycopg2
+      - name: Pyodbc and Pymysql and Psycopg2 and Pyarrow
         env:
           TOX_PARALLEL_NO_SPINNER: 1
         run: |
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -19,7 +19,7 @@ jobs:
         with:
           python-version: ${{ matrix.python-versions }}
 
-      - name: Pyodbc and Pymysql and Psycopg2 and Pyarrow
+      - name: Pyodbc and Pymysql and Psycopg2 and Fastparquet
         env:
           TOX_PARALLEL_NO_SPINNER: 1
         run: |
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -24,7 +24,7 @@ jobs:
       - name: Install Tox
         run: pip install tox
 
-      - name: Run linter and all tests except Pyodbc's, Pymysql's, Psycopg2's, and Pyarrow
+      - name: Run linter and all tests except Pyodbc's, Pymysql's, Psycopg2's, and Fastparquet
         if: matrix.python-versions == '3.10' && matrix.os == 'windows-latest'
         env:
           TOX_PARALLEL_NO_SPINNER: 1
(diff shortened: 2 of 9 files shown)


arcanaxion:
> 2\. The filter argument does not work with `fastparquet`, only work with `pyarrow`. It is not that `fastparquet` doesn’t support filter, just not though pandas interface.

Hi! Just to chime in on this, the fastparquet package handles filters differently from pyarrow. From my understanding, it only performs the filter with partitions (and also "row-groups" which I'm less familiar with), i.e., when the data is saved to a folder. The documentation isn't very clear about this, and I found out about it from [this post](https://stackoverflow.com/a/53541258) on stackoverflow.

A small example demonstrating this:

```python
import pandas as pd
df = pd.DataFrame({"id": range(10), "letters": list("abcdefghij")})

file_path = "df.parquet"
df.to_parquet(file_path)
partition_path = "df_partition" 
df.to_parquet(partition_path, partition_cols=["id"])

filters = [("id", "<", 5)]

pyarrow_file = pd.read_parquet(file_path, engine="pyarrow", filters=filters) # filter applied
fastparquet_file = pd.read_parquet(file_path, engine="fastparquet", filters=filters) # filter NOT applied
pyarrow_file.equals(fastparquet_file) # False

pyarrow_partition = pd.read_parquet(partition_path, engine="pyarrow", filters=filters) # filter applied
fastparquet_partition = pd.read_parquet(partition_path, engine="fastparquet", filters=filters) # filter applied
pyarrow_partition.equals(fastparquet_partition) # True
```

trgiangdo:
Update on `fastparquet` on Modin:
1. The PR to fix this problem is merged.
2. The next release of Modin is scheduled to Dec 7th, which is lucky for us.

jrobinAV:
OK. That sounds good. I guess we can wait for Dec 7th to make the decision.

jrobinAV:
On one hand, for `fastparquet`, filters work only for the partition columns (here "id"), but not on all columns (filtering on "letters" does not work). So that creates a coupling between the `write_kwargs` provided in the config and the usage of the filter method at runtime. 
This makes our API and its documentation more complex!
Supporting filtering on columns not used in partitioning requires more code to implement and maintain, and more performance to monitor. This is painful.

On the other hand, for `pyarrow`, filters work whatever the partition columns provided. The API is simpler, easier to document and maintain.

**I would definitely set `pyarrow` as the default engine! Make it required.**
**Even more, I would not support the full filtering feature with `fastparquet`**

trgiangdo:
> I would not support the full filtering feature with fastparquet

I think we still can provide `fastparquet` as an optional package.
Or if we get rid of `fastparquet`, the `engine` property should still exist, don't you think?

jrobinAV:
> > I would not support the full filtering feature with fastparquet
> 
> I think we still can provide `fastparquet` as an optional package. Or if we get rid of `fastparquet`, the `engine` property should still exist, don't you think?

Yes, I agree. We should make `fastparquet` optional. But in this case, we should not support the filtering on columns not provided in the config as partition columns. And it must be clearly documented.

jrobinAV (approved):

Issue #355, opened 2022-11-07
Dev mode : Modify the `_load_all` and `load_all_by` to take a version number into account


joaoandre-avaiga:
### Description

Change the repository APIs `_load_all` and `load_all_by` to list only the entities related to a version.

- Make sure to modify the manager APIs if necessary
- By default, the current version should be used
- Evaluate new performances in integration-testing repository

trgiangdo:
- Extend the behavior of `--experiment` mode as the current `--production` mode
- Change the `--production` mode behavior to be able to load entities from older production version.

PR #375, opened 2022-12-07
Feature/#355 - Modify the _load_all and load_all_by to take a version number into account

trgiangdo:
https://github.com/Avaiga/taipy-core/issues/355

### DIFF
--- a/src/taipy/core/_version/_version_manager.py
+++ b/src/taipy/core/_version/_version_manager.py
@@ -58,32 +58,32 @@ class _VersionManager(_Manager[_Version]):
         return cls._repository._load_all_by(by, version_number)
 
     @classmethod
-    def set_current_version(cls, version_number: str, override: bool):
+    def _set_current_version(cls, version_number: str, override: bool):
         cls.get_or_create(version_number, override)
         cls._repository._set_current_version(version_number)
 
     @classmethod
-    def get_current_version(cls) -> str:
+    def _get_current_version(cls) -> str:
         return cls._repository._get_current_version()
 
     @classmethod
-    def set_development_version(cls, version_number: str):
+    def _set_development_version(cls, version_number: str):
         cls.get_or_create(version_number, override=True)
         cls._repository._set_development_version(version_number)
 
     @classmethod
-    def get_development_version(cls) -> str:
+    def _get_development_version(cls) -> str:
         return cls._repository._get_development_version()
 
     @classmethod
-    def replace_version_number(cls, version_number):
+    def _replace_version_number(cls, version_number):
         if version_number is None:
             version_number = cls.__DEFAULT_VERSION
 
         if version_number == cls.__CURRENT_VERSION_NUMBER:
-            return cls.get_current_version()
+            return cls._get_current_version()
         if version_number in cls.__DEVELOPMENT_VERSION_NUMBER:
-            return cls.get_development_version()
+            return cls._get_development_version()
         if version_number in cls.__ALL_VERSION_NUMBER:
             return ""
         if version := cls._get(version_number):
--- a/src/taipy/core/_version/_version_manager.py
+++ b/src/taipy/core/_version/_version_manager.py
@@ -26,6 +26,8 @@ class _VersionManager(_Manager[_Version]):
     __CURRENT_VERSION_NUMBER = "current"
     __ALL_VERSION_NUMBER = ["all", ""]
 
+    __DEFAULT_VERSION = __CURRENT_VERSION_NUMBER
+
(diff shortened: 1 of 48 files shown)


trgiangdo:
The propose of the current PR:
- The `__to_entity` can accept `by` param as a list (as well as a string) so that we can check for both the version number and other condition.
- Both `_load_all` and `_load_all_by` will take `version_number` as parameter. The default value is "current" which means we only get the entities of the current version.
- This can hurt the performance of FS repo when `_load_all()`.

The PR is not running yet, just one of the direction to implement this feature.
Feel free to comment and discard this if this is not what we want.

src/taipy/core/_version/_version_manager.py
```diff
@@ -35,6 +43,20 @@ def get_or_create(cls, id: str, override: bool) -> _Version:
...
+
+    @classmethod
+    def _get_all_by(cls, by, version_number: Optional[str] = "all") -> List[_Version]:
+        """
+        Returns all entities based on a criteria.
+        """
+        return cls._repository._load_all_by(by, version_number)
+
     @classmethod
     def set_current_version(cls, version_number: str, override: bool):
```
jrobinAV: I believe this method should be private. Otherwise it should be documented.

src/taipy/core/_version/_version_manager.py
```diff
@@ -52,3 +74,19 @@ def set_development_version(cls, version_number: str):
     @classmethod
     def get_development_version(cls) -> str:
```
jrobinAV: same remark

src/taipy/core/_version/_version_manager.py
```diff
@@ -52,3 +74,19 @@ def set_development_version(cls, version_number: str):
...
         return cls._repository._get_development_version()
+
+    @classmethod
+    def replace_version_number(cls, version_number):
```
jrobinAV: same remark

jrobinAV (approved):
Globally the tests are showing that the feature works fine. So I approve the PR for merging.

However, I am a bit lost in understanding the difference between version_number set to None compared to "all" and "current".
What about making the version_number mandatory on the repositories API (either a real version_number, or "all")? So we enforce managers to provide the correct value when calling the repositories based on the version_manager.

I would also need to discuss with you the _core.py for an eventual refactoring in another PR.

src/taipy/core/cycle/_cycle_manager.py
```diff
@@ -41,6 +41,20 @@ def _create(
...
+        """
+        return cls._repository._load_all(version_number)
+
+    @classmethod
+    def _get_all_by(cls, by, version_number: Optional[str] = "all") -> List[Cycle]:
+        """
+        Returns all entities based on a criteria.
+        """
+        return cls._repository._load_all_by(by, version_number)
+
```
jrobinAV, line 57: This looks strange to me since the cycles are not versioned. Can't we just use the default implementation from the parent class ?

jrobinAV (dismissed):

src/taipy/core/cycle/_cycle_manager.py
trgiangdo: On `get_all()` and `get_all_by()` on parent `_Manager` class, `"version_number"` would be `"current"` since most of the manager will need to get entities on the current version by default.

The only 2 exceptions are Cycle and _Version which doesn't have a version number to it, so the `"version_number"` would be `"all"`.
`"all"` basically means that we do not check for version when load from repository.

src/taipy/core/_version/_version_manager.py
```diff
@@ -52,3 +74,19 @@ def set_development_version(cls, version_number: str):
...
+        if version_number is None:
+            version_number = cls.__DEFAULT_VERSION
```
trgiangdo, line 81: The `"version_number"` is set to None as default in many place is because it will be set to the __DEFAULT_VERSION here.
If __DEFAULT_VERSION changes (it should not but if), we don't need to worry to much.

The reason we need `"current"`, `"development"` instead of a real version number is that I don't want to call `_VersionManager.get_current_version()` on every repositories API. This function will handle that.

src/taipy/core/cycle/_cycle_manager.py
jrobinAV: OK.
Thx

trgiangdo:
The test is failing due to

> PermissionError: [WinError 32] The process cannot access the file because it is being used by another process: 'None'

This only appears on Windows.

jrobinAV (approved):

PR #378, opened 2022-12-09
Refactor conftest - Improving autouse fixture

trgiangdo:


tests/conftest.py
```diff
@@ -250,18 +255,10 @@ def tmp_sqlite(tmpdir_factory):
...
 
 
-@pytest.fixture(scope="function", autouse=True)
-def setup():
-    if Config.global_config.repository_type == "sql":
-        close_sql_database_session_connection()
-    init_config()
-    init_scheduler()
-    init_managers()
-    init_config()
```
jrobinAV: I would keep the setup. In case the previous test process is killed before the teardown method is executed

trgiangdo: In that case, we only need 1 fixture, with the `yield` command in the middle of all the `init_...()`.

If I'm correct, the `setup` and `teardown` naming in pytest doesn't affect when the fixture is run.

trgiangdo: Something like
```python
@pytest.fixture(scope="function", autouse=True)
def clean_repository():
    if Config.global_config.repository_type == "sql":
        close_sql_database_session_connection()
    init_config()
    init_scheduler()
    init_managers()
    init_config()
    
    yield

    if Config.global_config.repository_type == "sql":
        close_sql_database_session_connection()
    init_config()
    init_scheduler()
    init_managers()
    init_config()
```

jrobinAV: Yes absolutely. Something like that. Let me propose a slight change: 

```python
@pytest.fixture(scope="function", autouse=True)
def clean_repository():
    if Config.global_config.repository_type == "sql":
        close_sql_database_session_connection()
        init_managers() #  We do remove SQL data before switching to FS.
    init_config()
    init_scheduler()
    init_managers()
    init_config()
    
    yield

    if Config.global_config.repository_type == "sql":
        close_sql_database_session_connection()
        init_managers() #  We do remove SQL data before switching to FS.
    init_config()
    init_scheduler()
    init_managers()
    init_config()
```

jrobinAV (approved):

PR #389, opened 2022-12-14
Fix writing modin df to ParquetDataNode

arcanaxion:
Fixing an issue with writing modin DataFrames to ParquetDataNode.

I'm re-implementing the behavior from CSVDataNode which I wrongly removed previously. Without this, passing a modin df to pd.DataFrame causes it to lose the column names, which is not what we want.


trgiangdo (approved):
I believe the fix is great. Thank you very much

PR #394, opened 2022-12-23
bug/#392 fixed bug cycle dn has the same behavior as global dn

toan-quach:
**Note:** Please review the unit tests carefully as they can be very confusing and error-prone!

### DIFF
--- a/src/taipy/__init__.py
+++ b/src/taipy/__init__.py
@@ -13,11 +13,17 @@ from importlib.util import find_spec
 
 if find_spec("taipy"):
     if find_spec("taipy.config"):
-        from taipy.config import Config
-        from taipy.config import Scope
-        from taipy.config import Frequency
+        from taipy.config import Config, Frequency, Scope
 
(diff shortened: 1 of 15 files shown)


src/taipy/__init__.py
```diff
@@ -13,11 +13,17 @@
 
 if find_spec("taipy"):
     if find_spec("taipy.config"):
-        from taipy.config import Config
-        from taipy.config import Scope
-        from taipy.config import Frequency
+        from taipy.config import Config, Frequency, Scope
```
jrobinAV, line 16: It would be better to change the __init__ file in a separate PR so the change can easily be done in all repositories simultaneously.
In any case, it s just import refactoring. Nothing changed. right?

jrobinAV (approved):

src/taipy/__init__.py
toan-quach: @jrobinAV 
yep nothing changed! I think mypy or something else from the pre-commit made the change

PR #399, opened 2023-01-03
feature/#377 added log error when dn is not ready to be read

toan-quach:


src/taipy/core/_scheduler/_scheduler.py
```diff
@@ -177,9 +178,17 @@ def _is_blocked(obj: Union[Task, Job]) -> bool:
...
         data_manager = _DataManagerFactory._build_manager()
-        return any(not data_manager._get(dn.id).is_ready_for_reading for dn in data_nodes)
+
+        def get_dn_is_ready_for_reading(dn: DataNode):
+            dn = data_manager._get(dn.id)
+            if dn.is_ready_for_reading is None:
+                cls.__logger.error(f"{dn.id} is not ready to be read.")
+            return dn.is_ready_for_reading
+
+        return any(not get_dn_is_ready_for_reading(dn) for dn in data_nodes)
```
jrobinAV: `dn.is_ready_for_reading` can't be None. It is either True or False. I believe the msg is never logged.

Plus, sometimes, it is normal to have a blocked job. It is not an error but a warning.
I would go for a more precise msg. Maybe we can have something like that. What do you think?

```python
        def get_dn_is_ready_for_reading(dn: DataNode):
            dn = data_manager._get(dn.id)
            if dn.is_ready_for_reading is False:
                if not dn._last_edit_date:
                    if dn.storage_type() == "csv' or 'excel' or any other file
                        cls.__logger.warning(f"{dn.id} cannot be read because it has never been written. Hint: The data node may refer to a wrong path : {dn.path}")
                    else:
                        cls.__logger.warning(f"{dn.id} cannot be read because it has never been written.")
            return dn.is_ready_for_reading

        return any(not get_dn_is_ready_for_reading(dn) for dn in data_nodes)
```

src/taipy/core/_scheduler/_scheduler.py
```diff
@@ -177,9 +178,26 @@ def _is_blocked(obj: Union[Task, Job]) -> bool:
...
         """
+
         data_nodes = obj.task.input.values() if isinstance(obj, Job) else obj.input.values()
         data_manager = _DataManagerFactory._build_manager()
-        return any(not data_manager._get(dn.id).is_ready_for_reading for dn in data_nodes)
+
+        def get_dn_is_ready_for_reading(dn: DataNode):
+            dn = data_manager._get(dn.id)
+            if dn.is_ready_for_reading is False and not dn._last_edit_date:
+                if dn.storage_type() in [
```
trgiangdo, line 188: Probably we can add Parquet here as well.

And how about pickle?

tests/core/_scheduler/test_scheduler.py
```diff
@@ -16,6 +16,7 @@
 from datetime import datetime, timedelta
 from functools import partial
 from time import sleep
+from unittest import mock
```
trgiangdo:
```suggestion
```
This line is redundant, isn't it?

trgiangdo (dismissed):
LGTM

trgiangdo (approved):

Issue #397, opened 2022-12-30
Enforce the Core service for a core application


trgiangdo:
***Description***

The propose is to enforce the following steps for a core application.

1. Build the config (*manipulated by user*)
2. Check it
3. Lock it
4. Check the mode
5. Set up the version and eventually clean entities
6. Start the dispatcher
7. Use taipy-core APIs (*manipulated by user*)

We need to check before step 7 that Core.run **has been executed**.

***Caveats***

The notebook usage will be limited to a development mode.

***Documentation***
- [x] Add release notes
- [x] Add description on running taipy services page
- [x] Add ref manual

PR #400, opened 2023-01-03
Feat/#397 - Enforce the Core service for a core application

trgiangdo:
https://github.com/Avaiga/taipy-core/issues/397

With this PR, when a scenario (or task or pipeline) is submitted when the Core service is not running yet, a warning will be shown.

The `Core().stop()` may be used on notebook environment between cells of code.

However, in case [run both GUI and Core services as the docs](https://docs.taipy.io/en/latest/manuals/running_services/#running-different-taipy-services-together) or the Core service is started after submit, the warning will still be shown.
This is not good.


src/taipy/core/_scheduler/_dispatcher/_job_dispatcher.py
```diff
@@ -138,7 +137,6 @@ def _run_wrapped_function(cls, mode, config_as_string, job_id: JobId, task: Task
         try:
             if mode != JobConfig._DEVELOPMENT_MODE:
                 Config._applied_config = _TomlSerializer()._deserialize(config_as_string)
-                Config.block_update()
```
jrobinAV: I don't understand this change.
This is executed in a subprocess. We build a new singleton config. then we want to block the config update. No? Otherwise we do not protect a task from updating the config.

jrobinAV:
FYI, The documentation you provided on the description is the 2.0 version. the develop version (2.1.0.dev0) is different.
Also, we have a ticket to change the documentation https://github.com/Avaiga/taipy-doc/issues/280.

src/taipy/core/_scheduler/_dispatcher/_job_dispatcher.py
trgiangdo: Oh yes. I forgot to block config update on subprocess. My bad.

I moved the `Config.block_update()` to the `Core().run()` method so that it will be consistent with the `Core().stop()` method.
However, it might not be necessary.

trgiangdo:
> FYI, The documentation you provided on the description is the 2.0 version. the develop version (2.1.0.dev0) is different.
Also, we have a ticket to change the documentation https://github.com/Avaiga/taipy-doc/issues/280.

Yes, but all the services are still started at the end of the script, which still raise a warning "The Core service is NOT running" anyway.

However, I don't think it will be much of a problem. If we start the Core service before GUI service at the start of the script, the `--help` command will mess up again.

src/taipy/core/taipy.py
```diff
@@ -76,6 +78,8 @@ def submit(
         timeout (Union[float, int]): The optional maximum number of seconds to wait for the jobs to be finished before returning.
 
     """
+    if _SchedulerFactory._dispatcher is None:
+        _warn_no_core_service()
```
jrobinAV: Minor remark: I would put that small piece of code in a decorator so it is easy to re-use it in other entry points if needed.

jrobinAV (dismissed):

jrobinAV (dismissed):

tests/test_taipy.py
```diff
@@ -251,8 +262,10 @@ def test_block_config_when_core_is_running_in_development_mode(self):
         pipeline_cfg_1 = Config.configure_pipeline("p1", task_cfg_1)
         scenario_cfg_1 = Config.configure_scenario("s1", pipeline_cfg_1, Frequency.DAILY)
 
+        Core().run()
+
         scenario_1 = tp.create_scenario(scenario_cfg_1)
-        tp.submit(scenario_1, wait=True)
+        tp.submit(scenario_1)
```
jrobinAV, line 268: Why do you remove the wait=True?

jrobinAV (approved):

tests/test_taipy.py
trgiangdo: because in development mode, `wait=True` doesn't affect if we wait for the job to finish or not (it always wait).

PR #402, opened 2023-01-04
Fix/ Pickle file was not closed when read/write

trgiangdo:


toan-quach (approved):
I thought it will automatically close the file for us 🤣

PR #403, opened 2023-01-05
:green_heart: Address mypy issues

gmarabout:
## What does this PR do?
Addresses mypy errors on "develop" branch, which are blocking development on the related files:
```
src/taipy/core/_repository/_repository_factory.py:26: error: Incompatible return value type (got "Type[NotImplementedError]", expected "Type[_AbstractRepository[Any, Any]]")
src/taipy/core/_repository/_sql_repository.py:135: error: Argument 2 to "next" has incompatible type "None"; expected "_TaipyModel"
src/taipy/core/_manager/_manager_factory.py:26: error: Incompatible return value type (got "Type[NotImplementedError]", expected "Type[_Manager[Any]]")
src/taipy/core/data/data_node.py:290: error: Incompatible return value type (got "Type[NotImplementedError]", expected "str")
src/taipy/core/data/generic.py:99: error: Property "last_edit_date" defined in "DataNode" is read-only
src/taipy/core/_scheduler/_abstract_scheduler.py:37: error: Incompatible return value type (got "Type[NotImplementedError]", expected "List[Job]")
src/taipy/core/_scheduler/_abstract_scheduler.py:50: error: Incompatible return value type (got "Type[NotImplementedError]", expected "Job")
Found 7 errors in 6 files (checked 1 source file)
```

### Changes

* the method `storage_type()` on `DataNode` has now type hint no to raise a typing error despite the fact it returns `NotImplementedError`. Changing the `return` into a `raise` would require bigger refactoring, which is out of scope of this PR.
* Methods that were returning a `NotImplementedError` exception now raise this exception.
* `__get_entities_by_config_and_owner` has been slightly refactored to return only a single `_TaipyModel` so that mypy is no longer confused.
* In `generic.py`, the typing verification has been disabled (using `# type: ignore`) to match similar code (like here https://github.com/gmarabout/taipy-core/blob/5173b8e48fd29c6d0a360c8f2ff39da51f97791a/src/taipy/core/data/excel.py#L136). 

## Jira ticket number?
[404](https://github.com/Avaiga/taipy-core/issues/404)

## Checklist before merging
  
  - [ ] No more Mypy errors in the pre-commit phase. You should force a small modification in any of those files and manually run `pre-commit run`.


src/taipy/core/_repository/_sql_repository.py
```diff
@@ -139,17 +139,15 @@ def _get_by_configs_and_owner_ids(self, configs_and_owner_ids):
...
                 entity = self.__to_entity(entry)
                 key = config, owner
                 res[key] = entity
 
         return res
 
-    def __get_entities_by_config_and_owner(
-        self, config_id: str, owner_id: Optional[str] = "", only_first: bool = False
-    ):
+    def __get_entities_by_config_and_owner(self, config_id: str, owner_id: Optional[str] = "") -> _TaipyModel:
```
gmarabout: *Note to reviewers* This function is called only to retrieve one single entry element. As it is private, it is safe to just always call `query.first()` and then drop the `only_first` flag.

src/taipy/core/data/_data_model.py
```diff
@@ -24,7 +24,7 @@ class _DataNodeModel:
     id: str
     config_id: str
     scope: Scope
-    storage_type: str
+    storage_type: Optional[str]
```
gmarabout: **note to reviewers** The type is changed to `Optional[str]` to reflect the change of all `storage_type()` getters everywhere.

src/taipy/core/_repository/_sql_repository.py
```diff
@@ -129,8 +129,8 @@ def _search(self, attribute: str, value: Any, version_number: Optional[str] = No
         return self.__to_entity(entry)
 
     def _get_by_config_and_owner_id(self, config_id: str, owner_id: Optional[str]) -> Optional[Entity]:
-        entities = iter(self.__get_entities_by_config_and_owner(config_id, owner_id))
```
gmarabout, line 132: **Note to reviewers** I _believe_ this code has the same behavior as before (using an `iter` and then a `next`). Let me know if it is not the case...

src/taipy/core/_repository/_sql_repository.py
```diff
@@ -129,8 +129,8 @@ def _search(self, attribute: str, value: Any, version_number: Optional[str] = No
...
-        return self.__to_entity(next(entities, None))
+        entity = self.__get_entities_by_config_and_owner(config_id, owner_id)
```
gmarabout, line 162: **Note to reviewers** I _believe_ this code has the same behavior as before (using an `iter` and then a `next`). Let me know if it is not the case...

jrobinAV (dismissed):

src/taipy/core/data/_data_model.py
jrobinAV: I believe the storage_type cannot be optional but must be required. 
A data node without a storage type is inconsistent.

src/taipy/core/data/data_node.py
```diff
@@ -286,8 +286,8 @@ def __get_last_modified_datetime(self) -> Optional[datetime]:
 
     @classmethod
     @abstractmethod
-    def storage_type(cls) -> str:
-        raise NotImplementedError()
+    def storage_type(cls) -> Optional[str]:
+        return None
```
jrobinAV: I don't understand this change. 

This is an abstract method that must be overridden by all implementation classes. So if this code is executed, it must raise an error. 

Am I missing something?

gmarabout: I agree. However, this code is called by https://github.com/Avaiga/taipy-core/blob/5173b8e48fd29c6d0a360c8f2ff39da51f97791a/src/taipy/core/data/data_node.py#L521 
Raising an exception from this method would break a bit of everything. The code assumes the storage type can be `None` so returning None here looks like the only option for now...
What do you think?

gmarabout: That also explains why the type for storage type is now `Optional[str]`. 
Or maybe we should ask the storage type only for concrete DataNode subclasses?

src/taipy/core/data/_data_model.py
gmarabout: Ok, let's change the implementation so that this method returns a mandatory `str` or raises an exception.

src/taipy/core/_manager/_manager_factory.py
```diff
@@ -23,7 +23,7 @@ class _ManagerFactory:
     @classmethod
     @abstractmethod
     def _build_manager(cls) -> Type[_Manager]:  # type: ignore
-        return NotImplementedError
+        raise NotImplementedError()
```
trgiangdo:
```suggestion
        raise NotImplementedError
```

src/taipy/core/_repository/_repository_factory.py
```diff
@@ -23,7 +23,7 @@ class _RepositoryFactory:
     @classmethod
     @abstractmethod
     def _build_repository(cls) -> Type[_AbstractRepository]:  # type: ignore
-        return NotImplementedError
+        raise NotImplementedError()
```
trgiangdo:
```suggestion
        raise NotImplementedError
```

src/taipy/core/_scheduler/_abstract_scheduler.py
```diff
@@ -34,7 +34,7 @@ def submit(
         wait: bool = False,
         timeout: Optional[Union[float, int]] = None,
     ) -> List[Job]:
-        return NotImplementedError
+        raise NotImplementedError()
```
trgiangdo:
```suggestion
        raise NotImplementedError
```

src/taipy/core/_scheduler/_abstract_scheduler.py
```diff
@@ -47,9 +47,9 @@ def submit_task(
         wait=False,
         timeout: Optional[Union[float, int]] = None,
     ) -> Job:
-        return NotImplementedError
+        raise NotImplementedError()
```
trgiangdo:
```suggestion
        raise NotImplementedError
```

src/taipy/core/_scheduler/_abstract_scheduler.py
```diff
@@ -47,9 +47,9 @@ def submit_task(
...
 
     @classmethod
     @abstractmethod
     def cancel_job(cls, job):
-        return NotImplementedError
+        raise NotImplementedError()
```
trgiangdo:
```suggestion
        raise NotImplementedError
```

trgiangdo (commented):
I believe we don't need the `()`.

It's not important, but to be consistent with other taipy repositories which `raise NotImplementedError` without the `()` as well.

Also, there are a few more places where we `return NotImplementedError` when it should be `raise NotImplementedError`.
Can you add that to this PR as well @gmarabout?
Thank you in advance 😁

src/taipy/core/data/data_node.py
trgiangdo: Since the `DataNode` class should not be used by the end user, and we want all children of `DataNode` to have `storage_type`, I don't think we should change this.

Do we have any other method to avoid mypy error message here? Like a `# type: ignore` or something.

gmarabout: Yes, I agree @trgiangdo... Changing the implementation of `storage_type()` causes cascading other changes that goes beyond the scope of this PR. 
If you all agree, we shall keep `return NotImplemented` for now in this method, and this would be addressed in another dedicated PR.
Wdyt?

trgiangdo: yes, it should be sufficient for now. Thank you

jrobinAV (dismissed):
Looks good to me.
Regarding the storage_type issue. I created this [ticket](https://github.com/Avaiga/taipy-core/issues/417). We will need to work on it in a near future.

trgiangdo (dismissed):
Thank you @gmarabout. It's ready to merge now

src/taipy/core/_repository/_sql_repository.py
```diff
@@ -169,17 +170,15 @@ def _get_by_configs_and_owner_ids(self, configs_and_owner_ids):
...
                 entity = self.__to_entity(entry)
                 key = config, owner
                 res[key] = entity
 
         return res
 
-    def __get_entities_by_config_and_owner(
-        self, config_id: str, owner_id: Optional[str] = "", only_first: bool = False, version_number: Optional[str] = None
-    ):
+    def __get_entities_by_config_and_owner(self, config_id: str, owner_id: Optional[str] = "") -> _TaipyModel:
```
jrobinAV:
```suggestion
    def __get_entities_by_config_and_owner(self, config_id: str, owner_id: Optional[str] = "", version_number: Optional[str] = None) -> _TaipyModel:
```

jrobinAV (approved):
Great! I am merging it.

PR #406, opened 2023-01-05
bug/ fixed scheduler and job dispatcher using different lock

toan-quach:


jrobinAV (dismissed):

jrobinAV:
I cannot execute tests on local. I got 39 failing tests.
```
self = <src.taipy.core._version._version_sql_repository._VersionSQLRepository object at 0x00000195190C5CC0>
version_number = 'b7d0f133-5046-40ce-a942-455cc42b34d5'

    def _set_development_version(self, version_number):
>       self.repo._set_development_version(version_number)
E       AttributeError: '_SQLRepository' object has no attribute '_set_development_version'

..\src\taipy\core\_version\_version_repository.py:84: AttributeError
```

This is not related to this PR. It should be solved in develop.

tests/conftest.py
```diff
@@ -258,33 +258,15 @@ def tmp_sqlite(tmpdir_factory):
...
-
-def close_sql_database_session_connection():
-    _ScenarioRepositoryFactory._build_repository().repo.session.close_all()
-    _PipelineRepositoryFactory._build_repository().repo.session.close_all()
-    _DataRepositoryFactory._build_repository().repo.session.close_all()
-    _TaskRepositoryFactory._build_repository().repo.session.close_all()
-    _JobRepositoryFactory._build_repository().repo.session.close_all()
-    _CycleRepositoryFactory._build_repository().repo.session.close_all()
-    _VersionRepositoryFactory._build_repository().repo.session.close_all()
+    yield
```
trgiangdo, line 270:
```suggestion
    yield
    
    close_all_sessions()
    
    init_config()
    init_scheduler()
    init_managers()
    init_config()
```
Previously, we have one block of **clean everything** before `yield` and one after `yield`.

Do we need both?

trgiangdo (commented):
I have 1 question.
Otherwise, it's looking good to me.

I think we just need the problem on develop branch to be fixed first to run the test again and merge.

jrobinAV (approved):

PR #420, opened 2023-01-10
SQL Version Repository

joaoandre-avaiga:


jrobinAV:
I run everything locally, and it seems to solve also the issues. That is great.

I know it is still a draft, so please ignore my following comment if it is irrelevant or if it is going to change. 

I don't understand the design, and I don't get the naming convention either.
For instance, I don't get this design: 
`class _BaseSQLRepository(_AbstractRepository[ModelType, Entity])`
`class _TaipyModelTable(_BaseSQLRepository)`
`class _TaipyVersionTable(_BaseSQLRepository)`
`class _SQLRepository(_BaseSQLRepository)`
Let's discuss the design in a dedicated meeting if it s ok for you. 

Again, if you believe it is not necessary now because you have not finished your proposal yet, please do not hesitate to tell me.

joaoandre-avaiga:
> derstand the design, and I don't get the namin
Right now we have the following workflow:

-> Managers can change the repository based on config:

 ```_repository = _VersionRepositoryFactory._build_repository()```

-> The repository factory will return a instanve of the repository based on the config and a repository map:
```
_REPOSITORY_MAP = {"default": _VersionFSRepository, "sql": _VersionSQLRepository}
cls._REPOSITORY_MAP.get(Config.global_config.repository_type, cls._REPOSITORY_MAP.get("default"))() 
```
-> The map refers to a implementation of the Base Repository(FS, SQL, Mongo) for that model.

-> Each implementation has the abstract repository class as a base, like (CycleRepository, VersionRepository, etc)

-> Each model repository calls an adapter to select the concrete implementation of that repository, again, based on the config:

```
self.repo = _RepositoryAdapter.select_base_repository()(**kwargs)
```

-> Similarly to the factory, the RepositoryAdapter also has a map to be able to select the correct repository:

```
_REPOSITORY_MAP = {"default": _FileSystemRepository, "sql": _SQLRepository}
cls._REPOSITORY_MAP.get(Config.global_config.repository_type, cls._REPOSITORY_MAP.get("default"))
```

And here lies the issue. Because of this design, the adapter will always have to choose from the generic implementations of the repository(_FileSystemRepository, _SQLRepository, _MongoRepository). For file system and mongo that have a more flexible structure this works fine, but for SQL, for each model we have to know exactly what the table structure is, to be able to query it. All the models so far(Scenario, Pipeline, etc) are on a single table, therefore they have the same structure and we can reuse the queries to all of them. Now that we have the Version model, which doesn't have the same structure as the other models(and I don't think we should force it to have the same structure), we are not able to reuse the `_SQLRepository` implementation as it is right now. For instance, if I want to load a model the query for a scenario/pipeline/datanode would be:

```
self.session.query(_TaipyModel).filter_by(model_id=model_id).first()
```

While if I want to load a specific version, the query woould be:

```
self.session.query(_TaipyVersion).filter_by(id=id).first()
```

The data models `_TaipyModel` and `_TaipyVersion` are different, because they each represent a table on the database. I've considered try to condense and parametrize each method on `_SQLRepository`, depending on the model, but I've decided against it because it was becomming very confusing. So my proposal at this momment would be:

-> _BaseSQL: Only handles the connection creation, this allows for us to cache the database session, to avoid recreating it for each of the tables.

-> _SQLRepository: Now is just a interface to be exposed to the adapter, it calls the methods to query the tables based on the data model:

```
self._table = (
            _TaipyVersionTable(model, to_model_fct, from_model_fct)
            if model_name == "version"
            else _TaipyModelTable(model, model_name, to_model_fct, from_model_fct)
        )
```

-> _TaipyModelTable and _TaipyVersionTable: The queries are implemented in each class, here we hold the query logic for each table.

I hope this clarifies a bit. I understand that it is confusing, and to be honest I think the design should be revised, at least for SQL data model. We can still have the meeting to discuss if theres any questions. Also feel free to propose another solution.

I know is hard to push for deliveries that don't bring a direct value for the customer, but I would suggest putting some time in the fufure to review our current repository design and see what we can improve.

trgiangdo:
I don't have much to say about the implementation, and I would like to get involve on the discussion as well.
My suggestion for now is to add a `test_version_cli_with_sql_repo.py` to test the behavior on SQL repo.

src/taipy/core/_repository/_sql_model.py
```diff
@@ -42,7 +43,20 @@ class _TaipyVersion(Base):  # type: ignore
 
     id = Column(Text, primary_key=True)
     config = Column(Text)
-
-    def __int__(self, id: str, config: Json):
+    creation_date = Column(Text)
+    is_production = Column(Boolean)
+    is_development = Column(Boolean)
+    is_latest = Column(Boolean)
```
trgiangdo, line 49: This is not consistent with the current FS repo.

Can we have a separate table to store which one is production, which one is latest version, ... ?

src/taipy/core/_repository/_sql_repository.py
```diff
@@ -213,9 +252,238 @@ def _export(self, entity_id: str, folder_path: Union[str, pathlib.Path]):
...
+        self._table = _TaipyVersion
+        super().__init__()
+
+    def load(self, id: str) -> Entity:
+        entry = self.__get_by_id(id)
+        if entry is None:
+            raise ModelNotFound(self.model, id)  # type: ignore
+        return self.__to_entity(entry)
+
+    def _load_all(self, version_number: Optional[str] = None) -> List[Entity]:
```
trgiangdo, line 300: We are not using the `version_number` here to query right?

src/taipy/core/_repository/_sql_repository.py
```diff
@@ -213,9 +252,238 @@ def _export(self, entity_id: str, folder_path: Union[str, pathlib.Path]):
...
+        try:
+            entries = self.session.query(self._table).all()
+            return [self.__to_entity(e) for e in entries]
+        except NoResultFound:
+            return []
+
+    def _load_all_by(self, by, version_number: Optional[str] = None) -> List[Entity]:
```
trgiangdo, line 307: Same here, I think we are not using the `version_number`.

src/taipy/core/_repository/_sql_repository.py
```diff
@@ -213,9 +252,238 @@ def _export(self, entity_id: str, folder_path: Union[str, pathlib.Path]):
...
+
+    def _delete_all(self):
+        self.session.query(self._table).delete()
+        self.session.commit()
+
+    def _delete_many(self, ids: Iterable[str]):
+        self.session.execute(delete(self._table).where(self._table.id.in_(ids)))
+        self.session.commit()
+
+    def _search(self, attribute: str, value: Any, version_number: Optional[str] = None) -> Optional[Entity]:
```
trgiangdo, line 340: Same here, we are not using the `version_number`

trgiangdo (commented):
I have some kinda major comments. However, we can resolve these latter actually, depends on how long it might take.

Another suggestion for now is to add a `test_version_cli_with_sql_repo.py` to test the behavior on SQL repo.
However, we can add the test in a different PR after this one.

src/taipy/core/_repository/_sql_repository.py
jrobinAV: I found the following are not using version_number: 
- `_load_all`
- `_load_all_by`
- `_search`

I guess these will be tested on new tests on the other PR.

jrobinAV: Nevermind, Giang already spotted them.
:/

src/taipy/core/_repository/_sql_model.py
joaoandre-avaiga: Yes, it's not consistent and I don't think we should mimic the FS implementation. For this is the correct data model considering a SQL database. Same for mongo, I would rather that each version document would keep track of its status, be it latest, development or production. Adding another table over complicates things and don't take advantage of database capabilities. I understand your implementation on the FS, but I don't think it should be the model for the others.

src/taipy/core/_repository/_sql_repository.py
joaoandre-avaiga: Yes, but it is a problem of the current repository design. Since we are forced to follow the contract for the repositories and in the AbstractRepository we have the _load_all to take the version number, so every signature needs to have it.

src/taipy/core/_repository/_sql_repository.py
joaoandre-avaiga: Same as I've said before. 

Yes, but it is a problem of the current repository design. Since we are forced to follow the contract for the repositories and in the AbstractRepository we have the _load_all to take the version number, so every signature needs to have it.

.github/workflows/tests.yml
```diff
@@ -13,6 +13,7 @@ jobs:
       matrix:
         python-versions: ['3.8', '3.9', '3.10']
         os: [ubuntu-latest, windows-latest, macos-latest]
+      fail-fast: false
```
trgiangdo, line 16: This is great with the current status of core, I think 😅

trgiangdo (approved):
Imo, it's ready to merge now

jrobinAV (approved):

PR #430, opened 2023-01-16
Feature/#398 cacheable

jrobinAV:
DONE:
- [x] make the `cacheable` attribute of DatanodeConfig deprecated
- [x] remove the `cacheable` attribute from Datanode and make the property deprecated
- [x] Expose the `is_up_to_date` property on Datanode returning True if the data node is up_to_date regarding the validity period (replacing the hidden `_is_in_cache` property)
- [x] create the `skippable` attribute on TaskConfig
- [x] create the `skippable` attribute on Task and populate it from TaskConfig
- [x] Dispatcher now checks if the task is skippable before executing it
- [x] update TaskModel and DatanodeModel accordingly
- [x] update json schema on both taipy studio and taipy core
- [x] update config.pyi in taipy-config repo


jrobinAV:
@FlorianJacta Could you test the compatibility of a previous config with cacheable activated using this Branch? I would like to test it deeply before merging it.

tests/core/config/test_data_node_config.py
```diff
@@ -142,8 +142,8 @@ def test_block_datanode_config_update_in_development_mode():
     assert Config.data_nodes[data_node_id].default_path == "foo.p"
     assert Config.data_nodes[data_node_id].storage_type == "pickle"
     assert Config.data_nodes[data_node_id].scope == Scope.SCENARIO
-    assert Config.data_nodes[data_node_id].cacheable == False
-    assert Config.data_nodes[data_node_id].properties == {"default_path": "foo.p"}
+    assert Config.data_nodes[data_node_id].cacheable is False
+    assert Config.data_nodes[data_node_id].properties == {"cacheable": False, "default_path": "foo.p"}
```
toan-quach: I believe this should be skippable instead of cacheable right?

toan-quach (dismissed):
lgtm, just some minor remarks

tests/core/config/test_data_node_config.py
jrobinAV: Actually, I wanted to test that a user can still use "cacheable" as a custom property in data node configs.

toan-quach: ah i see, I'm all good with that :D though I'd suggest a different test for it, but it's nothing major, just personal preference

jrobinAV: It should be on a separate test. You are right.

toan-quach (dismissed):
lgtm! 😄

toan-quach (approved):

PR #431, opened 2023-01-16
fix default data node storage type

Dr-Irv:
There is an issue that if you configure the default data node to have a storage type other than `"pickle"`, using `configure_default_data_node()`, that value is not respected.  This PR fixes that issue.  Added a test to test it.

Note - it seems that black did some extra reformatting of some of the code.

### DIFF
--- a/src/taipy/core/config/data_node_config.py
+++ b/src/taipy/core/config/data_node_config.py
@@ -301,7 +301,7 @@ class DataNodeConfig(Section):
     @staticmethod
     def _configure(
         id: str,
-        storage_type: str = _DEFAULT_STORAGE_TYPE,
+        storage_type: Optional[str] = None,
         scope: Scope = _DEFAULT_SCOPE,
         cacheable: bool = False,
         **properties,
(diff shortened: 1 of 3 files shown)


src/taipy/core/config/data_node_config.py
```diff
@@ -294,14 +335,16 @@ def _configure_default(storage_type: str, scope: Scope = _DEFAULT_SCOPE, cacheab
...
+            _Config.DEFAULT_KEY, storage_type, scope, cacheable, **properties
+        )
         Config._register(section)
         return Config.sections[DataNodeConfig.name][_Config.DEFAULT_KEY]
 
     @staticmethod
     def _configure(
         id: str,
-        storage_type: str = _DEFAULT_STORAGE_TYPE,
+        storage_type: Optional[str] = None,
```
toan-quach, line 347: Just out of curiosity, but why is it None for storage type here? Since we're only providing limited options for the dn, would it be better to have a default value like pickle instead?

toan-quach, line 304: Just out of curiosity, but why is it None for storage type here? Since we're only providing limited options for the dn, would it be better to have a default value like pickle instead?

toan-quach (dismissed):
Hi! The PR looks great to me! Just some part I'm not sure I understand 😄 Thanks for your contribution in making taipy better!!!

src/taipy/core/config/data_node_config.py
Dr-Irv: > 

The issue here is that if you make the default `"pickle"` in the `_configure()` method, then it ignores the default set in `configure_default_data_node()`.   Using `None` makes it look at whatever was set in `configure_default_data_node()`

I updated the tests to test both the default without calling `configure_default_data_node()` and with calling it.

jrobinAV (dismissed):
Hi,
Thank you for your contribution and for the fix. 

Just a minor remark on the code formatting. 
In our formatting convention, the parameters of a method definition should remain on a single line if the line size is less than 120 characters. 
It s not blocking, but could you revert these formatting changes?
By the way, we are also using black. Did you use taipy black configuration or your own?

Dr-Irv:
> Hi, Thank you for your contribution and for the fix.
> 
> Just a minor remark on the code formatting. In our formatting convention, the parameters of a method definition should remain on a single line if the line size is less than 120 characters. It s not blocking, but could you revert these formatting changes? By the way, we are also using black. Did you use taipy black configuration or your own?

I have things set up in VSCode to use automatically use `black` when saving a file.  It picks up any configuration from a `pyproject.toml` file, if that file exists.  In my most recent commit, I have included that file, which might be useful for others as well.  That commit also takes out the formatting changes, but I also clarified the docs as well.

You only had `black` configured via a pre-commit, which is why it wasn't picking up your `line-length` setting.

jrobinAV (approved):
Thanks you!

PR #446, opened 2023-01-27
chore: add version information to init file

joaoandre-avaiga:


joaoandre-avaiga:
@jrobinAV When you say to put the version information on the root __init__ file did you mean the one inside src/taipy/core right? Just to make sure

joaoandre-avaiga:
I've tried to put the code to load the version in a module on: `src.taipy.core.version`, but import this on setup.py causes an error because taipy.config is not yet installed, so is not possible to import anything from src.taipy.core module in setup.py. Let me know if you have any suggestions where we could isolate this logic in a function to be reused everywhere on the project.

jrobinAV (approved):
I don't see how to factorize the function. :/

PR #460, opened 2023-02-02
[Hotfix 2.1]ref man

jrobinAV:


src/taipy/core/__init__.py
```diff
@@ -11,16 +11,17 @@
...
 
-To build a Taipy Core application, the first step consists of setting up the Taipy configuration to design the
-characteristics and the behavior of your application.
+To build a Taipy Core application, the first step consists of setting up the Taipy configuration to design
+your application's characteristics and behaviors.
 Import `Config^` from the `taipy.config^` module, then use the various methods of the `Config^` singleton class to
-configure your core application. In particular configure the [data nodes](../../../manuals/core/config/data-node-config,
-[tasks](../../../manuals/core/config/task-config), and [scenarios](../../../manuals/core/config/scenario-config).
+configure your core application. In particular, configure the
+[data nodes](../../../manuals/core/config/data-node-config), [tasks](../../../manuals/core/config/task-config),
```
toan-quach, line 23: what about the pipeline? :D

toan-quach (approved):
lgtm, just minor remark

src/taipy/core/__init__.py
jrobinAV: I am starting to progressively make the pipeline disappear from the main concepts in the documentation. 

This is a big change we are thinking about with Vincent.
We would like not to have to configure the pipeline directly at the config level anymore. But just an attribute of a scenario config. The purpose in the long term is to stop considering the pipeline as an entity but just as an attribute of a scenario. And the scenario would be : 
- a bunch of tasks that are connected or not through data nodes.
- a list of pipelines, where each pipeline is a set of connected tasks that must belong to the scenario.

With this, a pipeline only represents a set of tasks that can be submitted together. That's it.

Does it make sense?

Something like:
Instead of having the following
```
#Let's imagine we have defined : task_1_cfg, task_2_cfg, tack_3_cfg, tack_4_cfg
pipeline_1_cfg = Config.configure_pipeline("training", [task_1_cfg, task_2_cfg])
pipeline_2_cfg = Config.configure_pipeline("scoring", [task_2_cfg, task_3_cfg])

scenario_cfg = Config.configure_scenario(
        id="my_sc_id", 
        pipelines=[pipeline_1_cfg, pipeline_2_cfg, ], 
        comparators = {}
        frequency = MONTHLY
)
```
We would like to have this:
```
#Let's imagine we have defined : task_1_cfg, task_2_cfg, tack_3_cfg, tack_4_cfg
scenario_cfg = Config.configure_scenario(
        id="my_sc_id", 
        tasks=[task_1_cfg, task_2_cfg, tack_3_cfg, tack_4_cfg], 
        comparators = {}
        frequency = MONTHLY
)
scenario_cfg.add_pipeline("training", task_1_cfg, task_2_cfg)
scenario_cfg.add_pipeline("scoring", task_2_cfg, task_3_cfg)
```

In this example. scenario_cfg is made of 4 tasks.

jrobinAV: Does it make sense ?

toan-quach: I agree with the idea of removing the pipeline entity and config and combining it with the scenario config. But I think something like this would be simpler
```
scenario_cfg = Config.configure_scenario(
        id="my_sc_id", 
        comparators = {}
        frequency = MONTHLY
)
scenario_cfg.add_pipeline("training", tasks=[task_1_cfg, task_2_cfg])
scenario_cfg.add_pipeline("scoring", tasks=[task_1_cfg, task_2_cfg])
```
If we ask the user to provide the task configs in scenario configs, just to have to provide them again in add_pipeline is confusing, and the task configs not store in a list might introduce some confusion since we're asking the users to assume that whichever the task that they provide first, will be the 1st task, using a list can embed this meaning in a way that's much cleaner to work with and clearer to understand

jrobinAV: I agree.

In any case, this is for now just a general idea. Nothing is designed yet.  I guess a good API design should be the first and most important thing to do.

Thx,

Issue #421, opened 2023-01-12
Refactor the behavior of Config.config_data_node() method


trgiangdo:
**Description**

Currently, the generic `DataNodeConfig.config_data_node()` method creates a new section ignoring the default properties of specific storage types.

For example
```python
Config.config_data_node(storage_type="csv")
```
will ignore the `has_header` property of CSVDataNode otherwise `Config.config_csv_data_node()` would have in the `Config._applied_config`.

PR #463, opened 2023-02-03
Refactor/#421 Refactor the behavior of Config.config_data_node() method

trgiangdo:
https://github.com/Avaiga/taipy-core/issues/421

To summarize, the `DataNode._configure()` is now a classmethod to call to other configure.

### DIFF
--- a/src/taipy/core/config/data_node_config.py
+++ b/src/taipy/core/config/data_node_config.py
@@ -38,7 +38,7 @@ class DataNodeConfig(Section):
             "in_memory".
             The default value is "pickle".
             Note that the "in_memory" value can only be used when `JobConfig^`.mode is "standalone".
-        scope (Optional[Scope^]):  The `Optional[Scope^]` of the data nodes instantiated from the data node config.
+        scope (Optional[Scope^]):  The optional `Scope^` of the data nodes instantiated from the data node config.
             The default value is SCENARIO.
         **properties (dict[str, Any]): A dictionary of additional properties.
     """
--- a/src/taipy/core/config/data_node_config.py
+++ b/src/taipy/core/config/data_node_config.py
@@ -699,8 +699,8 @@ class DataNodeConfig(Section):
             db_extra_args (Optional[Dict[str, Any]]): A dictionary of additional arguments to be passed into database
                 connection string.
             read_query (Optional[str]): The SQL query string used to read the data from the database.
-            write_query_builder (Optional[Callable]): A callback function that takes the data as an input parameter and returns a
-                list of SQL queries.
+            write_query_builder (Optional[Callable]): A callback function that takes the data as an input parameter
+                and returns a list of SQL queries.
             exposed_type: The exposed type of the data read from SQL query. The default value is `pandas`.
             scope (Scope^): The scope of the SQL data node configuration. The default value is
                 `Scope.SCENARIO`.
--- a/src/taipy/core/config/data_node_config.py
+++ b/src/taipy/core/config/data_node_config.py
@@ -38,7 +38,7 @@ class DataNodeConfig(Section):
             "in_memory".
             The default value is "pickle".
             Note that the "in_memory" value can only be used when `JobConfig^`.mode is "standalone".
-        scope (Scope^):  The `Scope^` of the data nodes instantiated from the data node config. The default value is
+        scope (Optional[Scope^]):  The `Optional[Scope^]` of the data nodes instantiated from the data node config. The default value is
             SCENARIO.
         **properties (dict[str, Any]): A dictionary of additional properties.
     """
@@ -96,16 +96,18 @@ class DataNodeConfig(Section):
     # In memory
     _OPTIONAL_DEFAULT_DATA_IN_MEMORY_PROPERTY = "default_data"
     # SQL
-    _OPTIONAL_EXPOSED_TYPE_SQL_PROPERTY = "exposed_type"
     _REQUIRED_DB_USERNAME_SQL_PROPERTY = "db_username"
     _REQUIRED_DB_PASSWORD_SQL_PROPERTY = "db_password"
     _REQUIRED_DB_NAME_SQL_PROPERTY = "db_name"
     _REQUIRED_DB_ENGINE_SQL_PROPERTY = "db_engine"
     _REQUIRED_DB_ENGINE_SQLITE = "sqlite"
-    _REQUIRED_DB_ENGINE_MSSQL = "mssql"
+    _OPTIONAL_PORT_SQL_PROPERTY = "db_port"
+    _OPTIONAL_HOST_SQL_PROPERTY = "db_host"
+    _OPTIONAL_DRIVER_SQL_PROPERTY = "db_driver"
     _OPTIONAL_DB_EXTRA_ARGS_SQL_PROPERTY = "db_extra_args"
+    _OPTIONAL_EXPOSED_TYPE_SQL_PROPERTY = "exposed_type"
     # SQL_TABLE
-    _REQUIRED_TABLE_NAME_PROPERTY = "table_name"
+    _REQUIRED_TABLE_NAME_SQL_TABLE_PROPERTY = "table_name"
     # SQL
     _REQUIRED_READ_QUERY_SQL_PROPERTY = "read_query"
     _REQUIRED_WRITE_QUERY_BUILDER_SQL_PROPERTY = "write_query_builder"
@@ -113,21 +115,25 @@ class DataNodeConfig(Section):
     _REQUIRED_DB_NAME_MONGO_PROPERTY = "db_name"
     _REQUIRED_COLLECTION_NAME_MONGO_PROPERTY = "collection_name"
     _OPTIONAL_CUSTOM_DOCUMENT_MONGO_PROPERTY = "custom_document"
-    _OPTIONAL_DB_USERNAME_MONGO_PROPERTY = "db_username"
-    _OPTIONAL_DB_PASSWORD_MONGO_PROPERTY = "db_password"
+    _OPTIONAL_USERNAME_MONGO_PROPERTY = "db_username"
+    _OPTIONAL_PASSWORD_MONGO_PROPERTY = "db_password"
+    _OPTIONAL_HOST_MONGO_PROPERTY = "db_host"
+    _OPTIONAL_PORT_MONGO_PROPERTY = "db_port"
     _OPTIONAL_DB_EXTRA_ARGS_MONGO_PROPERTY = "db_extra_args"
     # Pickle
     _OPTIONAL_DEFAULT_PATH_PICKLE_PROPERTY = "default_path"
     _OPTIONAL_DEFAULT_DATA_PICKLE_PROPERTY = "default_data"
     # JSON
     _OPTIONAL_ENCODER_JSON_PROPERTY = "encoder"
-    _OPTIONAL_DECODER_TYPE_JSON_PROPERTY = "decoder"
-    _REQUIRED_DEFAULT_PATH_JSON_PROPERTY = "default_path"
+    _OPTIONAL_DECODER_JSON_PROPERTY = "decoder"
+    _OPTIONAL_DEFAULT_PATH_JSON_PROPERTY = "default_path"
     # Parquet
     _OPTIONAL_EXPOSED_TYPE_PARQUET_PROPERTY = "exposed_type"
     _OPTIONAL_DEFAULT_PATH_PARQUET_PROPERTY = "default_path"
-    _OPTIONAL_COLUMNS_PARQUET_PROPERTY = "columns"
+    _OPTIONAL_ENGINE_PARQUET_PROPERTY = "engine"
     _OPTIONAL_COMPRESSION_PARQUET_PROPERTY = "compression"
+    _OPTIONAL_READ_KWARGS_PARQUET_PROPERTY = "read_kwargs"
+    _OPTIONAL_WRITE_KWARGS_PARQUET_PROPERTY = "write_kwargs"
 
     _REQUIRED_PROPERTIES: Dict[str, List] = {
         _STORAGE_TYPE_VALUE_PICKLE: [],
@@ -136,7 +142,7 @@ class DataNodeConfig(Section):
             _REQUIRED_DB_PASSWORD_SQL_PROPERTY,
             _REQUIRED_DB_NAME_SQL_PROPERTY,
             _REQUIRED_DB_ENGINE_SQL_PROPERTY,
-            _REQUIRED_TABLE_NAME_PROPERTY,
+            _REQUIRED_TABLE_NAME_SQL_TABLE_PROPERTY,
         ],
         _STORAGE_TYPE_VALUE_SQL: [
             _REQUIRED_DB_USERNAME_SQL_PROPERTY,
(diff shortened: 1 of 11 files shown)


tests/core/config/test_data_node_config.py
```diff
@@ -46,14 +46,6 @@ def test_data_node_config_check():
...
         Config.check()
 
-    with pytest.raises(ConfigurationIssueError):
-        Config.configure_data_node("data_nodes", storage_type="sql")
-        Config.check()
-
-    with pytest.raises(ConfigurationIssueError):
-        Config.configure_data_node("data_nodes", storage_type="generic")
-        Config.check()
-
```
trgiangdo, line 56: Note: This 2 tests are removed because these methods will now return TypeError due to missing required properties

toan-quach (dismissed):
lgtm

tests/core/config/test_data_node_config.py
jrobinAV: Can't you just change the error raised?
```
with pytest.raises(TypeError):
        Config.configure_data_node("data_nodes", storage_type="sql")
```

tests/core/config/test_data_node_config.py
```diff
@@ -141,7 +133,7 @@ def test_block_datanode_config_update_in_development_mode():
     assert Config.data_nodes[data_node_id].default_path == "foo.p"
     assert Config.data_nodes[data_node_id].storage_type == "pickle"
     assert Config.data_nodes[data_node_id].scope == Scope.SCENARIO
-    assert Config.data_nodes[data_node_id].properties == {"default_path": "foo.p"}
+    assert Config.data_nodes[data_node_id].properties == {"default_data": None, "default_path": "foo.p"}
```
jrobinAV: First, I don't get why we have this new behavior. Where does the `"default_data": None` comes from?

Second, I believe this is a bug. I am guessing in the python_code_config, default_data has been set to None. We don't want to store a `None` property in the config if the user does not provide it explicitly. At least for two reasons:
1. If there is a default_data provided in the default_data_node_config, it will be overridden by None.
2. If we store None, how to differentiate this case from the case where the user does provide None as default_data

Then, I was surprised to discover this bug in these tests that are not maid for that. 
We should have created dedicated tests that guarantee that we don't store in properties some None attributes that are not explicitly passed in the class constructor.

tests/core/data/test_data_manager.py
```diff
@@ -70,7 +70,8 @@ def test_create_and_get_csv_data_node(self):
         assert _DataManager._get(csv_dn.id).job_ids == csv_dn.job_ids
         assert not _DataManager._get(csv_dn.id).is_ready_for_reading
         assert _DataManager._get(csv_dn.id).is_ready_for_reading == csv_dn.is_ready_for_reading
-        assert len(_DataManager._get(csv_dn.id).properties) == 3
+        assert len(_DataManager._get(csv_dn.id).properties) == 4
+        assert _DataManager._get(csv_dn.id).properties.get("default_path") is None
```
jrobinAV: That's the third reason we want to avoid the new behavior. We don't want to pass None attributes to entities at their creation. It is pointless.

jrobinAV (changes requested):
I believe we introduced a bug although I don't know why.

tests/core/config/test_data_node_config.py
trgiangdo: Yes this is because the configure_pickle_data_node has default_data is None.

I'll try to figure out a way to avoid this (and other None default values as well).

trgiangdo: Hi @jrobinAV, the behavior is actually from calling `config_pickle_datanode()`, which mean that None is pass to the default_data as well.

This means all attributes with `None` as default should be ignore when building Config.

tests/core/data/test_data_manager.py
trgiangdo: Resolve by the latest commit

tests/core/config/test_data_node_config.py
trgiangdo: Resolve by the latest commit

src/taipy/core/config/data_node_config.py
```diff
@@ -157,49 +163,72 @@ class DataNodeConfig(Section):
             _REQUIRED_READ_FUNCTION_GENERIC_PROPERTY,
             _REQUIRED_WRITE_FUNCTION_GENERIC_PROPERTY,
         ],
-        _STORAGE_TYPE_VALUE_JSON: [_REQUIRED_DEFAULT_PATH_JSON_PROPERTY],
+        _STORAGE_TYPE_VALUE_JSON: [],
         _STORAGE_TYPE_VALUE_PARQUET: [],
     }
 
     _OPTIONAL_PROPERTIES = {
```
trgiangdo, line 170: The `_OPTIONAL_PROPERTIES` was not used anywhere before.

I updated it to a dictionary, mapping optional property keys to the default value (for some attributes, most are None).

src/taipy/core/config/data_node_config.py
```diff
@@ -267,15 +296,26 @@ def _update(self, as_dict, default_section=None):
         self._storage_type = as_dict.pop(self._STORAGE_TYPE_KEY, self._storage_type)
         if self._storage_type is None and default_section:
             self._storage_type = default_section.storage_type
+
         self._scope = as_dict.pop(self._SCOPE_KEY, self._scope)
         if self._scope is None and default_section:
-            self._scope = default_section.scope
+            if default_section.scope:
```
trgiangdo: For now, the Scope is always assigned as the `default_section.scope` if defined (even with different storage_type).

If we only want the Scope to be `default_section.scope` when the storage_type is the same, add ` and self._storage_type == default_section.storage_type:` condition here.

src/taipy/core/config/data_node_config.py
```diff
@@ -267,15 +296,26 @@ def _update(self, as_dict, default_section=None):
...
+                self._scope = default_section.scope
+            else:
+                self._scope = self._DEFAULT_SCOPE
+
         self._properties.update(as_dict)
-        if default_section:
+        if default_section and self._storage_type == default_section.storage_type:
             self._properties = {**default_section.properties, **self._properties}
 
+        # Assign default value to optional properties if not defined by user
```
trgiangdo, line 311: This section is added because when configure a datanode, we ignore all optional fields that have `None` value.

Then when compile the Config, we check if that optional property is defined by user or not to assign to the data node.
If not, then use the pre-defined value on `_OPTIONAL_PROPERTIES`.

src/taipy/core/config/data_node_config.py
```diff
@@ -284,22 +324,23 @@ def _configure_default(storage_type: str, scope: Scope = _DEFAULT_SCOPE, **prope
...
+            scope (Optional[Scope^]): The default scope for all data node configurations.
                 The default value is `Scope.SCENARIO`.
             **properties (Dict[str, Any]): A keyworded variable length list of additional
                 arguments.
         Returns:
             `DataNodeConfig^`: The default data node configuration.
         """
         section = DataNodeConfig(_Config.DEFAULT_KEY, storage_type, scope, **properties)
-        Config._register(section)
+        Config._register_default(section)
```
trgiangdo, line 335: This is to differentiate with the generic configure methods, and to replace the "default" config instead of update like the generic one.

src/taipy/core/config/data_node_config.py
```diff
@@ -310,222 +351,232 @@ def _configure(
...
                 `(Config.)configure_default_data_node()^`).
             **properties (Dict[str, Any]): A keyworded variable length list of additional
                 arguments.
         Returns:
             `DataNodeConfig^`: The new data node configuration.
         """
-        section = DataNodeConfig(id, storage_type, scope, **properties)
-        Config._register(section)
-        return Config.sections[DataNodeConfig.name][id]
+        configuration_map: Dict[str, Callable] = {
```
trgiangdo, line 362: This dictionary map the `"storage_type"` string to the corresponding configure method.

src/taipy/core/config/data_node_config.py
```diff
@@ -310,222 +351,232 @@ def _configure(
...
+            cls._STORAGE_TYPE_VALUE_PARQUET: cls._configure_parquet,
+        }
+
+        if storage_type in cls._ALL_STORAGE_TYPES:
+            return configuration_map[storage_type](id=id, scope=scope, **properties)
+
+        return cls.__configure(id, storage_type, scope, **properties)
 
-    @staticmethod
+    @classmethod
```
trgiangdo, line 380: This is updated to `@classmethod` since I need to access to the `cls._OPTIONAL_...` everywhere.

src/taipy/core/config/data_node_config.py
```diff
@@ -310,222 +351,232 @@ def _configure(
...
-            **properties,
-        )
-        Config._register(section)
-        return Config.sections[DataNodeConfig.name][id]
+        if default_path is not None:
+            properties[cls._OPTIONAL_DEFAULT_PATH_CSV_PROPERTY] = default_path
+        if has_header is not None:
+            properties[cls._OPTIONAL_HAS_HEADER_CSV_PROPERTY] = has_header
+        if exposed_type is not None:
+            properties[cls._OPTIONAL_EXPOSED_TYPE_CSV_PROPERTY] = exposed_type
```
trgiangdo, line 410: This section is to ignore all `None` properties.

However, the code is really ugly atm. If anyone know how to refactor this, please let me know to make it easier to maintain.

trgiangdo (commented):
Their are a lot of changes in `config.data_node_config.py`. I left some comments to better explain the changes.

src/taipy/core/config/data_node_config.py
```diff
@@ -38,8 +38,8 @@ class DataNodeConfig(Section):
             "in_memory".
             The default value is "pickle".
             Note that the "in_memory" value can only be used when `JobConfig^`.mode is "standalone".
-        scope (Scope^):  The `Scope^` of the data nodes instantiated from the data node config. The default value is
-            SCENARIO.
+        scope (Optional[Scope^]):  The `Optional[Scope^]` of the data nodes instantiated from the data node config.
```
```suggestion
        scope (Optional[Scope^]):  The optional `Scope^` of the data nodes instantiated from the data node config.
```

src/taipy/core/config/data_node_config.py
```diff
@@ -267,15 +296,26 @@ def _update(self, as_dict, default_section=None):
...
         self._properties.update(as_dict)
-        if default_section:
+        if default_section and self._storage_type == default_section.storage_type:
             self._properties = {**default_section.properties, **self._properties}
 
+        # Assign default value to optional properties if not defined by user
+        if self._OPTIONAL_PROPERTIES.get(self._storage_type):
+            for optional_property, default_value in self._OPTIONAL_PROPERTIES[self._storage_type].items():
+                if default_value is not None and self._properties.get(optional_property) is None:
+                    self._properties[optional_property] = default_value
```
jrobinAV, line 315: I love it!

src/taipy/core/config/data_node_config.py
jrobinAV: I believe that s fine!

jrobinAV (dismissed):
That was complex, but we (you) made it. 
That looks great. Thx!

src/taipy/core/config/data_node_config.py
trgiangdo: Find/Replace error 😅 sorry

tests/core/version/test_version_cli.py
```diff
@@ -588,7 +579,7 @@ def twice(a):
 
 
 def config_scenario():
-    Config.configure_data_node(id="d0", storage_type="csv")
+    Config.configure_data_node(id="d0", default_path="temp.csv")
```
trgiangdo:
```suggestion
    Config.configure_data_node(id="d0")
```

src/taipy/core/config/data_node_config.py
```diff
@@ -38,8 +38,8 @@ class DataNodeConfig(Section):
             "in_memory".
             The default value is "pickle".
             Note that the "in_memory" value can only be used when `JobConfig^`.mode is "standalone".
-        scope (Scope^):  The `Scope^` of the data nodes instantiated from the data node config. The default value is
-            SCENARIO.
+        scope (Optional[Scope^]):  The optional `Scope^` of the data nodes instantiated from the data node config.
```
toan-quach: The OCD in me feels weird looking at it haha. It's very very very minor, I'd ignore it if I were you 🤣 
```suggestion
        scope (Optional[Scope^]): The optional `Scope^` of the data nodes instantiated from the data node config.
```

src/taipy/core/config/data_node_config.py
```diff
@@ -157,49 +163,72 @@ class DataNodeConfig(Section):
...
-            _OPTIONAL_COMPRESSION_PARQUET_PROPERTY,
-        ],
+        _STORAGE_TYPE_VALUE_GENERIC: {
+            _OPTIONAL_READ_FUNCTION_PARAMS_GENERIC_PROPERTY: None,
+            _OPTIONAL_WRITE_FUNCTION_PARAMS_GENERIC_PROPERTY: None,
+        },
+        _STORAGE_TYPE_VALUE_CSV: {
+            _OPTIONAL_DEFAULT_PATH_CSV_PROPERTY: None,
+            _OPTIONAL_HAS_HEADER_CSV_PROPERTY: True,
+            _OPTIONAL_EXPOSED_TYPE_CSV_PROPERTY: _DEFAULT_EXPOSED_TYPE,
```
toan-quach, line 178: should we have a specific _DEFAULT_EXPOSED_TYPE for each datanode, I think we can use a share default exposed type value for now, but when we want to expand many more data node types it will be handy to have a specific one for each

src/taipy/core/config/data_node_config.py
```diff
@@ -267,15 +296,26 @@ def _update(self, as_dict, default_section=None):
...
         if self._storage_type is None and default_section:
             self._storage_type = default_section.storage_type
+
         self._scope = as_dict.pop(self._SCOPE_KEY, self._scope)
         if self._scope is None and default_section:
-            self._scope = default_section.scope
+            if default_section.scope and self._storage_type == default_section.storage_type:
+                self._scope = default_section.scope
+            else:
+                self._scope = self._DEFAULT_SCOPE
```
toan-quach, line 305: this is great! but I strongly believes we should document it carefully as it can very easily become an "unexpected behavior" on the user side without detailed explanation in the doc

src/taipy/core/config/data_node_config.py
toan-quach: I wouldn't recommend putting a dictionary as a default value here, this link explains the potential problem in great detail https://stackoverflow.com/questions/26320899/why-is-the-empty-dictionary-a-dangerous-default-value-in-python. Of course this only applies if we made modification to the `configuration_map` and we're not doing so in the function so it's probably A-okay, I'll leave the judgement call to you 😃

src/taipy/core/config/data_node_config.py
toan-quach, line 410: I believe it's fine, but if you need something, here is something I came up with
```suggestion
        def replace_if_not_None(value_pairs, properties): # can be moved out of this function be used else where
                for key, value in value_pairs:
                        if value is not None:
                                properties[key] = value
                 return properties
         value_pairs = [
                 (cls._OPTIONAL_DEFAULT_PATH_CSV_PROPERTY, default_path),
                 (cls._OPTIONAL_HAS_HEADER_CSV_PROPERTY, has_header),
                 (cls._OPTIONAL_EXPOSED_TYPE_CSV_PROPERTY, exposed_type)
        ] # can also be a dictionary to look cleaner (maybe?)
        properties = replace_if_not_None(value_pairs, properties)
```

toan-quach (dismissed):
Job well done!! Congrats on finishing this heavy task!

src/taipy/core/config/data_node_config.py
trgiangdo: It is dangerous if we set the default value of a function parameter right? `configuration_map` here is a local variable though.

Or do you mean some default values in `_OPTIONAL_PROPERTIES`? If it is then I just commit a fix for that 😅 Thank you

src/taipy/core/config/data_node_config.py
trgiangdo: This seems great.

On second though, what I actually want here is to be able to use `_OPTIONAL_PROPERTIES` and `_REQUIRED_PROPERTIES` to automatically assign properties to the `properties`.

Using `replace_if_not_None()` here will be cleaner, but it will be more complex to understand the code, and we still need to list all properties manually for future DataNode anyway.

So, 1 vote for keeping it the way it is right now.
What do you think @toan-quach @jrobinAV ?

src/taipy/core/config/data_node_config.py
trgiangdo: How can I miss this 😓 Need to fix this asap, or else taipy-core will break

src/taipy/core/config/data_node_config.py
toan-quach: ohhh I didn't notice that, you're right!! Nevermind it then :D

src/taipy/core/config/data_node_config.py
toan-quach: I don't have a strong opinion on this so I'll leave the decision making to you :D

jrobinAV: I don't have a strong opinion either... A slight preference for what we have today.

src/taipy/core/config/data_node_config.py
jrobinAV: Yes. I agree.

src/taipy/core/config/data_node_config.py
jrobinAV: That's an interesting question, Indeed.

It depends on what is the future of the `exposed_type` attribute name. Today, it represents the return type of the tabular data nodes' read method. We don't use the `exposed_type` for Json or mongo data nodes. We use default or custom documents.
If we use `exposed_type` only for tabular data nodes, the default exposed type could be the same for all. Or not...
If we extend the attribute name `exposed_type` to other types of data (document-based, graph-based, literal, ...), we most probably want to have a specific default value per storage type.

Let s keep this in mind.

jrobinAV (approved):

Issue #464, opened 2023-02-03
BUG-Error when loading a config with no storage type


FlorianJacta:
**Description**

An error is raised when trying to load a Config TOML file that 'should be working'. It comes from the fact that no storage type has been written in the Config file for Data Nodes. No issue is raised when `storage_type = "pickle"` are added to the Config for Data Nodes.

This is the error:
```
Traceback (most recent call last):
  File "C:\Users\jacta\AppData\Local\R-MINI~1\envs\tp_2_1\Lib\site-packages\taipy\core\data\_data_manager.py", line 78, in __create
    return cls.__DATA_NODE_CLASS_MAP[data_node_config.storage_type](  # type: ignore
           ~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
KeyError: None

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "c:\Users\jacta\Desktop\Stage\Avaiga\Code\taipy-getting-started-core-git\taipy-getting-started-core\src\step_01_toml.py", line 20, in <module>
    scenario = tp.create_scenario(scenario_cfg)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\jacta\AppData\Local\R-MINI~1\envs\tp_2_1\Lib\site-packages\taipy\core\taipy.py", line 437, in create_scenario
    return _ScenarioManagerFactory._build_manager()._create(config, creation_date, name)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\jacta\AppData\Local\R-MINI~1\envs\tp_2_1\Lib\site-packages\taipy\core\scenario\_scenario_manager.py", line 103, in _create
    pipelines = [
                ^
  File "C:\Users\jacta\AppData\Local\R-MINI~1\envs\tp_2_1\Lib\site-packages\taipy\core\scenario\_scenario_manager.py", line 104, in <listcomp>
    _PipelineManagerFactory._build_manager()._get_or_create(p_config, cycle.id if cycle else None, scenario_id)       
  File "C:\Users\jacta\AppData\Local\R-MINI~1\envs\tp_2_1\Lib\site-packages\taipy\core\pipeline\_pipeline_manager.py", line 86, in _get_or_create
[...]

PR #466, opened 2023-02-07
bug/464 fix error mapping storage type is None to DATA_NODE_CLASS_MAP

toan-quach:
#464

### DIFF
--- a/tests/core/data/test_data_manager.py
+++ b/tests/core/data/test_data_manager.py
@@ -453,8 +453,7 @@ class TestDataManager:
         _DataManager._delete_all()
         assert not file_exists(generated_pickle_dn_3.path)
 
-    def test_create_dn_when_config_storage_type_none(self):
-
+    def test_create_dn_from_loaded_config_no_storage_type(self):
         file_config = NamedTemporaryFile(
             """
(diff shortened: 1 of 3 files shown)


jrobinAV:
It looks good to me. 
We should add a test that was failing before your fix and that is success after.

tests/core/data/test_data_manager.py
```diff
@@ -451,3 +452,57 @@ def test_delete_does_clean_generated_pickle_files(self, pickle_file_path):
...
+        Config.load(file_config.filename)
+
+        from src.taipy import core as tp
+
+        scenario = tp.create_scenario(Config.scenarios["my_scenario"])
+
+        assert isinstance(scenario.input, PickleDataNode)
+        assert isinstance(scenario.output, PickleDataNode)
+
+        # Config._register_default(DataNodeConfig('default_id', 'csv'))
```
toan-quach: @jrobinAV 
I want to add a test where we change the default storage type and the data manager should change the default class used for dn as well. But _register_default doesn't seem to do the work. How can I change the default storage type for dn in the config?

jrobinAV (approved):

Issue #443, opened 2023-01-23
Cannot use "parent" as a property in a Scenario property


Dr-Irv:
**Description**

Currently if you create a scenario and if you do `scenario.properties["parent"] = something`, you get an error, because `"parent"` is checked as a property key in `taipy\core\common\_properties.py` .  I think that either that property key should be using an underscore in front and/or the documentation should say that `"parent"` cannot be a property key.

jrobinAV:
Hello Irv.

Indeed, we have a few predefined properties that are reserved and not documented. 
it does not only concern scenarios but all entities: Cycles, Scenarios, Pipelines, Tasks, DataNodes, and Jobs.

First, we should make it explicit in the documentation set (Reference manual + User manual) that some property names cannot be used.

Second, I don't understand why we check for parents in _properties.py. I believe it is not necessary and we may refactor some code to remove this condition. (to be investigated)

Thank you.

toan-quach:
Hi Irv, 

Thank you for the great insight! The reason that we check for "parent" as an attribute of the class _Property is so that we know to which said entity the "properties" belongs to. In your case, it belongs to a scenario. So whenever the "properties" is updated, we will auto-save the scenario to make sure the changes are up-to-date. I admit "parent" is not a good name and underscore is a must! Therefore, I've proposed to rename "parent" to "_entity_owner" (hopefully this is a lesser used name!) I hope the answer is satisfying!!

Thanks!

jrobinAV:
We should also add a note in the documentation to inform the users that "_entity_owner" is not allowed.

jrobinAV:
Hi Irv, 

We have implemented the fix. Here is the approved PR.
https://github.com/Avaiga/taipy-core/pull/469

We also added a config checker to let the developer know the property "_entity_owner" is restricted. Here is the approved PR.
https://github.com/Avaiga/taipy-config/pull/45

The change will be part of the next release 2.2.0 planned for Q2 2023.

We also updated the documentation. Here is the PR (almost) merged.
https://github.com/Avaiga/taipy-doc/pull/388

Thank you again for raising the issue.

Dr-Irv:
Great!

PR #469, opened 2023-02-10
feature/443 fix cannot use parent as a property key

toan-quach:
#443


jrobinAV (commented):
We can add a config checker to check that "_entity_owner" is not used as a custom property?

jrobinAV (approved):

Issue #442, opened 2023-01-23
BUG- Unexpected warnings during Taipy Core execution


FlorianJacta:
**Description**
Executing a code will sometimes throw unexpected warnings.

It will thrown this kind of warning:

`[2023-01-23 10:52:12,890][Taipy][WARNING] DATANODE_month_data_688cbf1e-1242-4ae8-94bb-8bb91de633b1 cannot be read because it has never been written. Hint: The data node may refer to a wrong path : .data\pickles\DATANODE_month_data_688cbf1e-1242-4ae8-94bb-8bb91de633b1.p`

However, these files are present in the .data and the pipeline is executed without any problem.

**How to reproduce**

- A code fragment
```python
#  Taipy Core Data nodes - CSV, pickle
from taipy.core.config import Config
import taipy as tp
import datetime as dt
import pandas as pd


def filter_current(df):

    return df

def count_values(df):
    return len(df)

default_data = {"Date":["2021-01-07 10:00:00", "2021-01-08 10:00:00"],
              "Value":[1,2]}

historical_data_cfg = Config.configure_data_node(id="historical_data",
                                                     default_data=default_data)
month_values_cfg =  Config.configure_data_node(id="month_data")
nb_of_values_cfg = Config.configure_data_node(id="nb_of_values")


task_filter_cfg = Config.configure_task(id="filter_current",
                                                 function=filter_current,
                                                 input=historical_data_cfg,
                                                 output=month_values_cfg)

task_count_values_cfg = Config.configure_task(id="count_values",
                                                 function=count_values,
                                                 input=month_values_cfg,
                                                 output=nb_of_values_cfg)

pipeline_cfg = Config.configure_pipeline(id="my_pipeline",
                                         task_configs=[task_filter_cfg,
                                                       task_count_values_cfg])

scenario_cfg = Config.configure_scenario(id="my_scenario",
                                         pipeline_configs=[pipeline_cfg])

#scenario_cfg = Config.configure_scenario_from_tasks(id="my_scenario",
#                                                    task_configs=[task_filter_current_cfg,
#                                                    task_count_values_cfg])

if __name__ == "__main__":  
    tp.Core().run()

    scenario_1 = tp.create_scenario(scenario_cfg, creation_date=dt.datetime(2022,10,7), name="Scenario 2022/10/7")
    scenario_1.submit()

[...]

PR #472, opened 2023-02-11
Fix/#442 move warnings from scheduler to managers

jrobinAV:
#442

To solve the problem, I propose moving the warning message out of the scheduler. 
The manager is now responsible for ensuring the readiness of a task/pipeline/scenario before submitting it. Moreover, it is easier not to log unnecessary messages. Indeed, before submitting a scenario, we just need to check its input data nodes, not all of them.


src/taipy/core/scenario/scenario.py
```diff
@@ -337,6 +349,24 @@ def remove_tag(self, tag: str):
 
         return tp.untag(self, tag)
 
+    def _get_inputs(self) -> Set[DataNode]:
+        dag = self.__build_dag()
+        return {node for node, degree in dict(dag.in_degree).items() if degree == 0 and isinstance(node, DataNode)}
+
+    def __build_dag(self):
```
toan-quach: I think we can reuse the build_dag function from pipeline.py and by making it a static function or a common function in `common.py` What do you think?

toan-quach (dismissed):

src/taipy/core/scenario/scenario.py
jrobinAV: You are right. Thx.

src/taipy/core/pipeline/pipeline.py
```diff
@@ -63,50 +64,48 @@ def __init__(
...
         self._parent_ids = parent_ids or set()
-        self._tasks = tasks
-
-        self._subscribers = _ListAttributes(self, subscribers or list())
         self._properties = _Properties(self, **properties)
-
         self._version = version or _VersionManagerFactory._build_manager()._get_latest_version()
 
-    @property  # type: ignore
-    def parent_id(self):
```
toan-quach, line 78: oh this reminds me, do we have a plan to completely remove the use of parent_id here?

src/taipy/core/common/_submittable.py
```diff
@@ -0,0 +1,80 @@
...
+import networkx as nx
+
+from ..common._utils import _Subscriber
+from ..data.data_node import DataNode
+from ..job.job import Job
+from ..task.task import Task
+from ._listattributes import _ListAttributes
+
+
+class _Submittable:
```
toan-quach, line 25: should the Task entity also extend from the _Submittable class since we can also submit the Task entity alone?

toan-quach (approved):
lgtm! just some minor questions 😃

src/taipy/core/pipeline/pipeline.py
jrobinAV: Yep, indeed, we should have a plan.
We should wait for the 3.0 in September to remove the deprecated APIS. 
https://github.com/Avaiga/taipy-core/issues/479

src/taipy/core/common/_submittable.py
jrobinAV: I did not implement a Task as a Submittable for 2 reasons: 
- A circular import : ( Job Depends on Task that depends on Submittable that depends on Job)
- There is no Graph to build. So there is almost no code to factorize.

Issue #473, opened 2023-02-13
Fix openpyxl conflict with modin


jrobinAV:
HOT FIX for 2.1

- Add to release notes
- Fix the versions dependency of openpyxl in setup.py to <3.1.0

PR #475, opened 2023-02-14
bug/#473 fix openpyxl version not supported by modin

toan-quach:
#473


setup.py
```diff
@@ -29,7 +29,7 @@
 requirements = [
     "pyarrow>=10.0.1,<11.0",
     "networkx>=2.6,<3.0",
-    "openpyxl>=3.0.7,<4.0",
+    "openpyxl<3.1.0",
```
jrobinAV: should we have a min version as well to be sure? 

```suggestion
    "openpyxl>=3.0.7,<3.1",
```

toan-quach: that certainly makes sense!

jrobinAV (approved):

PR #488, opened 2023-02-28
feat: Add _AbstractFileDataNode class

joaoandre-avaiga:


src/taipy/core/data/abstract_file.py
```diff
@@ -0,0 +1,101 @@
...
+from taipy.config.common.scope import Scope
+
+from .._version._version_manager_factory import _VersionManagerFactory
+from ..common.alias import DataNodeId, Edit
+from ..exceptions.exceptions import InvalidExposedType, MissingRequiredProperty, UnknownDatabaseEngine
+from .data_node import DataNode
+
+
+class _AbstractFileDataNode(DataNode):
+    """Abstract base class for data node implementations (CSVDataNode, ParquetDataNode, ExcelDataNode and
```
jrobinAV: And Json ?

src/taipy/core/data/abstract_file.py
```diff
@@ -0,0 +1,101 @@
...
+        if not self._last_edit_date and os.path.exists(self._path):
+            self.last_edit_date = datetime.now()  # type: ignore
+        if default_value is not None and not os.path.exists(self._path):
+            self.write(default_value)
+
+    def _to_pandas_df(self, df):
+        return pd.DataFrame(df)
+
+    def _to_modin_df(self, df):
+        return modin_pd.DataFrame(df)
```
jrobinAV: **About `_to_pandas_df` and `_to_modin_df`:** 
These are only for tabular data, and not specific to file data. For example, Json doesn't have these methods. Am I correct?

src/taipy/core/data/abstract_file.py
```diff
@@ -0,0 +1,101 @@
...
+        validity_period: Optional[timedelta] = None,
+        edit_in_progress: bool = False,
+        properties: Dict = None,
+    ):
+        if properties is None:
+            properties = {}
+        if missing := set(self._REQUIRED_PROPERTIES) - set(properties.keys()):
+            raise MissingRequiredProperty(
+                f"The following properties " f"{', '.join(x for x in missing)} were not informed and are required"
+            )
```
jrobinAV: This is not specific to file-based data nodes. Right? 
I also believe it must be executed before calling `super().__init__()`. 
Can we put it in an upper level method. In DataNode?

src/taipy/core/data/abstract_file.py
```diff
@@ -0,0 +1,103 @@
...
+            edit_in_progress,
+            **properties,
+        )
+        if not self._last_edit_date and isfile(self._path):
+            self.last_edit_date = datetime.now()  # type: ignore
+
+    def __build_path(self):
+        from taipy.config.config import Config
+
+        dir_path = pathlib.Path(Config.global_config.storage_folder) / "pickles"
```
jrobinAV: We need to change the "pickles" folder name.

src/taipy/core/data/abstract_file.py
joaoandre-avaiga: Actually this does not work, unfortunatelly, neither on the AbstractClass nor on Datanode. Because the _REQUIRED_PROPERTIES are diferent for each file data node, and we don't have a way for the parent class to actually know. We could implement a _check_required_properties method on either _AbstractFile or on Datanode and call it on the chield class, otherwise won't really work.

jrobinAV: Oh, yes... I would create a method in Datanode so it can be used in AbstractFile and other like SQLDataNOde, etc...

src/taipy/core/data/abstract_file.py
```diff
@@ -0,0 +1,103 @@
...
+        if missing := set(self._REQUIRED_PROPERTIES) - set(properties.keys()):
+            raise MissingRequiredProperty(
+                f"The following properties " f"{', '.join(x for x in missing)} were not informed and are required"
+            )
+        # default_value = properties.pop(self.__DEFAULT_DATA_KEY, None)
+        self._path = properties.get(self.__PATH_KEY, properties.get(self.__DEFAULT_PATH_KEY))
+        if self._path is None:
+            self._path = self.__build_path()
+
+        properties[self.__PATH_KEY] = self._path
```
jrobinAV: Should we move this five lines after calling the super().__init__()?

src/taipy/core/data/abstract_file.py
joaoandre-avaiga: So things have change in this case. I've noticed that the file type datanodes don't have any required properties anymore(The attribute is empty in all of the classes), the only information that would raise a MissingRequiredProperty is the `path`, which it won't anymore because we added the path generation when is not informed. For this reason, I just removed this check from all of the file base datanodes.

As far as SQL goals, it has a implementation on the AbstractSQL class that is already in use.

src/taipy/core/data/abstract_file.py
joaoandre-avaiga: True. I've changed the function to properly point to the correct folder.

src/taipy/core/data/csv.py
```diff
@@ -113,8 +104,13 @@ def __init__(
             version or _VersionManagerFactory._build_manager()._get_latest_version(),
             validity_period,
             edit_in_progress,
-            **properties,
+            properties=properties,
         )
+        self._path = properties.get(self.__PATH_KEY, properties.get(self.__DEFAULT_PATH_KEY))
+        if not self._path:
+            self._path = self._build_path(self.storage_type())
+        properties[self.__PATH_KEY] = self._path
```
jrobinAV, line 116: should this line be on `self.properties` and not `properties` ?

src/taipy/core/data/abstract_file.py
```diff
@@ -0,0 +1,82 @@
...
+            version or _VersionManagerFactory._build_manager()._get_latest_version(),
+            validity_period,
+            edit_in_progress,
+            **properties,
+        )
+
+    def _build_path(self, storage_type):
+        from taipy.config.config import Config
+
+        folder = "pickles" if storage_type == "pickle" else "data_nodes"
```
jrobinAV:
```suggestion
        folder = self.__FOLDER_MAP.get(storage_type, "unknown")
```
I don't like the idea of mixing data files (csv, excel, ...) with data nodes. I would not use data_nodes.

src/taipy/core/data/abstract_file.py
```diff
@@ -0,0 +1,82 @@
...
+            parent_ids,
+            last_edit_date,
+            edits,
+            version or _VersionManagerFactory._build_manager()._get_latest_version(),
+            validity_period,
+            edit_in_progress,
+            **properties,
+        )
+
+    def _build_path(self, storage_type):
```
jrobinAV, line 29: May be we can pass the id as a parameter of the `_build_path` method.
```suggestion
    def _build_path(self, id, storage_type):
```

src/taipy/core/data/abstract_file.py
```diff
@@ -0,0 +1,82 @@
...
+        from taipy.config.config import Config
+
+        folder = "pickles" if storage_type == "pickle" else "data_nodes"
+        dir_path = pathlib.Path(Config.global_config.storage_folder) / folder
+        if not dir_path.exists():
+            dir_path.mkdir(parents=True, exist_ok=True)
+        return dir_path / f"{self.id}.{self.__EXTENSION_MAP.get(storage_type)}"
```
jrobinAV, line 36:
```suggestion
        return dir_path / f"{id}.{self.__EXTENSION_MAP.get(storage_type)}"
```

src/taipy/core/data/abstract_file.py
```diff
@@ -0,0 +1,82 @@
...
+    __EXPOSED_TYPE_PANDAS = "pandas"
+    __EXPOSED_TYPE_MODIN = "modin"
+    __VALID_STRING_EXPOSED_TYPES = [__EXPOSED_TYPE_PANDAS, __EXPOSED_TYPE_NUMPY, __EXPOSED_TYPE_MODIN]
+
+    __PATH_KEY = "path"
+    __DEFAULT_PATH_KEY = "default_path"
+    __HAS_HEADER_PROPERTY = "has_header"
+
+    _REQUIRED_PROPERTIES: List[str] = []
+    __EXTENSION_MAP = {"csv": "csv", "excel": "xlsx", "parquet": "parquet", "pickle": "p", "json": "json"}
```
jrobinAV, line 27:
```suggestion
    __FOLDER_MAP = {"csv": "csvs", "excel": "excels", "parquet": "parquets", "pickle": "pickles", "json": "jsons"}
    __EXTENSION_MAP = {"csv": "csv", "excel": "xlsx", "parquet": "parquet", "pickle": "p", "json": "json"}
```

src/taipy/core/data/abstract_file.py
```diff
@@ -0,0 +1,82 @@
...
+            name,
+            owner_id,
+            parent_ids,
+            last_edit_date,
+            edits,
+            version or _VersionManagerFactory._build_manager()._get_latest_version(),
+            validity_period,
+            edit_in_progress,
+            **properties,
+        )
```
jrobinAV: Since we do nothing specific in the init method, I would just remove it. 
The children's __init__ methods would call directly the DataNode.__init__. 

What do you think?

src/taipy/core/data/abstract_file.py
```diff
@@ -0,0 +1,82 @@
...
+    """Abstract base class for data node implementations (CSVDataNode, ParquetDataNode, ExcelDataNode,
+    PickleDataNode and JSONDataNode) that are file based."""
+
+    __STORAGE_TYPE = "NOT_IMPLEMENTED"
+
+    __EXPOSED_TYPE_PROPERTY = "exposed_type"
+    __EXPOSED_TYPE_NUMPY = "numpy"
+    __EXPOSED_TYPE_PANDAS = "pandas"
+    __EXPOSED_TYPE_MODIN = "modin"
+    __VALID_STRING_EXPOSED_TYPES = [__EXPOSED_TYPE_PANDAS, __EXPOSED_TYPE_NUMPY, __EXPOSED_TYPE_MODIN]
```
jrobinAV: I believe it should not be in the `_AbstractFileDataNode` class. JsonDataNode (a file based data node) does not use them, and SQLTableDataNode (NOT a file base data node) does use them.

src/taipy/core/data/abstract_file.py
```diff
@@ -0,0 +1,82 @@
...
+from typing import Dict, List, Optional, Set
+
+from taipy.config.common.scope import Scope
+
+from .._version._version_manager_factory import _VersionManagerFactory
+from ..common.alias import DataNodeId, Edit
+from .data_node import DataNode
+
+
+class _AbstractFileDataNode(DataNode):
```
jrobinAV: I believe it does not require to inherit from DataNode if we pass id as a parameter of the build path method.

tests/core/data/test_csv_data_node.py
```diff
@@ -225,10 +217,6 @@ def test_set_path(self):
         dn.path = "bar.csv"
         assert dn.path == "bar.csv"
 
-    def test_raise_error_when_path_not_exist(self):
```
jrobinAV, line 228: Can we replace this test with a test that the data node is created, the path is correct and the data node can be written?
SHould have a tessts for other types (excel, json, parquet) as well.

jrobinAV (commented):
If that s ok for you, I will propose some changes directly to your branch.

src/taipy/core/data/abstract_file.py
joaoandre-avaiga: Yes, you're right. I mixed things up.

src/taipy/core/data/csv.py
joaoandre-avaiga: It can be self.properties since is after the super().__init__. I will change it

src/taipy/core/data/abstract_file.py
joaoandre-avaiga: I don't think we need a map for the folders, since is the same name as the storage type + s.

src/taipy/core/data/abstract_file.py
joaoandre-avaiga: That would avoid the the recursion error, but still we would be able to call this method only after calling the init method, since the `id` is generated on the DataNode class

src/taipy/core/data/abstract_file.py
joaoandre-avaiga: That would imply in having two inheritances, from AbstractFile and DataNode. Which is not a deal breaker, just a little different from what we did on the SQL classes.

src/taipy/core/data/csv.py
joaoandre-avaiga: Just tested and changing to self.properties also triggers a recursion error.

tests/core/data/test_csv_data_node.py
joaoandre-avaiga: I've added a test to make sure the file is written when a default_data is passed, otherwise it will not be written(is the same behavior as pickle). However, this cannot be factorize because it depends on the write method on each DataNode class.

src/taipy/core/data/abstract_file.py
jrobinAV: You are right

src/taipy/core/data/data_node.py
```diff
@@ -30,7 +30,7 @@
 from ..common._reload import _reload, _self_reload, _self_setter
 from ..common._warnings import _warn_deprecated
 from ..common.alias import DataNodeId, Edit, JobId
-from ..exceptions.exceptions import NoData
+from ..exceptions.exceptions import MissingRequiredProperty, NoData
```
toan-quach, line 33: Do we use this exception in this file? I don't see any other changes in this file and the original version doesn't seem to raise this exception as well?

src/taipy/core/data/excel.py
```diff
@@ -237,56 +234,46 @@ def _read_as_numpy(self):
...
-    def _read_as_pandas_dataframe(self, sheet_names=None) -> Union[Dict[Union[int, str], pd.DataFrame], pd.DataFrame]:
+    def _do_read_excel(self, engine, sheet_names, kwargs) -> pd.DataFrame:
+        df = pd.read_excel(
+            self._path,
+            sheet_name=sheet_names,
+            **kwargs,
+        )
+        if isinstance(df, dict):
+            for key, value in df.items():
+                df[key] = pd.DataFrame(value) if engine == "pandas" else modin_pd.DataFrame(value)
```
toan-quach, line 245: I take that this is to simplify the read excel file with modin process, which was messy since the beginning (the issue I raised with the modin team is still not fixed 😞 ). `read_excel` will return a dict of {'sheet_name': pandas.DataFrame}` if the excel file has multiple sheet names 😃 
```suggestion
                df[key] = value if engine == "pandas" else modin_pd.DataFrame(value)
```

src/taipy/core/data/excel.py
```diff
@@ -237,56 +234,46 @@ def _read_as_numpy(self):
...
+            return df
+        return pd.DataFrame(df) if engine == "pandas" else modin_pd.DataFrame(df)
```
toan-quach, line 247:
```suggestion
        return df if engine == "pandas" else modin_pd.DataFrame(df)
```

src/taipy/core/data/excel.py
```diff
@@ -237,56 +234,46 @@ def _read_as_numpy(self):
...
         return sheets.to_numpy()
 
-    def _read_as_pandas_dataframe(self, sheet_names=None) -> Union[Dict[Union[int, str], pd.DataFrame], pd.DataFrame]:
+    def _do_read_excel(self, engine, sheet_names, kwargs) -> pd.DataFrame:
+        df = pd.read_excel(
+            self._path,
+            sheet_name=sheet_names,
+            **kwargs,
+        )
+        if isinstance(df, dict):
```
toan-quach, line 243: Why not make it check:
```suggestion
        if engine == "modin:
                if isinstance(df, dict):
                        for key, value in df.items():
                                ...
                else:
                        modin_pd.DataFrame(df)
```
So that if the engine is `pandas`, we can just return immediately?

toan-quach (approved):
Thanks for cleaning up the read modin function in excel dn!!! I have only some minor remarks 😃

jrobinAV (approved):

jrobinAV:
@joaoandre-avaiga I took the liberty to merge your PR and move the issue to Done.

PR #490, opened 2023-02-28
feature/#393 added doc for return type of submit methods

toan-quach:
Will merge after merging PR of branch #510

### DIFF
--- a/src/taipy/core/pipeline/pipeline.py
+++ b/src/taipy/core/pipeline/pipeline.py
@@ -255,7 +255,8 @@ class Pipeline(_Entity, _Submittable):
                 mode.
             timeout (Union[float, int]): The maximum number of seconds to wait for the jobs to be finished before
                 returning.
-
+        Returns:
+            The list of created `Job^`.
         """
         from ._pipeline_manager_factory import _PipelineManagerFactory
 
--- a/src/taipy/core/pipeline/pipeline.py
+++ b/src/taipy/core/pipeline/pipeline.py
@@ -256,7 +256,7 @@ class Pipeline(_Entity, _Submittable):
             timeout (Union[float, int]): The maximum number of seconds to wait for the jobs to be finished before
                 returning.
         Returns:
-            A list of created `Job^`.
+            A list of created `Job^`s.
         """
         from ._pipeline_manager_factory import _PipelineManagerFactory
 
--- a/src/taipy/core/pipeline/pipeline.py
+++ b/src/taipy/core/pipeline/pipeline.py
@@ -256,7 +256,7 @@ class Pipeline(_Entity, _Submittable):
             timeout (Union[float, int]): The maximum number of seconds to wait for the jobs to be finished before
                 returning.
         Returns:
-            The list of created `Job^`.
+            A list of created `Job^`.
         """
         from ._pipeline_manager_factory import _PipelineManagerFactory
 
--- a/src/taipy/core/scenario/scenario.py
+++ b/src/taipy/core/scenario/scenario.py
@@ -300,6 +300,8 @@ class Scenario(_Entity, _Submittable):
                 asynchronous mode.
             timeout (Union[float, int]): The optional maximum number of seconds to wait for the jobs to be finished
                 before returning.
+        Returns:
+            The dictionary with `PipelineId^` as the key and a list of created `Job^` as its value.
         """
         from ._scenario_manager_factory import _ScenarioManagerFactory
 
--- a/src/taipy/core/scenario/scenario.py
+++ b/src/taipy/core/scenario/scenario.py
@@ -301,8 +301,7 @@ class Scenario(_Entity, _Submittable):
             timeout (Union[float, int]): The optional maximum number of seconds to wait for the jobs to be finished
                 before returning.
         Returns:
-            A dictionary, with `PipelineId^` as the key and a list of created `Job^` corresponding
-                to that pipeline as its value.
+            A list of created `Job^`s.
         """
         from ._scenario_manager_factory import _ScenarioManagerFactory
 
--- a/src/taipy/core/scenario/scenario.py
+++ b/src/taipy/core/scenario/scenario.py
@@ -301,7 +301,7 @@ class Scenario(_Entity, _Submittable):
             timeout (Union[float, int]): The optional maximum number of seconds to wait for the jobs to be finished
                 before returning.
         Returns:
-            The dictionary with `PipelineId^` as the key and a list of created `Job^` as its value.
+            A dictionary, with `PipelineId^` as the key and a list of created `Job^` corresponding to that pipeline as its value.
         """
         from ._scenario_manager_factory import _ScenarioManagerFactory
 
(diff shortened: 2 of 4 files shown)


src/taipy/core/pipeline/pipeline.py
```diff
@@ -255,7 +255,8 @@ def submit(
                 mode.
             timeout (Union[float, int]): The maximum number of seconds to wait for the jobs to be finished before
                 returning.
-
+        Returns:
+            The list of created `Job^`.
```
```suggestion
            A list of created `Job^`.
```

src/taipy/core/scenario/scenario.py
```diff
@@ -300,6 +300,8 @@ def submit(
                 asynchronous mode.
             timeout (Union[float, int]): The optional maximum number of seconds to wait for the jobs to be finished
                 before returning.
+        Returns:
+            The dictionary with `PipelineId^` as the key and a list of created `Job^` as its value.
```
trgiangdo:
```suggestion
            A dictionary, with `PipelineId^` as the key and a list of created `Job^` corresponding to that pipeline as its value.
```

src/taipy/core/taipy.py
```diff
@@ -74,10 +74,17 @@ def submit(
...
+            in asynchronous mode.
+        timeout (Union[float, int]): The optional maximum number of seconds to wait
+            for the jobs to be finished before returning.
+    Returns:
+        The created `Job^` or a collection of the created `Job^` depends on the submitted entity.
 
+        - If a `Scenario^` is provided, it will return a dictionary with `PipelineId^` as the key and
+            a list of `Job^` as its value.
+        - If a `Pipeline^` is provided, it will return a list of `Job^`.
+        - If a `Task^` is provided, it will return a `Job^`.
```
trgiangdo:
```suggestion
        - If a `Task^` is provided, it will return the created `Job^`.
```
Just to be consistent with other doc

src/taipy/core/taipy.py
```diff
@@ -74,10 +74,17 @@ def submit(
...
-            returning.
+        wait (bool): Wait for the scheduled jobs created from the submission to be finished
+            in asynchronous mode.
+        timeout (Union[float, int]): The optional maximum number of seconds to wait
+            for the jobs to be finished before returning.
+    Returns:
+        The created `Job^` or a collection of the created `Job^` depends on the submitted entity.
 
+        - If a `Scenario^` is provided, it will return a dictionary with `PipelineId^` as the key and
+            a list of `Job^` as its value.
```
trgiangdo:
```suggestion
            a list of `Job^` corresponding to that pipeline as its value.
```

trgiangdo (commented):
I have some minor remarks.

Do we need to update anything on the documentation too?
Like on [documentation on `submit()`](https://docs.taipy.io/en/latest/manuals/core/entities/scheduling-and-job-execution/#submit-a-scenario-pipeline-or-task), we are not specifying the return type yet. A small update would be better imo.

toan-quach:
@trgiangdo 
To answer your remark, yes we do have an update on taipy-doc as well :D

trgiangdo (dismissed):
Great. I did not notice the update on doc. Thank you :D

src/taipy/core/pipeline/pipeline.py
```diff
@@ -255,7 +255,8 @@ def submit(
                 mode.
             timeout (Union[float, int]): The maximum number of seconds to wait for the jobs to be finished before
                 returning.
-
+        Returns:
+            A list of created `Job^`.
```
jrobinAV:
```suggestion
            A list of created `Job^`s.
```

src/taipy/core/scenario/scenario.py
```diff
@@ -300,6 +300,9 @@ def submit(
                 asynchronous mode.
             timeout (Union[float, int]): The optional maximum number of seconds to wait for the jobs to be finished
                 before returning.
+        Returns:
+            A dictionary, with `PipelineId^` as the key and a list of created `Job^` corresponding
```
jrobinAV: For consistency, we should return the same format as the pipeline submission. A list of created `Job^`s.

No?

src/taipy/core/taipy.py
```diff
@@ -74,10 +74,17 @@ def submit(
...
+        wait (bool): Wait for the scheduled jobs created from the submission to be finished
+            in asynchronous mode.
+        timeout (Union[float, int]): The optional maximum number of seconds to wait
+            for the jobs to be finished before returning.
+    Returns:
+        The created `Job^` or a collection of the created `Job^` depends on the submitted entity.
 
+        - If a `Scenario^` is provided, it will return a dictionary with `PipelineId^` as the key and
+            a list of `Job^` corresponding to that pipeline as its value.
+        - If a `Pipeline^` is provided, it will return a list of `Job^`.
```
```suggestion
        The `list of created `Job^s`.
```

jrobinAV (commented):
In order to make it consistent, I would prefer taipy.submit to always return a list of jobs whatever the input entity.

task.submit should return a single job.

src/taipy/core/scenario/scenario.py
toan-quach: Well no not exactly, since when we submit a scenario, it also means we're submitting multiple pipelines, and as the current implementation puts it, we need a way to separate the jobs based on the pipeline/pipelineid. Hence, we are return a dictionary with pipelineId as key and list of jobs as values 😃 ofcource we can put all jobs into a single list, but it'll be a hassle trying to differentiate which pipeline it belongs to (going through 2 layers of parents: job -> task -> pipeline isn't particularly fun).

toan-quach: Despite that, I agree that having some consistency would be so much better, just not sure how we can keep the consistency with pipeline submission in this case

src/taipy/core/taipy.py
toan-quach: Would it be confusing as it depends on the input entity, we will return different type of objects?

jrobinAV: Yes. That was under the assumption we return only a list. but you raised a good point about retrieving the pipeline. Let s discuss this in a daily meeting.

jrobinAV (changes requested):
We found a bug with the submission of pipelines/scenarios:

https://github.com/Avaiga/taipy-core/issues/494
The bug fix includes a change in the return type of the submit methods.
- `tp.submit(entity)` always returns a list of jobs.
- `scenario.submit()` returns a list of jobs.
- `pipeline.submit()` always returns a list of jobs.
- `task.submit()` returns a job.

This PR should be re done after 494 is fixed

src/taipy/core/scenario/scenario.py
toan-quach: This has been changed to "A list of created `Job^`s."

src/taipy/core/taipy.py
toan-quach: This has been changed to:
"""
        - If a `Scenario^` or a `Pipeline^` is provided, it will return a list of `Job^`.
        - If a `Task^` is provided, it will return the created `Job^`.
"""

jrobinAV (approved):

Issue #489, opened 2023-02-21
Implement an API to add migration functions and a checker for migration functions


trgiangdo:
**Description**

- [x] Implement `Config.add_migration_function` API
```python
Config.set_version_migration(
    source_version="1.0",
    target_version="2.0",
    data_node_migration_fct=...,
    task_migration_fct=...,
    pipeline_migration_fct=...,
    scenario_migration_fct=...,
)
```
to add migration functions to the Config singleton.

- [x] Implement a `MigrationFunctionChecker` to check if there is any migration function has been assigned for a specific "source_version"&rarr;"target_version" migration.

jrobinAV:
I would not implement the auto=True/False for now.
I believe we can have a first version without the automatic migration.
ANd once we are ready, we will open the auto parameter. What do you think ?

trgiangdo:
Agree. The `auto` parameter will be added with the automatic migration feature later.

PR #492, opened 2023-02-28
Feat/#489 - Version migration API on Config

trgiangdo:
https://github.com/Avaiga/taipy-core/issues/489


src/taipy/core/config/checkers/_version_migration_config_checker.py
```diff
@@ -0,0 +1,60 @@
...
+        if all(
+            getattr(version_migration_config, property_to_check) is None for property_to_check in properties_to_check
+        ):
+            self._warning(
+                "migration_fct",
+                None,
+                f"There is no migration function defined in VersionMigrationConfig `{version_migration_config_id}`"
+                f' from version "{getattr(version_migration_config, VersionMigrationConfig._SOURCE_VERISON_KEY)}"'
+                f' to version "{getattr(version_migration_config, VersionMigrationConfig._TARGET_VERISON_KEY)}".',
+            )
```
trgiangdo: This section here check if all the `..._migration_fct` is None, it will throw a warning said that there is no migration function defined (which make the VersionMigrationConfig kinda useless).

For now, the `field` parameter of `_warning()` is `"migration_fct"`, but in fact it should be `["data_node_migration_fct", "task_migration_fct", "pipeline_migration_fct", "scenario_migration_fct"]`. Is this ok?

Also, is the message clear enough?

tests/conftest.py
```diff
@@ -323,11 +325,18 @@ def init_config():
             ("configure_scenario_from_tasks", ScenarioConfig._configure_from_tasks),
         ],
     )
+    _inject_section(
+        VersionMigrationConfig,
+        "version_migrations",
+        VersionMigrationConfig.default_config(),
+        [("set_version_migration", VersionMigrationConfig._set)],
+    )
```
trgiangdo: What do you think about the method name `Config.set_version_migration()`?

src/taipy/core/config/__init__.py
```diff
@@ -71,9 +74,16 @@
         ("configure_scenario_from_tasks", ScenarioConfig._configure_from_tasks),
     ],
 )
+_inject_section(
+    VersionMigrationConfig,
+    "version_migrations",
```
jrobinAV: I would not use the name "version migration". We don't migrate versions, we migrate data (entities). What about just `MigrationConfig` and "migration_functions"?

src/taipy/core/config/__init__.py
```diff
@@ -71,9 +74,16 @@
...
+    VersionMigrationConfig.default_config(),
+    [("set_version_migration", VersionMigrationConfig._set)],
```
jrobinAV: "add_migration_function"?

src/taipy/core/config/checkers/_version_migration_config_checker.py
```diff
@@ -0,0 +1,60 @@
...
+# specific language governing permissions and limitations under the License.
+
+from taipy.config._config import _Config
+from taipy.config.checker._checkers._config_checker import _ConfigChecker
+from taipy.config.checker.issue_collector import IssueCollector
+
+from ..version_migration_config import VersionMigrationConfig
+
+
+class _VersionMigrationConfigChecker(_ConfigChecker):
```
jrobinAV: MigrationConfigChecker ?

src/taipy/core/config/version_migration_config.py
```diff
@@ -0,0 +1,169 @@
...
+
+from taipy.config._config import _Config
+from taipy.config.common._template_handler import _TemplateHandler as _tpl
+from taipy.config.config import Config
+from taipy.config.section import Section
+
+from .._version._version_manager import _VersionManager
+
+
+class VersionMigrationConfig(Section):
```
jrobinAV: MigrationConfig?

tests/conftest.py
jrobinAV: I don't really like it. What about "add_migration_function"?

tests/core/config/test_default_config.py
```diff
@@ -30,7 +32,7 @@ def _test_default_data_node_config(dn_config: DataNodeConfig):
     assert dn_config.id is not None
     assert dn_config.storage_type == "pickle"
     assert dn_config.scope == Scope.SCENARIO
-    assert len(dn_config.properties) == 0
+    assert len(dn_config.properties) == 0  # type: ignore
```
jrobinAV, line 34: Why do we need to ignore the type? Is it a code smell indicating we have a bad design?

src/taipy/core/config/checkers/_version_migration_config_checker.py
jrobinAV: I believe it is.

src/taipy/core/config/version_migration_config.py
```diff
@@ -0,0 +1,169 @@
...
+        )
+        self.data_node_migration_fct = as_dict.pop(self._DATA_NODE_MIGRATION_FCT_KEY, self.data_node_migration_fct)
+        self.task_migration_fct = as_dict.pop(self._TASK_MIGRATION_FCT_KEY, self.task_migration_fct)
+        self.pipeline_migration_fct = as_dict.pop(self._PIPELINE_MIGRATION_FCT_KEY, self.pipeline_migration_fct)
+        self.scenario_migration_fct = as_dict.pop(self._SCENARIO_MIGRATION_FCT_KEY, self.scenario_migration_fct)
+
+        self._properties.update(as_dict)
+
+    @staticmethod
+    def _set(
```
jrobinAV: _add_migration_function?

tests/core/config/test_version_migration_config.py
```diff
@@ -0,0 +1,116 @@
...
+
+    Config.set_version_migration("version_migrations3", "latest", "latest")
+    assert len(Config.version_migrations) == 4
+
+
+def test_version_migration_getitem():
+    version_migration_config_id = "version_migrations1"
+    version_migration = Config.set_version_migration(
+        version_migration_config_id,
+        target_version="latest",
```
jrobinAV: SHould this raise a config error ?

tests/core/config/test_version_migration_config.py
```diff
@@ -0,0 +1,116 @@
...
+
+import pytest
+
+from src.taipy.core._version._version_manager import _VersionManager
+from src.taipy.core.exceptions.exceptions import NonExistingVersion
+from taipy.config.config import Config
+
+
+def test_version_migration_config_creation():
+    version_migration_config = Config.set_version_migration("version_migrations1", "latest", "latest")
```
jrobinAV: The usage seems too complex for what it does. I believe we can improve it.
Having to provide 3 strings (id, source, target) looks useless to me. Only one seems needed at this point.

Let's discuss it.

src/taipy/core/config/version_migration_config.py
jrobinAV: Could it be a UniqueSection holding a dictionary of migration functions? We may gain in terms of readability and maintainability.
What do you think?

tests/conftest.py
trgiangdo: it's not really "add", but "set" functions for the migration. How about `Config.set_migration_functions()`?

src/taipy/core/config/__init__.py
trgiangdo: I like `MigrationConfig`.

But `migration_functions` is not sufficient to describe the config here.
Because when we have multiple "`migration_functions`" (or multiple `MigrationConfig`), each migration function actually contains multiple functions.

tests/core/config/test_default_config.py
trgiangdo: It is because we does not define a specific type for `properties` on DataNodeConfig, which cause the `len()` method to complain about incorrect type.

tests/core/config/test_version_migration_config.py
trgiangdo: It's not, because the `_VersionManager` can translate "latest" -> the latest version. 
Similar for "dev" or "development" -> the development version.

When implement the feature itself, we may want to limit the target and source version to production versions only, but for now, I'll leave it as it is.

src/taipy/core/config/version_migration_config.py
trgiangdo: There maybe multiple Migration Config when there is multiple versions on production (0.9 -> 1.0 -> 1.1 -> ...).

I'm not sure if I understand your point correctly.

tests/core/config/test_version_migration_config.py
trgiangdo: For now, the default source and target version is latest.

However, when a user define a MigrationConfig, he/she should specify the source and target version so Taipy know which one to migrate.

src/taipy/core/config/version_migration_config.py
jrobinAV: What I had in mind is to have a single UniqueSection called MigrationConfig. It makes more sense to me since the functions are not independent; they are linked to each other by the version number. Plus, as a client, it is simpler to have a single entry point to retrieve all the functions.

This MigrationConfig could hold all the migration functions for all the versions.
it could be stored in a variable as a `Dict[targetVersionNumber, dict[config_id, function]]`
for example.
We could implement some smart accessors, like 
```
def get_functions(versionNumber: str, config_id: str) -> List[Callable]:
```

`Config.migrationConfig.get_functions("0.9", "input_data")` can return all the functions to migrate data nodes from config input_data from version 0.9 to the latest version. Meaning all functions to use to migrate data nodes to 1.0 version then to 1.1 version. So when taipy reads an entity, we just need to execute the functions returned by `Config.migrationConfig.get_functions(entity.version_number, entity.config_id)` in the correct order.

Does it make sense?

src/taipy/core/config/version_migration_config.py
jrobinAV: I believe we don't need the source version.

jrobinAV: maybe it is easier for the user to have multiple methods : 

```
Config._add_data_node_migration_function(target_version, data_node_config)
Config._add_task_migration_function(target_version, task_config)
Config._add_pipeline_migration_function(target_version, pipeline_config)
Config._add_scenario_migration_function(target_version, scenario_config)
```

What do you think?

tests/conftest.py
jrobinAV: Oh, you are right. That's because I misunderstood the VersionMigrationConfig section. I though it was a unique section.
If we implement it as a Section, I agree that `set_migration_function` is better.
If we implement it as a UniqueSection, I prefer `add_migration_function`.

src/taipy/core/config/__init__.py
jrobinAV: Yes sorry, I though it was a UniqueSection, and not a Section.

src/taipy/core/config/__init__.py
jrobinAV: Sorry, I though it was a UniqueSection and not a Section.
If we implement it as a Section, I prefer set_migration_function.
If we implement it as a UniqueSection, I prefer add_migration_function.

src/taipy/core/config/version_migration_config.py
```diff
@@ -0,0 +1,169 @@
...
+    _SOURCE_VERISON_KEY = "source_version"
+    _TARGET_VERISON_KEY = "target_version"
+    _DATA_NODE_MIGRATION_FCT_KEY = "data_node_migration_fct"
+    _TASK_MIGRATION_FCT_KEY = "task_migration_fct"
+    _PIPELINE_MIGRATION_FCT_KEY = "pipeline_migration_fct"
+    _SCENARIO_MIGRATION_FCT_KEY = "scenario_migration_fct"
+
+    def __init__(
+        self,
+        id: str,
```
jrobinAV: If we use a UniqueSection, the user does not need to provide an id.

src/taipy/core/config/version_migration_config.py
```diff
@@ -0,0 +1,169 @@
...
+        source_version: str,
```
jrobinAV: Since the production versions are ordered, I believe the target_version is sufficient. We don't need the source_version. 
Am I right?

tests/core/config/test_version_migration_config.py
jrobinAV: I mean what is the meaning of migrating from version latest to version latest?

src/taipy/core/config/checkers/_migration_config_checker.py
```diff
@@ -0,0 +1,51 @@
...
+
+    def _check_valid_version(self, source_version):
+        try:
+            _VersionManager._replace_version_number(source_version)
+        except NonExistingVersion:
+            self._error(
+                MigrationConfig._MIGRATION_FCTS_KEY,
+                source_version,
+                "The source version for a migration function must be a valid version number.",
+            )
```
jrobinAV: I don't get this checker function `_check_valid_version()`. What is the use case covered?

We only want to migrate production versions. The user must provide a valid production version number. So I don't see the purpose of using the `replace_version_number(source)` function.

src/taipy/core/config/migration_config.py
```diff
@@ -0,0 +1,192 @@
...
+from .scenario_config import ScenarioConfig
+from .task_config import TaskConfig
+
+
+class MigrationConfig(UniqueSection):
+    """
+    Configuration fields needed to register migration functions from an old version to newer one.
+
+    Attributes:
+        migration_fcts (Dict[str, Dict[str, Callable]]): A dictionary that maps the version that entities are
```
jrobinAV, line 28: Each migration function should correspond to the `target_version` and not the `source_version`. That is what the user wants to provide.

However, the client (Taipy core repositories) wants to retrieve all methods to apply to an existing entity. So the client wants to provide the `source_target` and the `config_id`.

That means we should have another accessor method in MigrationConfig named for example `get_migration_functions`.
It should take as a parameter the `source_version` and the `config_id` and return all functions to apply, i.e. all functions where the target_version is greater than the source_version provided.

Does it make sense?

src/taipy/core/config/migration_config.py
```diff
@@ -0,0 +1,192 @@
...
+
+        self._properties.update(as_dict)
+
+    @staticmethod
+    def _add_data_node_migration_function(
+        source_version: str,
+        data_node_config: Union[DataNodeConfig, DataNodeId],
+        migration_fct: Callable,
+        **properties,
+    ):
```
jrobinAV: As a user, when developing version 2.0 of my app, I want to provide a single function to migrate old entities (from multiple old versions <2.0) to the single version 2.0.
So the API should ask for the target version, not the source version.

```suggestion
    def _add_data_node_migration_function(
        target_version: str,
        data_node_config: Union[DataNodeConfig, DataNodeId],
        migration_fct: Callable,
        **properties,
    ):
```

Ex:

Config._add_task_migration_function("2.0", my_datanode_cfg, my_fct)

src/taipy/core/config/migration_config.py
```diff
@@ -0,0 +1,192 @@
...
+                and returns a compatible DataNode with the next production version.
+            **properties (Dict[str, Any]): A keyworded variable length list of additional arguments.
+        Returns:
+            `MigrationConfig^`: The Migration configuration.
+        """
+        return MigrationConfig.__add_migration_function(source_version, data_node_config, migration_fct, **properties)
+
+    @staticmethod
+    def _add_task_migration_function(
+        source_version: str,
```
jrobinAV: same remark abt target_version.

src/taipy/core/config/migration_config.py
```diff
@@ -0,0 +1,192 @@
...
+                and returns a compatible DataNode with the next production version.
+            **properties (Dict[str, Any]): A keyworded variable length list of additional arguments.
+        Returns:
+            `MigrationConfig^`: The Migration configuration.
+        """
+        return MigrationConfig.__add_migration_function(source_version, task_config, migration_fct, **properties)
+
+    @staticmethod
+    def _add_pipeline_migration_function(
+        source_version: str,
```
jrobinAV: same remark abt target_version.

src/taipy/core/config/migration_config.py
```diff
@@ -0,0 +1,192 @@
...
+                and returns a compatible DataNode with the next production version.
+            **properties (Dict[str, Any]): A keyworded variable length list of additional arguments.
+        Returns:
+            `MigrationConfig^`: The Migration configuration.
+        """
+        return MigrationConfig.__add_migration_function(source_version, pipeline_config, migration_fct, **properties)
+
+    @staticmethod
+    def _add_scenario_migration_function(
+        source_version: str,
```
jrobinAV: same remark abt target_version.

src/taipy/core/config/migration_config.py
```diff
@@ -0,0 +1,192 @@
...
+                and returns a compatible DataNode with the next production version.
+            **properties (Dict[str, Any]): A keyworded variable length list of additional arguments.
+        Returns:
+            `MigrationConfig^`: The Migration configuration.
+        """
+        return MigrationConfig.__add_migration_function(source_version, scenario_config, migration_fct, **properties)
+
+    @staticmethod
+    def __add_migration_function(
+        source_version: str,
```
jrobinAV: same remark abt target_version.

src/taipy/core/config/checkers/_migration_config_checker.py
trgiangdo: Oh you're right. This function should check if the `source_version` is production or not. I'll update it

src/taipy/core/config/migration_config.py
trgiangdo: I think it makes more sense for the user to provide the source version (or both).

Like when look at a `Config.add_data_node_migration_function(...)` method, he/she can instantly conclude which version will be affected by the migration function.

src/taipy/core/config/checkers/_migration_config_checker.py
trgiangdo: In the newest commit, `replace_version_number(source)` is still used since the latest version maybe a production version.

This is subject to change, since if we want the user to provide the target or both source-target version, we will need this. If we only want the source, then we can remove this.

src/taipy/core/config/__init__.py
```diff
@@ -79,10 +79,10 @@
     "migration_functions",
     MigrationConfig.default_config(),
     [
-        ("add_data_node_migration_function", MigrationConfig._add_data_node_migration_function),
-        ("add_task_migration_function", MigrationConfig._add_task_migration_function),
-        ("add_pipeline_migration_function", MigrationConfig._add_pipeline_migration_function),
-        ("add_scenario_migration_function", MigrationConfig._add_scenario_migration_function),
+        ("add_data_node_migration_function", MigrationConfig._add_data_node_migration_function),  # type: ignore
```
jrobinAV: Why do we need to ignore type? Most of the time, it is a clue for a wrong design...

src/taipy/core/config/migration_config.py
jrobinAV: Actually, we want to migrate the entity from the source_version, but not only; All the entities from previous versions, as well. That's also why it seems easier for me to use the target_version.

trgiangdo: How can one define a function that can migrate from all version to a target version?

Sometimes it might be, sometimes it's not that simple

src/taipy/core/config/__init__.py
trgiangdo: Honestly, I have no idea.

It complains about the return type, but we do not declare any specific return type on the inject method.

The migration functions are subjects to change in the short future, so for now I ignore the message.

We will need to comeback on this soon

jrobinAV: OK, sounds good. We just need to be sure we don't forget to get back on it.

src/taipy/core/config/checkers/_migration_config_checker.py
jrobinAV: In my mind, when adding a migration function, defining the version is mandatory. Whether it is the source_version or the target_version, or both.

So, in the checker, we want to ensure:
1. that the version is actually provided
2. that the version provided is a valid production version.

Are we aligned on this?

If yes, I still don't get the point. What is the reason for replacing the version provided by the user?
Am I missing something?

src/taipy/core/config/migration_config.py
```diff
@@ -0,0 +1,192 @@
...
+            for k, v in u.items():
+                if isinstance(v, collections.abc.Mapping):
+                    d[k] = deep_update(d.get(k, {}), v)
+                else:
+                    d[k] = v
+            return d
+
+        migration_fcts = as_dict.pop(self._MIGRATION_FCTS_KEY)
+        for source_version in list(migration_fcts.keys()):
+            if (proper_version := _VersionManager._replace_version_number(source_version)) != source_version:
```
jrobinAV: I missed this before, but I have the same misunderstanding here. Why do we use `_VersionManager._replace_version_number(source_version)` and not `source_version` directly

src/taipy/core/config/migration_config.py
trgiangdo: After discussion, `target_version` is better.

src/taipy/core/config/checkers/_migration_config_checker.py
trgiangdo: We don't need to replace the version.

src/taipy/core/config/migration_config.py
trgiangdo: We don't need to replace the version.

src/taipy/core/config/checkers/_migration_config_checker.py
```diff
@@ -25,33 +25,51 @@ def __init__(self, config: _Config, collector: IssueCollector):
...
+                        "The target version for a migration function must be a production version.",
+                    )
+            except NonExistingVersion:
                 self._error(
                     MigrationConfig._MIGRATION_FCTS_KEY,
-                    version_number,
-                    "The source version for a migration function must be a production version.",
+                    target_version,
+                    "The target version for a migration function must be a valid version number.",
+                )
```
jrobinAV: Is this still needed? 'cause the if condition tests if the version is a production version. Whether the version is an experiment or does not exist at all. 
Plus, nobody raises a NonExistingVersion error.

I believe we can remove the try except.

src/taipy/core/config/migration_config.py
```diff
@@ -0,0 +1,188 @@
...
+    def _add_data_node_migration_function(
+        target_version: str,
+        data_node_config: Union[DataNodeConfig, DataNodeId],
+        migration_fct: Callable,
+        **properties,
+    ):
+        """Add a data node migration function to the configuration.
+
+        Parameters:
+            target_version (str): The version that entities are migrated from.
```
jrobinAV:
```suggestion
            target_version (str): The version that entities are migrated to.
```

src/taipy/core/config/migration_config.py
```diff
@@ -0,0 +1,188 @@
...
+    def _add_task_migration_function(
+        target_version: str,
+        task_config: Union[TaskConfig, TaskId],
+        migration_fct: Callable,
+        **properties,
+    ):
+        """Add a task migration function to the configuration.
+
+        Parameters:
+            target_version (str): The version that entities are migrated from.
```
jrobinAV:
```suggestion
            target_version (str): The version that entities are migrated to.
```

src/taipy/core/config/migration_config.py
```diff
@@ -0,0 +1,188 @@
...
+    def _add_pipeline_migration_function(
+        target_version: str,
+        pipeline_config: Union[PipelineConfig, PipelineId],
+        migration_fct: Callable,
+        **properties,
+    ):
+        """Add a pipeline migration function to the configuration.
+
+        Parameters:
+            target_version (str): The version that entities are migrated from.
```
jrobinAV:
```suggestion
            target_version (str): The version that entities are migrated to.
```

src/taipy/core/config/migration_config.py
```diff
@@ -0,0 +1,188 @@
...
+    def _add_scenario_migration_function(
+        target_version: str,
+        scenario_config: Union[ScenarioConfig, ScenarioId],
+        migration_fct: Callable,
+        **properties,
+    ):
+        """Add a scenario migration function to the configuration.
+
+        Parameters:
+            target_version (str): The version that entities are migrated from.
```
jrobinAV:
```suggestion
            target_version (str): The version that entities are migrated to.
```

src/taipy/core/config/checkers/_migration_config_checker.py
trgiangdo: You're right.

src/taipy/core/_core.py
```diff
@@ -45,8 +45,8 @@ def run(self, force_restart=False):
 
         This function check the configuration, start a dispatcher and lock the Config.
         """
-        self.__check_config()
         self.__manage_version(*self.cli_args)
+        self.__check_config()
```
trgiangdo: This is to create the production version first before checking the config to make sure the MigrationConfig checker can detect the production version

jrobinAV: Oh?!?!

So we first create a version before checking if the config is correct? 

What if the config has errors? A corrupted version will exist? and we will store a corrupted config?

I don't like the idea.

trgiangdo: You're right. I'll revert the change and find another way to do this.

src/taipy/core/_core.py
```diff
@@ -96,6 +97,9 @@ def __manage_version(self, mode, _version_number, force, clean_entities):
 
             version_setter[mode](current_version_number, force)
 
+            if mode == "production":
+                Config._check_one(_MigrationConfigChecker)
```
jrobinAV: I believe the implementation is too complex. 
1. First, we don't need a generic framework for running one checker and we don't need to delegate this particular checker to the Config. It is just a special case that does not require a generic API on the config side. 
Why not just run:
```
_MigrationConfigChecker(Config._applied_config, IssueCollector())._check()
if len(collector._errors) != 0:
    raise SystemExit("Configuration errors found. Please check the error log for more information.")
```
I guess that would make the PR on Config useless. Right?

2. Second, I also believe that does not fix the problem. If the migration config is incorrect, the version is already created. I believe the code above should be moved up, right after having retrieved the current_version_number, and before setting the version.

trgiangdo: 1. There are 2 reasons for the implementation on Config. 
First, we also need to log the messages. This can be error if the function is not callable, or `_check_if_entity_property_key_used_is_predefined()` found an error, or, of course, not a production version.
Second, this conveniently collects all Issue into the same IssueCollector on Config singleton, which follow the design of a IssueCollector.

2. Oh no, actually.
So at this point in the process, all other Configs except from the MigrationConfig should be valid.
So if there is any error, it should come from the migration function. And the user can edit the MigrationConfig since the _ConfigComparator ignores the MigrationConfig, which defined in the `core/config/__init__.py`. My reason for this is that we don't want to keep track of the migration functions between versions, since it does not belong to any version.

jrobinAV: 1. There is no need to reuse the same collector. The "normal" checker already passed without any error. Otherwise, the checker would have raised an error. So we just need to run the _MigrationConfigChecker with a new collector to check for potential new errors or new info. 

2. I don't get the point.

jrobinAV (dismissed):
I would not implement a _check_one(Checker) API just for a special case.

src/taipy/core/_core.py
trgiangdo: 1. Besides from the collector, we will need a different method to log the message as well.
However, I believe this can be done without changing the Config.check() API.

2. I mean the ConfigMigrationChecker need the version to be created first. Then, that version will not be a corrupted version since the user can update the migration config without causing any conflict on the ConfigComparator.

trgiangdo:
On the latest commit, the `_MigrationConfigChecker` is handled separately without using the `Config.check()`.

However, we can not check for the version before setting the version up, since that will require us to pass a parameter into the `.check()` method, and the abstract `_Checker.check()` method does not have any parameter.
I believe this will not be a major issue.

Let me know what you think @jrobinAV. Thank you.

src/taipy/core/_core.py
```diff
@@ -96,6 +98,11 @@ def __manage_version(self, mode, _version_number, force, clean_entities):
 
             version_setter[mode](current_version_number, force)
 
+            if mode == "production":
+                collector = _MigrationConfigChecker(Config._applied_config, IssueCollector())._check()
```
jrobinAV: I would prefer to log the migrationConfig issues here. See remarks on [Config PR 56](https://github.com/Avaiga/taipy-config/pull/56). What do you think?

src/taipy/core/config/migration_config.py
```diff
@@ -0,0 +1,188 @@
...
+            **properties (Dict[str, Any]): A keyworded variable length list of additional arguments.
+        Returns:
+            `MigrationConfig^`: The Migration configuration.
+        """
+        return MigrationConfig.__add_migration_function(target_version, pipeline_config, migration_fct, **properties)
+
+    @staticmethod
+    def _add_scenario_migration_function(
+        target_version: str,
+        scenario_config: Union[ScenarioConfig, ScenarioId],
```
jrobinAV:
```suggestion
        scenario_config: Union[ScenarioConfig, str],
```

ScenarioId is an alias for the id of a scenario. not a section id (ScenarioConfigId).

src/taipy/core/config/migration_config.py
```diff
@@ -0,0 +1,188 @@
...
+            **properties (Dict[str, Any]): A keyworded variable length list of additional arguments.
+        Returns:
+            `MigrationConfig^`: The Migration configuration.
+        """
+        return MigrationConfig.__add_migration_function(target_version, task_config, migration_fct, **properties)
+
+    @staticmethod
+    def _add_pipeline_migration_function(
+        target_version: str,
+        pipeline_config: Union[PipelineConfig, PipelineId],
```
jrobinAV:
```suggestion
        pipeline_config: Union[PipelineConfig, str],
```

src/taipy/core/config/migration_config.py
```diff
@@ -0,0 +1,188 @@
...
+            **properties (Dict[str, Any]): A keyworded variable length list of additional arguments.
+        Returns:
+            `MigrationConfig^`: The Migration configuration.
+        """
+        return MigrationConfig.__add_migration_function(target_version, data_node_config, migration_fct, **properties)
+
+    @staticmethod
+    def _add_task_migration_function(
+        target_version: str,
+        task_config: Union[TaskConfig, TaskId],
```
jrobinAV:
```suggestion
        task_config: Union[TaskConfig, str],
```

src/taipy/core/config/migration_config.py
```diff
@@ -0,0 +1,188 @@
...
+
+        migration_fcts = as_dict.pop(self._MIGRATION_FCTS_KEY)
+        deep_update(self.migration_fcts, migration_fcts)
+
+        self._properties.update(as_dict)
+
+    @staticmethod
+    def _add_data_node_migration_function(
+        target_version: str,
+        data_node_config: Union[DataNodeConfig, DataNodeId],
```
jrobinAV:
```suggestion
        data_node_config: Union[DataNodeConfig, str],
```

src/taipy/core/config/migration_config.py
```diff
@@ -0,0 +1,188 @@
...
+            **properties (Dict[str, Any]): A keyworded variable length list of additional arguments.
+        Returns:
+            `MigrationConfig^`: The Migration configuration.
+        """
+        return MigrationConfig.__add_migration_function(target_version, scenario_config, migration_fct, **properties)
+
+    @staticmethod
+    def __add_migration_function(
+        target_version: str,
+        config: Any,
```
jrobinAV: Should `Any` be replaced by `Union[Section, str]`

jrobinAV:
> On the latest commit, the `_MigrationConfigChecker` is handled separately without using the `Config.check()`.
> 
> However, we can not check for the version before setting the version up, since that will require us to pass a parameter into the `.check()` method, and the abstract `_Checker.check()` method does not have any parameter. I believe this will not be a major issue.
> 
> Let me know what you think @jrobinAV. Thank you.

I believe we are good. We can just call the `MigrationConfig.check()` method after setting up the version as you did. Independently from the global `Config.Check()` that is called before.

I also believe the logging should be the responsibility of the core [See Config PR 56](https://github.com/Avaiga/taipy-config/pull/56).

Just minor remarks otherwise.

trgiangdo:
One thing to note about the "add_..._migration_function" APIs is that they all do the same thing, get a Section (or the config_id) and a function to run on it.

There is no separation between datanode, task, etc.

Do you think we should provide only 1 API? @jrobinAV 

Sorry I forgot to mention this before 🥲

jrobinAV:
> One thing to note about the "add_..._migration_function" APIs is that they all do the same thing, get a Section (or the config_id) and a function to run on it.
> 
> There is no separation between datanode, task, etc.
> 
> Do you think we should provide only 1 API? @jrobinAV
> 
> Sorry I forgot to mention this before 🥲

Oh, yes. You are right. We probably need only one API...

jrobinAV (approved):

jrobinAV (approved):

Issue #451, opened 2023-01-27
Change default read/write params for generic Data Node to empty lists


FlorianJacta:
**Description**
This will simplify the construction of the Generic Data Node.  The below creates an error because _read_fct_params_ is None by default when it should be a list. Making the default _read_fct_params_/_write_fct_params_ as an empty list would avoid any error.

```python
import pandas as pd 
from taipy.config import Config
import taipy as tp


def read_data():
    return "test"

def f(re):
    return re

generic_data_node_cfg = Config.configure_generic_data_node(id="my_data_node",
                                                           read_fct=read_data)
output_cfg = Config.configure_data_node(id="output", storage_type="pickle")
t_cfg = Config.configure_task(id="t",
                              input=generic_data_node_cfg,
                              output=output_cfg,
                              function=f)
pipeline_cfg = Config.configure_pipeline(id="p", task_configs=[t_cfg])


tp.Core().run()
pip= tp.create_pipeline(pipeline_cfg)
pip.submit()
print(pip.output.read())
```

Error message:
```
[2023-01-27 15:48:51,940][Taipy][ERROR]  1 errors occurred during execution of job JOB_t\data_node.py", line 306, in read_or_raise
    return self._read()
  File "C:\Users\jacta\AppData\Local\R-MINI~1\envs\dev\lib\site-packages\taipy\core\data\generic.py", line 106, in _read
    return read_fct(*self.properties[self._READ_FUNCTION_PARAMS_PROPERTY])
TypeError: __main__.read_data() argument after * must be an iterable, not NoneType      
```

toan-quach:
@FlorianJacta @jrobinAV 
I just ran the code again and noticed that it failed at `generic_data_node_cfg = Config.configure_generic_data_node(id="my_data_node", read_fct=read_data)` with missing `write_fct` parameter, I'll add the error message below. This is not related to the issue that you raised for this ticket but I believe it's still crucial to make sure we have the same output or understanding of it! Can you confirm this problem? Or let me know which version you're running so I have a look. Thanks! Also can we confirm if this is a desired behavior or not? Or should the `write_fct` and `read_fct` be optional parameters?

Error message:
```
generic_data_node_cfg = Config.configure_generic_data_node(id="my_data_node", read_fct=read_data)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
TypeError: _configure_generic() missing 1 required positional argument: 'write_fct'
```

trgiangdo:
[...]

PR #493, opened 2023-03-01
feature/#451 added default value for read and write fct in generic dn config

toan-quach:
#451

### DIFF
--- a/src/taipy/core/config/checkers/_data_node_config_checker.py
+++ b/src/taipy/core/config/checkers/_data_node_config_checker.py
@@ -32,7 +32,7 @@ class _DataNodeConfigChecker(_ConfigChecker):
             self._check_scope(data_node_config_id, data_node_config)
             self._check_required_properties(data_node_config_id, data_node_config)
             self._check_callable(data_node_config_id, data_node_config)
-            self._check_generic_read_write_fct_and_params(data_node_config_id, data_node_config)
+            self._check_generic_read_write_fct_and_args(data_node_config_id, data_node_config)
             self._check_exposed_type(data_node_config_id, data_node_config)
         return self._collector
 
@@ -94,21 +94,21 @@ class _DataNodeConfigChecker(_ConfigChecker):
                                 f"property `{required_property}` for type `{storage_type}`.",
                             )
 
-    def _check_generic_read_write_fct_and_params(self, data_node_config_id: str, data_node_config: DataNodeConfig):
+    def _check_generic_read_write_fct_and_args(self, data_node_config_id: str, data_node_config: DataNodeConfig):
         if data_node_config.storage_type == DataNodeConfig._STORAGE_TYPE_VALUE_GENERIC:
             properties_to_check = [
-                DataNodeConfig._OPTIONAL_READ_FUNCTION_PARAMS_GENERIC_PROPERTY,
-                DataNodeConfig._OPTIONAL_WRITE_FUNCTION_PARAMS_GENERIC_PROPERTY,
+                DataNodeConfig._OPTIONAL_READ_FUNCTION_ARGS_GENERIC_PROPERTY,
+                DataNodeConfig._OPTIONAL_WRITE_FUNCTION_ARGS_GENERIC_PROPERTY,
             ]
             for prop_key in properties_to_check:
                 if data_node_config.properties and prop_key in data_node_config.properties:
                     prop_value = data_node_config.properties[prop_key]
-                    if not isinstance(prop_value, list):
+                    if not isinstance(prop_value, (list, tuple)):
                         self._error(
                             prop_key,
                             prop_value,
                             f"`{prop_key}` field of DataNodeConfig"
-                            f" `{data_node_config_id}` must be populated with a List value.",
+                            f" `{data_node_config_id}` must be populated with a List or a Tuple value.",
                         )
             if data_node_config_id != DataNodeConfig._DEFAULT_KEY:
                 properties_to_check_at_least_one = [
--- a/src/taipy/core/config/checkers/_data_node_config_checker.py
+++ b/src/taipy/core/config/checkers/_data_node_config_checker.py
@@ -103,12 +103,12 @@ class _DataNodeConfigChecker(_ConfigChecker):
             for prop_key in properties_to_check:
                 if data_node_config.properties and prop_key in data_node_config.properties:
                     prop_value = data_node_config.properties[prop_key]
-                    if not isinstance(prop_value, (list, tuple)):
+                    if not isinstance(prop_value, list):
                         self._error(
                             prop_key,
                             prop_value,
                             f"`{prop_key}` field of DataNodeConfig"
-                            f" `{data_node_config_id}` must be populated with a List or a Tuple value.",
+                            f" `{data_node_config_id}` must be populated with a List value.",
                         )
             if data_node_config_id != DataNodeConfig._DEFAULT_KEY:
                 properties_to_check_at_least_one = [
--- a/src/taipy/core/config/checkers/_data_node_config_checker.py
+++ b/src/taipy/core/config/checkers/_data_node_config_checker.py
@@ -32,7 +32,7 @@ class _DataNodeConfigChecker(_ConfigChecker):
             self._check_scope(data_node_config_id, data_node_config)
             self._check_required_properties(data_node_config_id, data_node_config)
             self._check_callable(data_node_config_id, data_node_config)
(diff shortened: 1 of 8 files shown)


src/taipy/core/config/checkers/_data_node_config_checker.py
```diff
@@ -94,28 +94,45 @@ def _check_required_properties(self, data_node_config_id: str, data_node_config:
...
-                DataNodeConfig._OPTIONAL_READ_FUNCTION_PARAMS_GENERIC_PROPERTY,
-                DataNodeConfig._OPTIONAL_WRITE_FUNCTION_PARAMS_GENERIC_PROPERTY,
+                DataNodeConfig._OPTIONAL_READ_FUNCTION_ARGS_GENERIC_PROPERTY,
+                DataNodeConfig._OPTIONAL_WRITE_FUNCTION_ARGS_GENERIC_PROPERTY,
             ]
             for prop_key in properties_to_check:
                 if data_node_config.properties and prop_key in data_node_config.properties:
                     prop_value = data_node_config.properties[prop_key]
-                    if not isinstance(prop_value, list):
+                    if not isinstance(prop_value, (list, tuple)):
```
jrobinAV: On the documentation, it is said we support the list only. Is there a specific reason to add a tuple here?

jrobinAV (dismissed):

src/taipy/core/config/checkers/_data_node_config_checker.py
toan-quach: no particular reason, I just think it provides user with more option for input type :D if it doesn't make sense to you, we can move it back to list only :D

jrobinAV: Not a strong opinion on it. Keeping a single way (only list) makes it easier to document, but I am fine with keeping tuples as well. It s up to you.

toan-quach: Hmm I agree and I think it's something that developers would also expect as a default behavior also. I'll revert it back to only list then. Thanks!

src/taipy/core/config/checkers/_data_node_config_checker.py
```diff
@@ -110,12 +110,29 @@ def _check_generic_read_write_fct_params(self, data_node_config_id: str, data_no
...
+                    DataNodeConfig._OPTIONAL_READ_FUNCTION_GENERIC_PROPERTY,
+                    DataNodeConfig._OPTIONAL_WRITE_FUNCTION_GENERIC_PROPERTY,
+                ]
+                has_at_least_one = False
+                for prop_key in properties_to_check_at_least_one:
+                    if data_node_config.properties and prop_key in data_node_config.properties:
+                        has_at_least_one = True
+                if not has_at_least_one:
+                    self._error(
+                        ", ".join(properties_to_check_at_least_one),
```
trgiangdo, line 124: Is this a proper way to assign multiple values to an Issue object?

If yes, let me know because I have the same problem on one of my PR as well. This would be great.

src/taipy/core/data/generic.py
```diff
@@ -101,15 +111,19 @@ def storage_type(cls) -> str:
...
 
     def _write(self, data: Any):
-        if write_fct := self.properties[self._REQUIRED_WRITE_FUNCTION_PROPERTY]:
-            if self._WRITE_FUNCTION_PARAMS_PROPERTY in self.properties.keys():
-                return write_fct(data, *self.properties[self._WRITE_FUNCTION_PARAMS_PROPERTY])
+        if write_fct := self.properties[self._OPTIONAL_WRITE_FUNCTION_PROPERTY]:
+            if write_fct_args := self.properties.get(self._WRITE_FUNCTION_ARGS_PROPERTY, None):
+                if not isinstance(write_fct_args, list):
+                    return write_fct(data, *[write_fct_args])
+                return write_fct(data, *write_fct_args)
```
trgiangdo, line 127: Since we require the read/write args to be a list, is this necessary anymore?

src/taipy/core/data/generic.py
```diff
@@ -101,15 +111,19 @@ def storage_type(cls) -> str:
...
 
     def _read(self):
-        if read_fct := self.properties[self._REQUIRED_READ_FUNCTION_PROPERTY]:
-            if self._READ_FUNCTION_PARAMS_PROPERTY in self.properties.keys():
-                return read_fct(*self.properties[self._READ_FUNCTION_PARAMS_PROPERTY])
+        if read_fct := self.properties[self._OPTIONAL_READ_FUNCTION_PROPERTY]:
+            if read_fct_args := self.properties.get(self._READ_FUNCTION_ARGS_PROPERTY, None):
+                if not isinstance(read_fct_args, list):
+                    return read_fct(*[read_fct_args])
+                return read_fct(*read_fct_args)
```
trgiangdo, line 118: Since we require the read/write args to be a list, is this necessary anymore?

tests/core/config/test_data_node_config.py
```diff
@@ -60,8 +60,16 @@ def test_data_node_config_check(caplog):
     with pytest.raises(TypeError):
         Config.configure_data_node("data_nodes", storage_type="sql")
 
-    with pytest.raises(TypeError):
+    # with pytest.raises(TypeError):
```
trgiangdo:
```suggestion
```

tests/core/data/test_generic_data_node.py
```diff
@@ -139,28 +168,52 @@ def test_read_write_generic_datanode(self):
...
+
+    def test_read_write_generic_datanode_with_non_list_arguments(self):
+        generic_dn = GenericDataNode(
+            "foo",
+            Scope.PIPELINE,
+            properties={
+                "read_fct": read_fct_with_args,
+                "write_fct": write_fct_with_args,
+                "read_fct_args": 1,
+                "write_fct_args": 2,
```
trgiangdo, line 201: Again, since we check if the read/write args is a List or not in the Config, this should not be able to happen (from the user perspective) right?

trgiangdo (commented):
This is a great improvement.

I only have some minor remark on the read/write args.

src/taipy/core/config/checkers/_data_node_config_checker.py
toan-quach: I think so? It makes sense to me, is it the right way? I don't know :D

src/taipy/core/data/generic.py
toan-quach: yep still do, I'm assuming you're talking about line 125 and 126, we require args to be list and we enforce it in dataconfigchecker, but just in case somehow some ways, the generic dn was created without the use of config, it will still work. It's another layer to avoid system exit error :D

tests/core/data/test_generic_data_node.py
toan-quach: Yes you're right. The test and the list wrapper as said earlier, is just another layer to avoid system exit error. We alr have checker that will log error about this. This is just a backup of backup :D

src/taipy/core/data/generic.py
toan-quach: same answer as above :D

src/taipy/core/config/checkers/_data_node_config_checker.py
```diff
@@ -110,12 +110,29 @@ def _check_generic_read_write_fct_params(self, data_node_config_id: str, data_no
...
+                        "",
```
jrobinAV:
```suggestion
                        None,
```

jrobinAV: WHy don't you use None instead of empty string?

src/taipy/core/config/checkers/_data_node_config_checker.py
jrobinAV: The case where multiple properties were involved in an issue is not really handled by the Config Issue framework. Actually, I would create two errors, one for each property. But your solution looks fine to me as well.

src/taipy/core/data/generic.py
jrobinAV: FYI, it is important to keep the if condition since the data node entity may have been modified after its creation. That means the entity could be wrong even if the config is correct.

jrobinAV (dismissed):

src/taipy/core/config/checkers/_data_node_config_checker.py
toan-quach: Oh right my bad 😅

jrobinAV (approved):

Issue #508, opened 2023-03-07
BUG-Configuration does not recogize 'modin' as an exposed_type


FlorianJacta:
**Description**
Configuration checker does not recogize 'modin' as an exposed_type.

Error:
```[2023-03-07 17:30:40,889][Taipy][ERROR] Issue(level='ERROR', field='exposed_type', value='modin', message='The `exposed_type` of the DataNodeConfig `raw_crime` must be either "pandas", "modin", "numpy" or a custom type.', tag='_DataNodeConfigChecker')```

**How to reproduce**

- Create a DataNode with modin as exposed type
- Try to run it with a task and scenario

Possible fix:
In  _core/config/data_node_config.py_:

Add line 76:
`   
 _EXPOSED_TYPE_MODIN = "modin"
`

Add in list the _EXPOSED_TYPE_MODIN :
```
    _ALL_EXPOSED_TYPES = [
        _EXPOSED_TYPE_PANDAS,
        _EXPOSED_TYPE_NUMPY,
        _EXPOSED_TYPE_MODIN
    ]
```

**Expected behaviour**

We should get no error with this exposed type, and we should be able to submit a scenario and read and write a Data Node with this exposed type.

**Runtime environment**
Please specify relevant indications.
 - Taipy version: 2.1.0

PR #509, opened 2023-03-08
fix modin exposed type checked

jrobinAV:
(cherry picked from commit db774f569fda9d5caa99118b7293842f07924c43)

We forgot modin as an accepted exposed type value in the data_node_config. The result is that the config checker throw an error when using modin.

https://github.com/Avaiga/taipy-core/issues/508


toan-quach (dismissed):
LGTM!!!

toan-quach (approved):

PR #510, opened 2023-03-08
bug/#494 A scenario submission can execute multiple times the same task

toan-quach:


src/taipy/core/common/_submittable.py
```diff
@@ -49,13 +49,17 @@ def unsubscribe(self, callback: Callable[[_Submittable, Job], None], params: Opt
         raise NotImplementedError
 
     @abc.abstractmethod
-    def _get_tasks(self) -> Dict[str, Task]:
+    def _get_tasks(self):
+        raise NotImplementedError
```
jrobinAV:
```suggestion
```

src/taipy/core/common/_submittable.py
```diff
@@ -49,13 +49,17 @@ def unsubscribe(self, callback: Callable[[_Submittable, Job], None], params: Opt
...
+
+    @abc.abstractmethod
+    def _get_set_tasks(self) -> Set[Task]:
```
jrobinAV: I would name it `_get_set_of_tasks` or `_get_tasks_set`

src/taipy/core/pipeline/pipeline.py
```diff
@@ -169,7 +169,7 @@ def _is_consistent(self) -> bool:
             is_data_node = not is_data_node
         return True
 
-    def _get_tasks(self) -> Dict[str, Task]:
+    def _get_tasks(self):
```
jrobinAV: There is no need to inherite this method.
```suggestion
    def _get_tasks(self)) -> Dict[str, Task]:
```

src/taipy/core/scenario/scenario.py
```diff
@@ -135,19 +135,26 @@ def pipelines(self, pipelines: Union[List[PipelineId], List[Pipeline]]):
 
     @property
     def tasks(self) -> Dict[str, Task]:
```
jrobinAV:
```suggestion
    def tasks(self) -> Dict[str, List[Task]]:
```
Since there is no auto reload on the `tasks` property, we can just remove _get_tasks. No?

src/taipy/core/scenario/scenario.py
```diff
@@ -135,19 +135,26 @@ def pipelines(self, pipelines: Union[List[PipelineId], List[Pipeline]]):
...
-                if not isinstance(t, Task):
-                    raise NonExistingTask(v)
-                tasks[k] = v
+        list_dict_tasks = [pipeline.tasks for pipeline in self.pipelines.values()]
+        for dict_task in list_dict_tasks:
+            for task_config_id, task in dict_task.items():
+                if task_config_id in tasks.keys() and task not in tasks[task_config_id]:
+                    tasks[task_config_id].append(task)
+                else:
+                    tasks[task_config_id] = [task]
```
jrobinAV, line 145: _

src/taipy/core/scenario/scenario.py
```diff
@@ -135,19 +135,26 @@ def pipelines(self, pipelines: Union[List[PipelineId], List[Pipeline]]):
 
     @property
     def tasks(self) -> Dict[str, Task]:
-        task_manager = _TaskManagerFactory._build_manager()
+        return self._get_tasks()
+
+    def _get_tasks(self):
```
jrobinAV:
```suggestion
    def _get_tasks(self) -> Dict[str, List[Task]]:
```

tests/core/pipeline/test_pipeline.py
```diff
@@ -130,6 +130,22 @@ class FakeDataNode:
 
 
 def test_get_sorted_tasks():
+    def assert_two_sorted_list_of_tasks(tasks_a, tasks_b) -> bool:
```
jrobinAV:
```suggestion
    def assert_equal(tasks_a, tasks_b) -> bool:
```

tests/core/scenario/test_scenario.py
```diff
@@ -420,3 +420,292 @@ def test_get_inputs():
...
+
+    task_5 = Task("waldo", {}, print, id=TaskId("t5"))
+    pipeline_3 = Pipeline("dlugh", {}, [task_1, task_2, task_3], PipelineId("p3"))
+    pipeline_4 = Pipeline("xyzzyx", {}, [task_5, task_4], PipelineId("p4"))
+    scenario_2 = Scenario("scenario_2", [pipeline_3, pipeline_4], {}, ScenarioId("s2"))
+    assert scenario_2._get_set_tasks() == {task_1, task_2, task_3, task_4, task_5}
+
+
+def test_get_sorted_tasks():
+    def assert_two_sorted_list_of_tasks(tasks_a, tasks_b) -> bool:
```
jrobinAV:
```suggestion
    def assert_equal(tasks_a, tasks_b) -> bool:
```

tests/core/scenario/test_scenario.py
```diff
@@ -420,3 +420,292 @@ def test_get_inputs():
...
+    #  p1  s1 ---
+    #            |
+    #            |---> t1 ---|      -----> t3
+    #            |           |      |
+    #      s2 ---             ---> s4 ---> t4 ---> s7
+    #  p2  t2 ---> s4 ---> t3
+    #  p3  s8 ---> t5 ---> s4 ---> t4 ---> s7
+    #  p4  s4 ---> t4 ---> s7
+    print(scenario._get_sorted_tasks())
+    print([[task_5, task_2, task_1, task_4], [task_3]])
```
jrobinAV:
```suggestion
```

tests/core/scenario/test_scenario.py
```diff
@@ -420,3 +420,292 @@ def test_get_inputs():
...
+    # assert assert_two_sorted_list_of_tasks(scenario._get_sorted_tasks(), [[task_5, task_2, task_1, task_4], [task_3]])
```
jrobinAV: Should we uncomment the line? Do you know why it does not work?

src/taipy/core/common/_submittable.py
toan-quach: Not sure why you'd want to delete this but okay :D

src/taipy/core/scenario/scenario.py
toan-quach: I think for consistency, why not keep it?

src/taipy/core/scenario/scenario.py
toan-quach: Do we need to check `task not in tasks[task_config_id]`? Because, if not, there can be the same task in the list of tasks because the same task can belong to multiple pipeline

tests/core/scenario/test_scenario.py
toan-quach: Not yet unfortunately :( I'm still investigating

toan-quach:
I added several unit tests for submitting pipeline and scenario in `test_scheduler.py`. Though most of them might not be so helpful, as they're testing the same thing in different forms. And the drawback of this is it prolongs the time of each pytest run (which may or may not be expensive on GitHub action). If you guys think the tests are not necessary, I can remove them 😃

src/taipy/core/scenario/scenario.py
```diff
@@ -134,20 +134,27 @@ def pipelines(self, pipelines: Union[List[PipelineId], List[Pipeline]]):
...
-        for task in list_tasks:
-            for k, v in task.items():
-                t = task_manager._get(v, v)
-                if not isinstance(t, Task):
-                    raise NonExistingTask(v)
-                tasks[k] = v
+    def tasks(self) -> Dict[str, List[Task]]:
+        return self._get_tasks()
+
+    def _get_tasks(self) -> Dict[str, List[Task]]:
```
jrobinAV: I believe we should remove `_get_tasks()`, which is purely internal and does the same thing as `tasks()`. For me, it s just adding confusion to keep them both.

src/taipy/core/scenario/scenario.py
jrobinAV: You are right. We should have something like that: 

```
if task_config_id in tasks: 
    if task not in tasks[task_config_id]:
        tasks[task_config_id].append(task)
else:
    tasks[task_config_id] = [task]
```

By the way, the function could return a `Dict[str, Set[Task]]`. Am I right?

tests/core/_scheduler/test_scheduler.py
```diff
@@ -93,7 +95,7 @@ def test_submit_task():
     assert job.is_completed()
 
 
-def test_submit_pipeline_generate_unique_submit_id(pipeline, task):
+def test_submit_pipeline_and_scenario_generate_unique_submit_id(pipeline, task):
```
jrobinAV:
```suggestion
def test_submit_pipeline_generate_unique_submit_id(pipeline, task):
```

jrobinAV (commented):
A lot of UT. That s good!

A few comments otherwise.

src/taipy/core/scenario/scenario.py
toan-quach: sure that's good with me too

toan-quach: to answer your point below "By the way, the function could return a Dict[str, Set[Task]]. Am I right?" => It will return a list that doesn't have duplicated tasks. Is it a set -> not necessarily, it's still a list object just without duplication, its function and attributes are still that of a list to keep it consistent with what we have in pipeline.py

src/taipy/core/scenario/scenario.py
toan-quach: Yes and no 😃 we are also checking if the Task is already existing in the dictionary[task_config_id], if yes, we won't append the Task => no duplication, this behavior is similar to that of a Set object. However, the returning value is still a List, just with no duplication

jrobinAV: Let's imagine the following scenario: 
```
A -> t1 ---> B -> t2  -> C -> t3 -> D
             |
              --> t2' -> E
```
with 3 pipelines in this order (t1, t2), (t1, t2'), and (t2, t3). 
t2 and t2' are two different tasks, but they have the same config_id.

Here is the value of 
```
list_dict_tasks = [
{cfg_1: t1, cfg_2: t2}
{cfg_1: t1, cfg_2: t2'}
{cfg_2: t2, cfg_3: t3}
]
```

Here are the successive values of tasks.

tasks = {cfg_1: [t1]}
tasks = {cfg_1: [t1], cfg_2: [t2]}
tasks = {cfg_1: [t1], cfg_2: [t2]}
tasks = {cfg_1: [t1], cfg_2: [t2, t2']}
tasks = {cfg_1: [t1], cfg_2: [t2]}
tasks = {cfg_1: [t1], cfg_2: [t2], cfg_3: [t3]}

At the end t2' is missing.

Can you implement such test to be sure?

I believe the following solves the issue:
```suggestion
if task_config_id in tasks: 
    if task not in tasks[task_config_id]:
        tasks[task_config_id].append(task)
else:
    tasks[task_config_id] = [task]
```

jrobinAV (approved):

Issue #497, opened 2023-03-06
FSRepository: Extract the business logic from repository to the Entity or Manager


jrobinAV:
Ensure the business logic from all children repository classes is moved either to the Entity or the Manager.

Issue #496, opened 2023-03-06
FSRepository : Implement chid classes


jrobinAV:
- [ ] Implement all children classes 
- _CycleFSRepository
- _ScenarioFSRepository
- _PipelineFSRepository
- _TaskFSRepository
- _DataNodeFSRepository
- _JobFSRepository
- _VersionFSRepository

- [ ] Ensure all classes are fully unit tested

PR #518, opened 2023-03-15
feature/#497 extract business logic from repo to entity (p1)

toan-quach:
#497 
Only extracting business logic of:
- [x] scenario
- [x] pipeline
- [x] task
- [x] job
- [x] version

Notes:
The lack of new tests is due to what we are expecting from this ticket. The behaviors should all work the same after the refactoring. Hence, the current tests are more than sufficient to make sure that happens!

### DIFF
--- a/src/taipy/core/pipeline/pipeline.py
+++ b/src/taipy/core/pipeline/pipeline.py
@@ -12,6 +12,7 @@
 from __future__ import annotations
 
 import uuid
+from collections import defaultdict
 from typing import Any, Callable, Dict, List, Optional, Set, Union
 
 import networkx as nx
@@ -20,6 +21,7 @@ from taipy.config.common._template_handler import _TemplateHandler as _tpl
 from taipy.config.common._validate_id import _validate_id
 
 from .._version._version_manager_factory import _VersionManagerFactory
+from ..common import _utils
 from ..common._entity import _Entity
 from ..common._listattributes import _ListAttributes
 from ..common._properties import _Properties
@@ -32,6 +34,7 @@ from ..data.data_node import DataNode
 from ..exceptions.exceptions import NonExistingTask
 from ..job.job import Job
 from ..task.task import Task
+from ._pipeline_model import _PipelineModel
 
 
 class Pipeline(_Entity, _Submittable):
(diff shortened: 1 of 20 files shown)


src/taipy/core/pipeline/pipeline.py
```diff
@@ -260,3 +263,45 @@ def submit(
...
+            model.tasks,
+            model.id,
+            model.owner_id,
+            set(model.parent_ids),
+            [
+                _Subscriber(_utils._load_fct(it["fct_module"], it["fct_name"]), it["fct_params"])
+                for it in model.subscribers
+            ],  # type: ignore
+            model.version,
+        )
```
trgiangdo: Why do we remove the `except NonExistingTask:` and `except KeyError:` compare to the logic in _pipeline_repository.py?

src/taipy/core/pipeline/pipeline.py
```diff
@@ -260,3 +263,45 @@ def submit(
         from ._pipeline_manager_factory import _PipelineManagerFactory
 
         return _PipelineManagerFactory._build_manager()._submit(self, callbacks, force, wait, timeout)
+
+    @classmethod
+    def _to_model(cls, pipeline):
```
```suggestion
    def _to_model(cls, pipeline: Pipeline) -> _PipelineModel:
```

src/taipy/core/pipeline/pipeline.py
```diff
@@ -260,3 +263,45 @@ def submit(
...
+            list(pipeline._parent_ids),
+            pipeline.config_id,
+            pipeline._properties.data,
+            cls.__to_task_ids(pipeline._tasks),
+            _utils._fcts_to_dict(pipeline._subscribers),
+            pipeline.version,
+        )
+
+    @classmethod
+    def _from_model(cls, model):
```
```suggestion
    def _from_model(cls, model: _PipelineModel) -> Pipeline:
```

src/taipy/core/scenario/scenario.py
```diff
@@ -362,3 +365,48 @@ def __get_pipelines(self):
                 raise NonExistingPipeline(pipeline_or_id)
             pipelines[p.config_id] = p
         return pipelines
+
+    @classmethod
+    def _to_model(cls, entity):
```
trgiangdo:
```suggestion
    def _to_model(cls, entity: Scenario) -> _ScenarioModel:
```

src/taipy/core/scenario/scenario.py
```diff
@@ -362,3 +365,48 @@ def __get_pipelines(self):
...
+            creation_date=entity._creation_date.isoformat(),
+            primary_scenario=entity._primary_scenario,
+            subscribers=_utils._fcts_to_dict(entity._subscribers),
+            tags=list(entity._tags),
+            version=entity.version,
+            cycle=cls.__to_cycle_id(entity._cycle),
+        )
+
+    @classmethod
+    def _from_model(cls, model):
```
trgiangdo:
```suggestion
    def _from_model(cls, model: _ScenarioModel) -> Scenario:
```

src/taipy/core/task/task.py
```diff
@@ -192,3 +196,49 @@ def submit(
         from ._task_manager_factory import _TaskManagerFactory
 
         _TaskManagerFactory._build_manager()._submit(self, callbacks, force, wait, timeout)
+
+    @classmethod
+    def _to_model(cls, task) -> _TaskModel:
```
trgiangdo, line 201:
```suggestion
    def _to_model(cls, task: Task) -> _TaskModel:
```

src/taipy/core/task/task.py
```diff
@@ -192,3 +196,49 @@ def submit(
...
+            function_name=task._function.__name__,
+            function_module=task._function.__module__,
+            output_ids=cls.__to_ids(task.output.values()),
+            version=task.version,
+            skippable=task._skippable,
+            properties=task._properties.data.copy(),
+        )
+
+    @classmethod
+    def _from_model(cls, model: _TaskModel):
```
trgiangdo, line 217:
```suggestion
    def _from_model(cls, model: _TaskModel) -> Task:
```

trgiangdo (commented):
I have a question and some minor recommendation on type hints. 

Otherwise I think the PR looks good.

src/taipy/core/pipeline/pipeline.py
toan-quach: because the model is storing a list of TaskId, and we're creating Pipeline using TaskId from the model without getting any actual Task, so the exception is not necessary 😃

src/taipy/core/task/task.py
toan-quach: @trgiangdo 
For some reason, type hint doesn't work for `_to_model` and `_from_model` for Task and Datanode, it does for scenario and pipeline. pre-commit is also complaining non-consistent between the 2 functions in _Entity class and actual entity classes. So I decided it's not worth the pain 😃

jrobinAV (approved):
Will this create some conflict with core-https://github.com/Avaiga/taipy-core/issues/496?

PR #519, opened 2023-03-16
feature/#497 extract business logic from repo to entity (p2)

toan-quach:
#497 
Only extracting business logic of:
- [x] datanode

Notes:
The lack of new tests is due to what we are expecting from this ticket. The behaviors should all work the same after the refactoring. Hence, the current tests are more than sufficient to make sure that happens!

### DIFF
--- a/src/taipy/core/data/abstract_sql.py
+++ b/src/taipy/core/data/abstract_sql.py
@@ -232,3 +232,19 @@ class _AbstractSQLDataNode(DataNode):
         if key in self.__ENGINE_PROPERTIES:
             self._engine = None
         return super().__setattr__(key, value)
+
+    def _serialize_datanode_properties(self):
+        properties = super()._serialize_datanode_properties()
+        properties = super()._serialize_exposed_type(
+            properties, self.__EXPOSED_TYPE_PROPERTY, self.__VALID_STRING_EXPOSED_TYPES
+        )
+        return properties
+
+    @classmethod
+    def _deserialize_datanode_properties(cls, data_node_model):
+        properties = super()._deserialize_datanode_properties(data_node_model)
+        properties = super()._deserialize_exposed_type(
+            properties, cls.__EXPOSED_TYPE_PROPERTY, cls.__VALID_STRING_EXPOSED_TYPES
+        )
+
+        return properties
--- a/src/taipy/core/data/data_node.py
+++ b/src/taipy/core/data/data_node.py
@@ -14,6 +14,7 @@ import uuid
 from abc import abstractmethod
 from datetime import datetime, timedelta
 from functools import reduce
+from pydoc import locate
 from typing import Any, List, Optional, Set, Tuple, Union
 
 import modin.pandas as modin_pd
@@ -31,6 +32,7 @@ from ..common._reload import _reload, _self_reload, _self_setter
 from ..common._warnings import _warn_deprecated
 from ..common.alias import DataNodeId, Edit, JobId
 from ..exceptions.exceptions import NoData
+from ._data_model import _DataNodeModel
 from ._filter import _FilterDataNode
 from .operator import JoinOperator, Operator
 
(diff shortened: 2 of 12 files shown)


jrobinAV (approved):

src/taipy/core/data/data_node.py
```diff
@@ -534,3 +536,97 @@ def all_subclasses(cls):
...
+                elif isinstance(properties[exposed_type_key], List):
+                    properties[exposed_type_key] = [
+                        v if v in valid_str_exposed_types else f"{v.__module__}.{v.__qualname__}"
+                        for v in properties[exposed_type_key]
+                    ]
+                else:
+                    properties[exposed_type_key] = (
+                        f"{properties[exposed_type_key].__module__}." f"{properties[exposed_type_key].__qualname__}"
+                    )
+        return properties
```
jrobinAV, line 558: The deserialize_exposed_type method is specific for tabular data nodes only. It should not be in data_node class.

src/taipy/core/data/data_node.py
```diff
@@ -534,3 +536,97 @@ def all_subclasses(cls):
...
+                elif isinstance(properties[exposed_type_key], dict):
+                    properties[exposed_type_key] = {
+                        k: v if v in valid_str_exposed_types else locate(v)
+                        for k, v in properties[exposed_type_key].items()
+                    }
+                elif isinstance(properties[exposed_type_key], List):
+                    properties[exposed_type_key] = [
+                        v if v in valid_str_exposed_types else locate(v) for v in properties[exposed_type_key]
+                    ]
+        return properties
```
jrobinAV, line 575: Same remark as the serialize_exposed_type method.

src/taipy/core/data/abstract_sql.py
```diff
@@ -232,3 +232,19 @@ def __setattr__(self, key: str, value) -> None:
...
+        )
+        return properties
+
+    @classmethod
+    def _deserialize_datanode_properties(cls, data_node_model):
+        properties = super()._deserialize_datanode_properties(data_node_model)
+        properties = super()._deserialize_exposed_type(
+            properties, cls.__EXPOSED_TYPE_PROPERTY, cls.__VALID_STRING_EXPOSED_TYPES
+        )
+
```
jrobinAV, line 249: That could be the opportunity to put these methods in a TabularDataNode class for factorization.

https://app.zenhub.com/workspaces/taipy-613ef328affde6000ff77fb3/issues/gh/avaiga/taipy-core/350

src/taipy/core/data/data_node.py
```diff
@@ -534,3 +536,97 @@ def all_subclasses(cls):
...
+
+    def _serialize_datanode_properties(self) -> dict:
+        properties = self._properties.data.copy()
+        return properties
+
+    @classmethod
+    def _deserialize_datanode_properties(cls, data_node_model) -> dict:
+        properties = data_node_model.data_node_properties.copy()
+        return properties
```
jrobinAV, line 584: I don't see a lot of value in having concrete methods for _serialize_datanode_properties and _deserialize_datanode_properties in the data_node class. It is just one line factorized.

I would make these 2 methods abstract in data node class.

jrobinAV (commented):
Your implementation seems to work, but it reveals the need of factorizing some code for tabular data nodes. 

We can benefit from this current ticket to implement the other ticket https://app.zenhub.com/workspaces/taipy-613ef328affde6000ff77fb3/issues/gh/avaiga/taipy-core/350

What do you think?

src/taipy/core/data/abstract_sql.py
toan-quach: well sure why not :D should we implement it the way we did with _AbstractFileDataNode?

src/taipy/core/data/data_node.py
toan-quach: oh initially, only generic, json and json dn has an implementation of these functions, while csv, parquet, excel don't. So I keep it as a concrete class so I don't have to recreate them in classes that didn't need it. But it's not the case at the end of the PR :D I'll fix this

Issue #482, opened 2023-02-21
Implementation data migration API on production and experiment mode


trgiangdo:
**What would that feature address**

When there is multiple version of a Taipy core application on production environment, when there is major conflict, it is not possible to run (or get) old entities.

***Description of the ideal solution***

1. On experiment mode
When the user wants to update an existing experiment version:
- Without the `--force` option:
  &rarr; Prompt the ConfigComparator output
  &rarr; Prompt a suggestion: “Add a new  experiment version or use the `--force` option to override the Config of the current version”.
- With the `--force` option:
  &rarr; Overwrite the Config.
  &rarr; The `--force` option does not migrate entities:
  &rarr; If the user wants to migrate entities of an experiment, then he/she should delete the version and create a new one with the same name (which is a bit restrictive). This should be clarified on the doc.

There is no support for migrating experiment entities at the moment. Because the purpose of an experiment version is to store a stable version of the application, and if the Config is forced to update, then it is not really stable.
 
2. On production mode
- When the user wants to update an existing production version:
  - Without the `--force` option:
  &rarr; Prompt the ConfigComparator output
  &rarr; Prompt there are migration function(s) that also need to be updated.
  &rarr; Prompt suggestions: “Add a new  production version with the migration function or Use the `--force` option to override the current version, and don’t forget to update the migration functions to this version as well”.
  - With the `--force` option:
  &rarr; Only force update the Config on the current version, not the older ones.
  &rarr; Migrate entities of the older on the fly (when there are conflicts) using the migration method provided by the users.

- When the user pushes a new production version:
  - Ignore the `--force` option.
  - We only need to apply the ConfigComparator with the latest production version before this.
We check if there is any migration function or not; if there is not, then print out a Warning.

***Caveats***
 
With `job_executionmode="standalone"`, we need to make sure that the migration function is applied to entities on subprocess as well.

jrobinAV:
@trgiangdo, Can you set up a meeting this week to discuss the issue and try to unblock it?

PR #521, opened 2023-03-20
Feature/#482 - Data migration API

trgiangdo:
https://github.com/Avaiga/taipy-core/issues/482

In this PR:

- Implement the feature as discussed in the ticket.
- The migration function set by API `Config.add_migration_function(target_version, config, func)` is called when an entity that need to be migrate get accessed.
- The `_reload(MANAGER)` helper is now a method of `Reloader()` singleton class. When inside the `Reloader()` context, the `_reload(MANAGER)` method will be bypassed.


tests/core/version/test_production_version_migration.py
```diff
@@ -0,0 +1,264 @@
...
+def twice(a):
+    return a * 2
+
+
+def triple(a):
+    return a * 3
+
+
+def migrate_pickle_path(dn):
+    dn.data_node_properties["path"] = "bar.pkl"
```
trgiangdo: This usage is not optimal. The user would prefer
```python
dn.path = "bar.pkl"
```

However, since path is not an attribute of the DataNodeModel, it will not work.

src/taipy/core/_version/_utils.py
```diff
@@ -22,3 +25,11 @@ def _version_migration() -> str:
...
         _VersionManager._get_or_create("LEGACY-VERSION", True)
         return "LEGACY-VERSION"
+
+
+def _migrate_model(model):
+    if (latest_version := _VersionManager._get_latest_version()) in _VersionManager._get_production_version():
+        if migration_fcts := MigrationConfig._get_migration_fcts_to_latest(model.version, model.config_id):
+            for fct in migration_fcts:
+                model = fct(model)
+            model.version = latest_version
```
jrobinAV: The model cannot be exposed to the end user, so fct cannot take a model as a parameter.

src/taipy/core/config/migration_config.py
```diff
@@ -186,3 +186,25 @@ def __add_migration_function(
...
+
+        for version in versions_to_migrate:
+            if (
+                migration_fct := Config.unique_sections[MigrationConfig.name]
+                .migration_fcts.get(version, {})
+                .get(config_id)
+            ):
+                migration_fcts_to_latest.append(migration_fct)
+
+        return migration_fcts_to_latest
```
jrobinAV: Nice!

src/taipy/core/data/data_node.py
```diff
@@ -166,7 +166,7 @@ def last_edition_date(self):
         _warn_deprecated("last_edition_date", suggest="last_edit_date")
         return self.last_edit_date
 
-    @last_edition_date.setter  # type: ignore
+    @last_edition_date.setter
```
jrobinAV, line 171: YEs !!!! That's always a victory when removing a "# type: ignore"!

jrobinAV (changes requested):

trgiangdo:
This PR is blocked due to conflicts with the ongoing refactoring of repositories.

src/taipy/core/common/_reload.py
```diff
@@ -10,29 +10,46 @@
...
 
-@functools.lru_cache
-def _get_manager(manager: str):
-    from ..cycle._cycle_manager_factory import _CycleManagerFactory
-    from ..data._data_manager_factory import _DataManagerFactory
-    from ..job._job_manager_factory import _JobManagerFactory
-    from ..pipeline._pipeline_manager_factory import _PipelineManagerFactory
-    from ..scenario._scenario_manager_factory import _ScenarioManagerFactory
-    from ..task._task_manager_factory import _TaskManagerFactory
+class SingletonMeta(type):
```
trgiangdo: I'm not sure if this is a good place to put this class. It's only used here for now, so what do you think?

jrobinAV: I believe we don't even need the parent class. We can just merge both `SingletonMeta` and `Reloader`.

jrobinAV (commented):
I believe the global idea is good.

1- Since `Reload()` is a singleton, I wonder whether it needs to be thread-safe since `_no_reload_context` can be modified concurrently through various UI requests.

2- We may not want to lock all requests if we need to be thread-safe. We may want to lock only requests related o the same entity. That means we may manage one lock per entity.

jrobinAV (approved):

src/taipy/core/_version/_version_manager.py
```diff
@@ -94,17 +94,23 @@ def _get_development_version(cls) -> str:
...
+                f" for this experiment."
             )
 
-        cls._get_or_create(version_number, force)
+        try:
+            cls._get_or_create(version_number, force)
+        except ConflictedConfigurationError:
+            raise SystemExit(
+                f"Please add a new experiment version or run your application with --force option to override"
+                f" the Config of experiment {version_number}."
```
trgiangdo:
```suggestion
                f"Please add a new experiment version or run your application with --force-run option to"
                f" override the Config of experiment {version_number}."
```

src/taipy/core/_version/_version_manager.py
```diff
@@ -119,27 +125,16 @@ def _get_latest_version(cls) -> str:
...
         if version_number == cls._get_development_version():
             cls._set_development_version(str(uuid.uuid4()))
 
-        cls._get_or_create(version_number, force)
+        try:
+            cls._get_or_create(version_number, force)
+        except ConflictedConfigurationError:
+            raise SystemExit(
+                f"Please add a new production version with a migration function or run your application with --force"
+                f" option to override the Config of production version {version_number}."
```
trgiangdo:
```suggestion
                f"Please add a new production version with a migration function or run your application with"
                f" --force-run option to override the Config of production version {version_number}."
```

trgiangdo (commented):
Rename `--force` to `--force-run` in the error message

Issue #87, opened 2022-03-15
BUG-SQL datanode not working with Azure SQL Database


FlorianJacta:
**Description**
The problem raises when trying to connect to a Azure SQL database.
This is the error:
(pyodbc.InterfaceError) ('IM002', '[IM002] [Microsoft][Gestionnaire de pilotes ODBC] Source de données introuvable et nom de pilote non spécifié (0) (SQLDriverConnect)')

**How to correct it**
- Change in SQLDataNode.py line 97 (__build_conn_string):
```
return "mssql+pyodbc:///?odbc_connect=" + urllib.parse.quote_plus(
                f"DRIVER=FreeTDS;SERVER={host};PORT={port};DATABASE={database};UID={username};PWD={password};TDS_Version=8.0;"
            )
```
->
```
return "mssql+pyodbc:///?odbc_connect=" + "DRIVER={ODBC Driver 18 for SQL Server};" + urllib.parse.quote_plus(
                f"SERVER={host};PORT={port};DATABASE={database};UID={username};")+f"PWD={password};TDS_Version=8.0;"
```

- Change the default port from 143 to 1433. so change in taipy.py in line 636 (configure_sql_data_node)
```
db_port: int = 143,
```
->
```
db_port: int = 1433,
```

**How to improve it**
- Make the 'db_host' parameter directly visible for the configure_sql_data_node function in taipy.py.
Line 628-639
```
def configure_sql_data_node(
    id: str,
    db_username: str,
    db_password: str,
    db_name: str,
    db_engine: str,
    read_query: str,
    write_table: str,
    db_port: int = 143,
    scope: Scope = DataNodeConfig._DEFAULT_SCOPE,
    **properties,
):
```
->
```
def configure_sql_data_node(
    id: str,
    db_username: str,
    db_password: str,
    db_name: str,
    db_engine: str,
    read_query: str,
    write_table: str,
    db_port: int = 1433,
    db_host: str = 'localhost',
    scope: Scope = DataNodeConfig._DEFAULT_SCOPE,
    **properties,
):
```

- Make write_table optional if we don't want to write in it

**How to reproduce**
Run the demo 'classification_csv_churn_v2' from 'taipy_0_13' branch present here: https://github.com/Avaiga/demos
An error is raised because of the SQL datanode.

**Expected behavior**
We shouldn't have an error.

**Runtime environment**
Windows 11

PR #527, opened 2023-03-22
Feature/notification

jrobinAV:
This PR is just a POC for the Core part of ticket [taipy #87](https://github.com/Avaiga/taipy/issues/87). It won't be merged as it is.

[Here are the requirements](https://docs.google.com/document/d/19R0SiykrlGALWU9ONDdtmni4PNBmJ-95s9i755FhWgg).

[Here are the specifications](https://docs.google.com/document/d/10GMDUZj0PJ5Tz24nOueA3whgTLHfU8WYvjwHoUv6K1w)

The following picture is the proposed architecture: 

Gui controls for Core - Specifications

Taipy-core should hold the notification system, including the registration and the publish. 
 
When calling Core().run, a Notifier singleton should be instantiated. Other possible names: Event_Channel, Event_Exchange, Event_Notifier, … 

Its purpose is to:
1. Hold in memory the existing registrations

2. Expose to GUI controls the `Notifier.register` and `Notifier.unregister` APIs. 
- The Notifier is not exposed except to taipy repositories. More functionalities will be needed before exposing it. Managing registered clients, limiting registration in time, extending registration, …
- The register method is responsible for
    - Creating and holding a new queue attached to a topic (depending on register parameters)
    - returning the queue.

3. Expose to taipy-core the `Notifier.publish` API takes as a parameter an event.
- Other possible names: emit, notify, produce, …
- The `publish` method is responsible for:
    - mapping the event received with the appropriate existing queues
    - Put the event in all appropriate queues.

An Event object is a simple data class holding the characteristics of the event to publish, so the `publish` method should be able to filter the registrations to notify.

TODO:
- [ ] add lock
- [ ] refactor _registrations in Notifier class to use a dict
- [ ] implement logic to see if event matches the topic
- [ ] generate ids when register
- [ ] add unit tests


src/taipy/core/notification/notifier.py
```diff
@@ -0,0 +1,46 @@
...
+
+    @classmethod
+    def unregister(cls, register_id: str = "gui"):
+        # TODO lock before updating cls._registrations
+        cls._registrations = [reg for reg in cls._registrations if reg.register_id != register_id]
+
+    @classmethod
+    def publish(cls, event: Event):
+        for registration in cls._registrations:
+            if cls.is_matching(event, registration.topic):
```
toan-quach: Should we add a lock when publishing a new event? As multiple threads would be looking at the queue at this point, we don't have to lock the registration but I think we should lock individual queue when we want to update it. (individual queue instead of all to avoid performance downgrade). But I'm not sure if we have 1 thread putting new element and another getting the element, would that create a conflict? If NO, then locking isn't necessary then?

src/taipy/core/notification/notifier.py
```diff
@@ -0,0 +1,46 @@
...
+
+class Notifier:
+    # TODO add a lock on _registration
+    _registrations: List[Registration] = []
+
+    @classmethod
+    def register(cls, entity_type, entity_id, operation, attribute_name, time_to_live) -> Tuple[str, Queue]:
+        register_id = "gui"  # TODO generate an id to return so the client can unregister.
+        registration = Registration(register_id, entity_type, entity_id, operation, attribute_name, time_to_live)
+        # TODO lock before updating cls._registrations
```
toan-quach: I dont think it's necessary to lock here as the registrations should be done before starting a core event consumer thread to get the queue, we won't have any conflict of resources at this point.

src/taipy/core/notification/notifier.py
```diff
@@ -0,0 +1,46 @@
...
+        cls._registrations.append(registration)
+        return register_id, registration.queue
+
+    @classmethod
+    def unregister(cls, register_id: str = "gui"):
+        # TODO lock before updating cls._registrations
+        cls._registrations = [reg for reg in cls._registrations if reg.register_id != register_id]
```
toan-quach: The same with this part, when we unregister a sub, the thread of the consumer should be stop first before attempting to unregister the subscription => no resource conflict

src/taipy/core/notification/event.py
```diff
@@ -0,0 +1,46 @@
...
+from typing import Optional
+
+from ..common._repr_enum import _ReprEnum
+
+
+class EventOperation(_ReprEnum):
+    CREATION = 1
+    UPDATE = 2
+    DELETION = 3
+    # SUBMISSION = 4
```
toan-quach: Why was it commented out? 😃

src/taipy/core/notification/core_event_consumer.py
```diff
@@ -0,0 +1,49 @@
...
+# specific language governing permissions and limitations under the License.
+
+import abc
+import threading
+from queue import Empty, SimpleQueue
+
+from .event import Event
+
+
+class CoreEventConsumer(threading.Thread):
```
toan-quach, line 19: neat! I think this example is really great, so we will only need to retrieve the register_id and the queue from the notifier to instantiate the consumer, which IMO is simple and straightforward to implement 😃 I agree this class can be a parent/abstract class so that we can hide away the thread manipulation part (can be a pain from time to time) and only exposing the process_event (or the business logic)!

toan-quach (commented):
LGTM! Though I do have some concern regarding multiple queues involved when we have multiple registrations, this can be quite resource-intensive especially if we let the application runs for an extensive time. 1 of the way to prevent this as we talked about is to limit the thread size, event time, have a cold storage or have some persistency in place.

src/taipy/core/notification/notifier.py
jrobinAV: Yes, you are right. We need a lock per queue.

src/taipy/core/notification/event.py
jrobinAV: No reason. It s just a draft for now.

src/taipy/core/notification/notifier.py
```diff
@@ -0,0 +1,43 @@
...
+from .registration import Registration
+from .topic import Topic
+
+
+class Notifier:
+    _registrations: List[Registration] = []
+
+    @classmethod
+    def register(cls, entity_type, entity_id, operation, attribute_name) -> Tuple[str, Queue]:
+        register_id = "gui"  # TODO generate an id to return so the client can unregister.
```
toan-quach: Since the user will be using this register_id, it should be simple and straight-forward. I'm thinking of something like all the parameters combined perhaps f"{entity_type}_{entity_id}_{operation}_{attribute_name}", but this is quite long :/ . We can also ask the user to provide a name that they prefer?

src/taipy/core/notification/notifier.py
```diff
@@ -0,0 +1,43 @@
...
+from .event import Event
+from .registration import Registration
+from .topic import Topic
+
+
+class Notifier:
+    _registrations: List[Registration] = []
+
+    @classmethod
+    def register(cls, entity_type, entity_id, operation, attribute_name) -> Tuple[str, Queue]:
```
toan-quach: The parameter is really specific, but what if I want to register an entity with all of its operation? Do I have to do it multiple times? Instead of entity_type, entity_id, I think can ask the user to just provide the entity. What do you think?

src/taipy/core/notification/notifier.py
```diff
@@ -18,20 +18,17 @@
 
 
 class Notifier:
-    # TODO add a lock on _registration
     _registrations: List[Registration] = []
```
toan-quach: a dictionary would be better for adding/removing by id

src/taipy/core/notification/notifier.py
toan-quach: apparently, I think Python has already provided us with a thread-safe queue so we don't need to use any locking mechanism to prevent race condition. Further reading: https://superfastpython.com/thread-queue/

toan-quach: If I understand correctly, then we shouldn't need any lock in the orchestrator as well and it should work fine. But maybe we still need it when we want to queue a list of jobs, just to make sure all of the queuing is done prior to starting any executions

jrobinAV: In the orchestrator, we are using a lock for multiple variables. A queue for the jobs_to_run and a list for the blocked_jobs. They must be synchronized. So I believe the case is different.

toan-quach: Ah you're right :D

src/taipy/core/notification/notifier.py
jrobinAV: I suggest making all the parameters optional. Like:

- SCENARIO, None, None, None -> Get notified for any operation on any scenario
- SCENARIO, None, UPDATE, None -> Get notified for any UPDATE operation on any scenario
- None, ID, UPDATE, DESCRIPTION -> Get notified for any UPDATE operation on attribute DESCRIPTION of any entity with id ID


The drawback is that some combinations do not make sense. 

Maybe instead of accepting four parameters, we can just accept a Topic parameter built by the client (GUI). And maybe having a smarter Topic class or multiple Topic classes could help.

toan-quach: So for a registration, we can have multiple topics yes? And also we must have at least either entity_type or entity_id right?

src/taipy/core/notification/notifier.py
jrobinAV: **Adding more context:** We need more design here because a single client (GUI) with a single `register_id` may register multiple topics. So maybe the name `register_id` is bad. It could be `client_id`, for example.  That also means: 
- the `register` method must behave like adding a new Registration linked to the `register_id`.
- the `unregister` method must remove all the registrations associated with the `register_id`.

**About the `_registrations` variable type:**
I initially thought about a dictionary as well. It should be a `dict[str, List[Registration]]`.
Pros: 
- That makes un-registration easier. 
Cons: 
- That makes the publication much harder. Indeed, the parsing of the registrations become much tricky, and so is the matching of the topic with the event published.

src/taipy/core/notification/notifier.py
jrobinAV: All fields can be None... You may want to register to ALL events, for instance.

src/taipy/core/notification/notifier.py
toan-quach: we will generate a unique id on our side first. We will allow users to provide their own id later on but not in this ticket. (User-provided id must be unique)

src/taipy/core/notification/notifier.py
toan-quach: We will go with a dictionary (key is topics, value is a list of registration object) due to huge benefit in matching topics and events while un-registration for a subscriber is expected to be not so complicated

src/taipy/core/notification/notifier.py
```diff
@@ -10,34 +10,45 @@
...
+        attribute_name: Optional[str],
+    ) -> Registration:
+        registration = Registration(entity_type, entity_id, operation, attribute_name)
+
+        if registrations := cls._registrations.get(registration.topic, None):
+            registrations.append(registration)
+        else:
+            cls._registrations[registration.topic] = [registration]
+
+        return registration
```
jrobinAV: I believe it s easier for the client to receive a queue than a Registration object.

src/taipy/core/notification/registration.py
```diff
@@ -9,12 +9,25 @@
 # an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the
 # specific language governing permissions and limitations under the License.
 from queue import Queue
```
jrobinAV: Using a queue.SimpleQueue should be better. No?

src/taipy/core/notification/event.py
```diff
@@ -34,10 +33,10 @@ class EventEntityType(_ReprEnum):
...
     def __init__(
         self,
-        entity_type: Optional[EventEntityType] = None,
-        entity_id: Optional[str] = None,
-        operation: Optional[EventOperation] = None,
-        attribute_name: Optional[str] = None,
+        entity_type: EventEntityType,
+        entity_id: str,
+        operation: EventOperation,
+        attribute_name: str,
```
jrobinAV: I believe they should remain optional. At least for `attribute_name`. For instance an entity creation or deletion event cannot provide any `attribute_name`

src/taipy/core/notification/topic.py
```diff
@@ -9,10 +9,140 @@
...
+        return (self.entity_type, self.entity_id, self.operation, self.attribute_name)
+
+    @classmethod
+    def __preprocess_entity_type_and_entity_id(cls, entity_type: Optional[str], entity_id: Optional[str]):
+        if entity_type and entity_id:
+            tmp_entity_type = cls.__get_entity_type_from_id(entity_id)
+            if tmp_entity_type != entity_type:
+                raise InvalidEntityType
+            else:
+                entity_type = tmp_entity_type
```
jrobinAV: useless?

src/taipy/core/notification/notifier.py
```diff
@@ -10,34 +10,45 @@
...
+            registrations.remove(registration)
+            if len(registrations) == 0:
+                del cls._registrations[registration.topic]
 
     @classmethod
     def publish(cls, event: Event):
-        for registration in cls._registrations:
-            if cls.is_matching(event, registration.topic):
-                registration.queue.put(event)
+        generated_matched_topics_with_event = Topic.generate_topics_from_event(event)
```
jrobinAV: I would not delegate the matching responsibility to the Topic class. It should remain in Notifier or be in a new class dedicated to comparing events and topics.

src/taipy/core/notification/topic.py
```diff
@@ -9,10 +9,140 @@
...
+    @classmethod
+    def __preprocess_operation(cls, operation):
+        pass
+
+    @classmethod
+    def __preprocess_attribute_name(cls, attribute_name):
+        pass
+
+    @staticmethod
+    def generate_topics_from_event(event: Event):
```
jrobinAV: I don't like the idea of generating topics from events. That looks strange to me.

src/taipy/core/notification/notifier.py
```diff
@@ -10,34 +10,45 @@
...
 
-    @classmethod
-    def is_matching(cls, event: Event, topic: Topic) -> bool:
-        # TODO implement logic to see if event matches the topic
-        return True
+        for topic in generated_matched_topics_with_event:
+            if registrations := cls._registrations.get(topic, None):
+                for registration in registrations:
+                    registration.queue.put(event)
```
jrobinAV: Instead of generating all possible topics, why not implement something like the following:

```
@classmethod
def publish(cls, event):
    for topic, registrations in self.registrations:
        if cls.is_matching(event, topic):
            for registration in registrations:
                registration.queue.put(event)

@classmethod
def is_matching(cls, event: Event, topic: Topic) -> bool:
    if event.entity_type == None or topic.entity_type == None or event.entity_type != topic.entity_type:
        return False
    if event.entity_id == None or topic.entity_id == None or event.entity_id != topic.entity_id :
        return False
    if event.operation == None or topic.operation == None or event.operation != topic.operation:
        return False
    if event.attribute_name == None or topic.attribute_name == None or event.attribute_name != topic.attribute_name:
        return False
    return True
```

I believe it should be easier to maintain and make it evolve. WHat do you think?

src/taipy/core/notification/event.py
toan-quach: Yep you're right :D I just realized the attribute_name should be optional when writing tests! The others don't have a usecase where they will be optional to me though

src/taipy/core/notification/notifier.py
toan-quach: agree, I have made the necessary changes in the new commit, will push later

src/taipy/core/notification/notifier.py
toan-quach: it's certainly easier to maintain and understand. But I think the drawback is that we have to go through all of the topics, this won't be bad if we only have a few topics, but it can be huge if the number of topics is huge (this would be O(N)). While if we generate what topic we should look for from the event, it'd be O(1). It's not necessarily better because we don't have a real use case for this. We might expect the number of topics to be low, so the option you proposed would be better, both performance and readability-wise, but if the number of topics is huge, I think the current implementation would be better. However, I agree that for now, it's better to focus on maintainability rather than trying to optimize it. :D So I'll go with what you proposed

src/taipy/core/notification/topic.py
toan-quach: not sure, we can have a case where the entity id doesn't match the provided entity type, no?

jrobinAV: The `if` block is useful but the `else` is not. 
if `tmp_entity_type != entity_type` means `tmp_entity_type == entity_type`. So we don't need to set `entity_type` to `tmp_entity_type`. Right?

src/taipy/core/notification/notifier.py
jrobinAV: To be honest, I did not though about the performance here. You are right that it will be faster when the nb of topics is big. I also agree with focusing first on readability than performance. We first need to measure the performance before optimizing it. Especially because 
when the nb of topics is small, like today, spending time on generating all topics can compensate for the time saved on searching topics. I am not even sure it is faster for now.

src/taipy/core/notification/topic.py
toan-quach: ohhh my bad you're right, I totally missed that!

src/taipy/core/notification/notifier.py
toan-quach: totally agree with you 😃 let's just focus on readability for now and roll it out, see how it's being used, and if optimization is needed. If we try to optimize it at this point, I think we might over-engineer

src/taipy/core/notification/registration.py
toan-quach: ohh you're right, I changed it without noticing 😅

src/taipy/core/notification/topic.py
toan-quach: as discussed below, we will focus on readability first, and this idea is mainly for optimization of the search for subscribers to publish. So this idea will be scratched 😃

src/taipy/core/notification/topic.py
```diff
@@ -55,24 +45,33 @@ def __preprocess_entity_type_and_entity_id(cls, entity_type: Optional[str], enti
...
         raise InvalidEntityId
 
-    # def __preprocess_attribute_name(self):
-    #     if self.operation == EventOperation.CREATION/DELETION and self.attribute_name is not None:
-    #         throw error?
+    @classmethod
+    def __preprocess_attribute_name(
+        cls, attribute_name: Optional[str] = None, operation: Optional[str] = None
+    ) -> Optional[str]:
+        # TODO: check if attribute_name exists in entity? what if attribute_name but operation is None?
```
jrobinAV: I believe you cannot test if attribute_name exists in the entity. You can have fully custom attributes in the property dict.

src/taipy/core/notification/topic.py
```diff
@@ -55,24 +45,33 @@ def __preprocess_entity_type_and_entity_id(cls, entity_type: Optional[str], enti
...
-            if entity_type and entity_type not in cls.__ENTITY_TYPE_PREFIXES:
+            if entity_type and entity_type not in _ENTITY_TYPE_PREFIXES:
                 raise InvalidEntityType
             if entity_id:
                 entity_type = cls.__get_entity_type_from_id(entity_id)
-
         return entity_type, entity_id
 
     @classmethod
     def __get_entity_type_from_id(cls, entity_id: str):
```
jrobinAV: I would not check the consistency of the entity_id with the entity_type. The client should guarantee the consistency of what he register to. 
This makes the topic know about how an id should be built depending on its type. It s bad design to me. It should not be part of the Topic responsibility.

src/taipy/core/notification/notifier.py
```diff
@@ -27,6 +27,7 @@ def register(
         operation: Optional[str],
         attribute_name: Optional[str],
     ) -> Registration:
+        # TODO: the signature should all parameter be string??
```
jrobinAV: Optional[str] for any attribute looks good to me.

src/taipy/core/notification/event.py
```diff
@@ -0,0 +1,92 @@
...
+
+
+_ENTITY_TYPE_PREFIXES = {
+    Cycle._ID_PREFIX: EventEntityType.CYCLE,
+    Scenario._ID_PREFIX: EventEntityType.SCENARIO,
+    Pipeline._ID_PREFIX: EventEntityType.PIPELINE,
+    Task._ID_PREFIX: EventEntityType.TASK,
+    DataNode._ID_PREFIX: EventEntityType.DATA_NODE,
+    Job._ID_PREFIX: EventEntityType.JOB,
+}
```
jrobinAV: I would not embed this logic in the event class. 
First, I am not sure it is useful.
Second, this prefix logic should remain the responsibility of the ENtity or the entity manager. Not the notifier.

src/taipy/core/notification/event.py
```diff
@@ -0,0 +1,92 @@
...
+        if tmp_entity_type != entity_type:
+            raise InvalidEntityType
+        return entity_id
+
+    @classmethod
+    def __get_entity_type_from_id(cls, entity_id: str) -> EventEntityType:
+        for entity_prefix, event_entity_type in _ENTITY_TYPE_PREFIXES.items():
+            if entity_id.startswith(entity_prefix):
+                return event_entity_type
+        raise InvalidEntityId
```
jrobinAV: I would not preprocess this.
```suggestion
```

src/taipy/core/notification/event.py
```diff
@@ -0,0 +1,92 @@
...
+    def __init__(
+        self,
+        entity_type: EventEntityType,
+        entity_id: str,
+        operation: EventOperation,
+        attribute_name: Optional[str] = None,
+    ):
+        self.creation_date = datetime.now()
+        self.entity_type = entity_type
+        self.entity_id = self.__preprocess_entity_id(entity_id, entity_type)
```
jrobinAV:
```suggestion
        self.entity_id = entity_id
```

src/taipy/core/notification/notifier.py
```diff
@@ -0,0 +1,131 @@
...
+            raise NonExistingRegistration(registration_id)
+
+    @classmethod
+    def publish(cls, event: Event):
+        generated_matched_topics_with_event = cls.generate_topics_from_event(event)
+
+        for topic in generated_matched_topics_with_event:
+            if registrations := cls._topics_subscribers_list.get(topic, None):
+                for registration in registrations:
+                    registration.queue.put(event)
```
jrobinAV: To be changed without generating all topics.

src/taipy/core/notification/registration.py
```diff
@@ -0,0 +1,40 @@
...
+        self,
+        entity_type: Optional[str] = None,
+        entity_id: Optional[str] = None,
+        operation: Optional[str] = None,
+        attribute_name: Optional[str] = None,
+    ):
+
+        self.register_id: str = self._new_id()
+        self.topic: Topic = Topic(entity_type, entity_id, operation, attribute_name)
+        self.queue: SimpleQueue = SimpleQueue()  # TODO: should we allow user to provide their own queue?
```
jrobinAV: I would say no for now, but it could be interesting in the future. We should wait for a concrete use case before implementing it.

src/taipy/core/notification/topic.py
```diff
@@ -0,0 +1,101 @@
...
+        self.operation = self.__preprocess_operation(operation, self.entity_type)
+        self.attribute_name = self.__preprocess_attribute_name(attribute_name, self.operation)
+
+    @classmethod
+    def __preprocess_entity_type_and_entity_id(
+        cls, entity_type: Optional[str], entity_id: Optional[str]
+    ) -> Tuple[EventEntityType, Optional[str]]:
+        try:
+            # TODO: add test for this
+            converted_entity_type = entity_type if not entity_type else EventEntityType.__members__[entity_type.upper()]
```
jrobinAV: This looks bad to me. I am kind of changing my mind. Maybe EntityType and Operation should be exposed to clients as an ENUM. 
That means Topics and Events would share both enums. And the register method signature would become:
```
def register(
        cls,
        entity_type: Optional[EntityType] = None,
        entity_id: Optional[str] = None,
        operation: Optional[Operation] = None,
        attribute_name: Optional[str] = None,
    ) -> Tuple[str, SimpleQueue]:
```

What do you think?

src/taipy/core/notification/topic.py
```diff
@@ -0,0 +1,101 @@
...
+            raise InvalidEntityType
+
+        if converted_entity_type and entity_id:
+            tmp_entity_type = cls.__get_entity_type_from_id(entity_id)
+            if tmp_entity_type != converted_entity_type:
+                raise InvalidEntityType
+            converted_entity_type = tmp_entity_type
+        else:
+            if entity_id:
+                converted_entity_type = cls.__get_entity_type_from_id(entity_id)
```
jrobinAV: I would not check this. Same remark as event preprocessing.

tests/core/notification/test_event.py
```diff
@@ -0,0 +1,278 @@
...
+
+    event_3 = Event(EventEntityType.CYCLE, "CYCLE_cycle_id", EventOperation.DELETION)
+    assert event_3.creation_date is not None
+    assert event_3.entity_type == EventEntityType.CYCLE
+    assert event_3.entity_id == "CYCLE_cycle_id"
+    assert event_3.operation == EventOperation.DELETION
+    assert event_3.attribute_name is None
+
+    with pytest.raises(InvalidEntityId):
+        _ = Event(EventEntityType.CYCLE, "cycle_id", EventOperation.CREATION)
```
jrobinAV: Should not raise an exception for me.

tests/core/notification/test_event.py
```diff
@@ -0,0 +1,278 @@
...
+
+    with pytest.raises(InvalidEntityType):
+        _ = Event(EventEntityType.SCENARIO, "CYCLE_cycle_id", EventOperation.CREATION)
```
jrobinAV: Should not raise an error for me.

tests/core/notification/test_event.py
```diff
@@ -0,0 +1,278 @@
...
+
+    with pytest.raises(InvalidEventAttributeName):
+        _ = Event(EventEntityType.CYCLE, "CYCLE_cycle_id", EventOperation.CREATION, "frequency")
```
jrobinAV: Frequency can be a custom attribute name. So this should not raise an error.

tests/core/notification/test_event.py
```diff
@@ -0,0 +1,278 @@
...
+
+    with pytest.raises(InvalidEventAttributeName):
+        _ = Event(EventEntityType.CYCLE, "CYCLE_cycle_id", EventOperation.DELETION, "frequency")
```
jrobinAV: Should not raise an error for me.

tests/core/notification/test_topic.py
```diff
@@ -0,0 +1,421 @@
...
+    assert topic_6.entity_type == EventEntityType.SCENARIO
+    assert topic_6.entity_id == "SCENARIO_scenario_id"
+    assert topic_6.operation == EventOperation.UPDATE
+    assert topic_6.attribute_name == "properties"
+
+    with pytest.raises(InvalidEntityId):
+        _ = Topic("SCENARIO", "scenario_id", "CREATION")
+
+    with pytest.raises(InvalidEntityType):
+        _ = Topic("CYCLE", "SCENARIO_scenario_id", "CREATION")
```
jrobinAV: It should not raise an error for me.

jrobinAV (commented):
- It s globally good
- A small remark on the preprocessing for events and topics. Especially on the entity_id prefix. 
- Maybe we can rename the Operation and EntityType enums, make them shared by events and topics, and expose them in the register signature
- A lot of unit tests. That is great! It helps a lot in reviewing the PR.

src/taipy/core/notification/topic.py
toan-quach: you're right, I'll remove this note then :D

src/taipy/core/notification/event.py
toan-quach: you're right, I think in the end, it's up to us to make sure we get the right id and entity type, checking it again feels redundant

src/taipy/core/notification/registration.py
toan-quach: The use can I have in mind is where a user want to subscribe to multiple use cases but want to use only 1 queue to manage all of the incoming event for simplicity. The implementation will be tricky as we can have an event published to a queue if they are under valid topics. But yes, we should keep this for future reference and not do anything about it for now.

src/taipy/core/notification/topic.py
toan-quach: It's certainly will be alot easier to implement as we're removing the converting part of str to EventOperation and EventEntityType 😃 not sure if it'll make things more complicated for the user, but it sure keeps things simpler for us

tests/core/notification/test_event.py
toan-quach: well if the operation is create, delete or submission, there should be no attribute names as they don't matter in such cases. We can set the attribute name to None in these cases. What do you think?

tests/core/notification/test_event.py
toan-quach: same remark as above

src/taipy/core/notification/topic.py
toan-quach: Its actually more beneficial for us to enforce the entity type from the entity type, it helps us figure out which operation is available for example submission only goes with scenario, pipeline and task, to know this we must know entity type

src/taipy/core/pipeline/_pipeline_manager.py
```diff
@@ -141,11 +144,14 @@ def _submit(
...
         if check_inputs_are_ready:
             _warn_if_inputs_not_ready(pipeline._get_inputs())
-        return (
+
+        jobs = (
             _TaskManagerFactory._build_manager()
             ._orchestrator()
             .submit(pipeline, callbacks=pipeline_subscription_callback, force=force, wait=wait, timeout=timeout)
         )
+        _publish_event(cls._EVENT_ENTITY_TYPE, pipeline.id, EventOperation.CREATION, None)
```
jrobinAV: The operation should be submission

src/taipy/core/notification/notifier.py
```diff
@@ -0,0 +1,80 @@
...
+
+def _publish_event(
+    entity_type: EventEntityType, entity_id: str, operation: EventOperation, attribute_name: Optional[str] = None
+):
+    Notifier.publish(Event(entity_type, entity_id, operation, attribute_name))
+
+
+class Notifier:
+    _topics_registrations_list: Dict[Topic, List[Registration]] = {}
+    _registrations: Dict[str, Registration] = {}
```
jrobinAV: Why keep both implementations? Only `_topics_registrations_list` is needed. `_registrations` can be removed. No?

toan-quach: No, you're right, I keep _registrations mainly to make it easier to unregister subscribers as it provides topic and registration objects. Without it, we will need to loop through all topics to find the registration to remove. But then again, as many topics we have had in this PR, we don't need it at this stage of the feature.

toan-quach (approved):

Issue #513, opened 2023-03-09
Username, password, folder_path, and extension properties on SQLite Data Node


trgiangdo:
**Description**
- User name and password should be optional when engine="sqlite"
- The `"path"` property should be renamed to `"folder_path"` and documented properly.
- Add a new optional `"extension"` property for SQLite to choose the proper database file extension (.sqlite3, .db, ...)
- db_extra_args should be properly documented in the usermanual.

PR #528, opened 2023-03-22
Feature/#513 - Optional properties on SQLite engine

trgiangdo:
https://github.com/Avaiga/taipy-core/issues/513

In this PR:
- `db_username`, `db_password` are optional for "sqlite" engine, and required for other engines.
- `sqlite_folder_path` and `sqlite_file_extension` are optional for "sqlite" engine.
- Update the docstring of SQL(Table)Datanode Config and entity.


tests/core/data/data_sample/example.sqlite3
joaoandre-avaiga, line 1: I'm not a huge fan of commiting database files to the repository. It's not a blocker for me, but I think it would be more interesting to create a fixture in conftest that would generate the database for testing in a tmp folder.

joaoandre-avaiga (dismissed):

src/taipy/core/data/data_node.py
```diff
@@ -106,15 +106,15 @@ def __init__(
 
         self._properties = _Properties(self, **kwargs)
 
-    @property  # type: ignore
+    @property
```
jrobinAV, line 109: 👍🏻

tests/core/data/data_sample/example.sqlite3
trgiangdo: Yea, I was trying to test if the "folder_path" and "file_extension" properties work.

I'll try for a fixture and a temp file. Thanks :D

tests/core/config/checkers/test_data_node_config_checker.py
```diff
@@ -170,6 +166,55 @@ def test_check_storage_type(self, caplog):
         ]
         assert all(message in caplog.text for message in expected_error_messages)
 
+    def test_check_sql_engine_properties(self, caplog):
```
jrobinAV: We can split the test in two to make it easier to read.

tests/core/data/data_sample/example.sqlite3
jrobinAV: I agree. It would be better but not blocking though.

jrobinAV (dismissed):

tests/core/data/data_sample/example.sqlite3
trgiangdo: Thanks for the suggestion @joaoandre-avaiga. I created 2 new fixtures and it is much cleaner now.

jrobinAV (approved):

Issue #524, opened 2023-03-21
BUG-Unable to read a parquet folder


FlorianJacta:
**Description**
Reading a Parquet folder will raise an error when it should normally work:

```
Traceback (most recent call last):
  File "C:\Users\jacta\OneDrive\Bureau\taipy\Code\taipy-tutorials\main.py", line 33, in <module>
    print(scenario.parquet_folder.read())
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\jacta\.conda\envs\tp_2_1\Lib\site-packages\taipy\core\data\data_node.py", line 316, in read
    return self.read_or_raise()
           ^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\jacta\.conda\envs\tp_2_1\Lib\site-packages\taipy\core\data\data_node.py", line 305, in read_or_raise
    if not self.last_edit_date:
           ^^^^^^^^^^^^^^^^^^^
  File "C:\Users\jacta\.conda\envs\tp_2_1\Lib\site-packages\taipy\core\common\_reload.py", line 60, in _do_reload
    return fct(self, *args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\jacta\.conda\envs\tp_2_1\Lib\site-packages\taipy\core\data\data_node.py", line 153, in last_edit_date
    if last_modified_datetime and last_modified_datetime > self._last_edit_date:
                                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: '>' not supported between instances of 'datetime.datetime' and 'NoneType'
```

**How to reproduce**

```python
from taipy.config import Config 
import taipy as tp
import pandas as pd
import os 

# -------- creation of the Parquet folder ----------
data = pd.DataFrame({"x":[1,2,3,4,5],"y":[4,5,3,1,2]})


if not os.path.exists("parquet"):
    os.mkdir("parquet")

data.to_parquet("parquet/data.parquet")

# the read of Pandas is working
print(pd.read_parquet("parquet/"))
# ----------------------------------------------------

input_cfg = Config.configure_parquet_data_node("parquet_folder", "parquet/")
output_cfg = Config.configure_data_node("output")

def dumb(data):
    return data

task_cfg = Config.configure_task("task", dumb, input_cfg, output_cfg)

[...]

PR #534, opened 2023-03-28
Fix/#524 - Unable to read a parquet folder

trgiangdo:
https://github.com/Avaiga/taipy-core/issues/524

In this PR:
- Fixed the bug, which was caused by `ParquetDataNode.last_edit_date` only check for file and not directory.
- Modified the `DataNode.__get_last_modified_datetime()` method to check for the last modified files inside the directory.
- Add test case for reading a parquet folder.


joaoandre-avaiga (dismissed):

src/taipy/core/data/data_node.py
```diff
@@ -106,15 +106,15 @@ def __init__(
 
         self._properties = _Properties(self, **kwargs)
 
-    @property  # type: ignore
+    @property
```
jrobinAV, line 109: Yeah !

src/taipy/core/data/data_node.py
```diff
@@ -285,9 +285,21 @@ def __getattr__(self, attribute_name):
...
+            for filename in os.listdir(path):
+                filepath = os.path.join(path, filename)
+                if (
+                    os.path.isfile(filepath)
+                    and last_modified_datetime
+                    and datetime.fromtimestamp(os.path.getmtime(filepath)) > last_modified_datetime
+                ):
+                    last_modified_datetime = datetime.fromtimestamp(os.path.getmtime(filepath))
+
+        return last_modified_datetime
```
jrobinAV: I don't get how it can work. Cause the purpose is to loop over the files and to get the latest time. That makes sense. But we initialize `last_modified_datetime` to `None`, and we update it only if `last_modified_datetime is not None`. So we never update it. Am I missing something?

If I am right, that also means we are missing some unit tests.

jrobinAV (changes requested):

src/taipy/core/data/data_node.py
trgiangdo: Ah hah, the logic is completely wrong. I will make sure we have test case for this.

trgiangdo:
The tests are failing on Linux. I'll investigate the problem, maybe due to how the timestamp is handled.

Fixed.

tests/core/data/test_parquet_data_node.py
```diff
@@ -252,20 +252,25 @@ def test_get_system_file_modified_date_instead_of_last_edit_date(self, tmpdir_fa
...
-        dn = ParquetDataNode("foo", Scope.PIPELINE, properties={"path": temp_folder_path, "exposed_type": "pandas"})
-
         pd.DataFrame([]).to_parquet(temp_file_path)
 
-        assert dn.last_edit_date == datetime.fromtimestamp(os.path.getmtime(temp_file_path))
-        previous_edit_date = dn.last_edit_date
+        dn = ParquetDataNode("foo", Scope.PIPELINE, properties={"path": temp_folder_path})
+        initial_edit_date = dn.last_edit_date
+
+        sleep(0.1)
```
jrobinAV, line 261: Can you let a comment to explain why?

jrobinAV (approved):

tests/core/data/test_parquet_data_node.py
trgiangdo: I did edited the above comment. But I guess here is much better 😅

So the problem seems like the after the creation of a file, Ubuntu does something with the metadata of the file, which after a few millisec change the modified date of the file.
This messes up the test since we instantly get the modified time of the file after creation.

To fix this, the file is created first in the test. Then the test was carried-on as usual.

jrobinAV: I was talking about a comment in the code for the following contributors... so they know why we sleep.

🤣 🤣

jrobinAV (dismissed):

tests/core/data/test_parquet_data_node.py
trgiangdo: Ah ... my bad

trgiangdo:
```suggestion
        # Sleep so that the file can be created successfully on Ubuntu
        sleep(0.1)
```

jrobinAV (approved):

tests/core/data/test_parquet_data_node.py
trgiangdo: @jrobinAV Do you think the comment is clear enough? I don't want it to be too long.

One can track it back to the PR anyway.

jrobinAV: YEp. It s clear enough.

Issue #530, opened 2023-03-24
BUG- Core().run() raises an error in VS Code Notebook


FlorianJacta:
**Description**
Executing `tp.Core().run()` will raise an error in VS Code Notebooks:
```

Output exceeds the [size limit](command:workbench.action.openSettings?%5B%22notebook.output.textLineLimit%22%5D). Open the full output data [in a text editor](command:workbench.action.openLargeOutput?e939e06d-a290-41ba-8539-292198447f8a)
---------------------------------------------------------------------------
ArgumentError                             Traceback (most recent call last)
File [c:\Users\jacta\.conda\envs\tp_2_1\Lib\argparse.py:1895](file:///C:/Users/jacta/.conda/envs/tp_2_1/Lib/argparse.py:1895), in ArgumentParser.parse_known_args(self, args, namespace)
   1894 try:
-> 1895     namespace, args = self._parse_known_args(args, namespace)
   1896 except ArgumentError as err:

File [c:\Users\jacta\.conda\envs\tp_2_1\Lib\argparse.py:2103](file:///C:/Users/jacta/.conda/envs/tp_2_1/Lib/argparse.py:2103), in ArgumentParser._parse_known_args(self, arg_strings, namespace)
   2102     # consume the next optional and any arguments for it
-> 2103     start_index = consume_optional(start_index)
   2105 # consume any positionals following the last Optional

File [c:\Users\jacta\.conda\envs\tp_2_1\Lib\argparse.py:2025](file:///C:/Users/jacta/.conda/envs/tp_2_1/Lib/argparse.py:2025), in ArgumentParser._parse_known_args..consume_optional(start_index)
   2024         msg = _('ignored explicit argument %r')
-> 2025         raise ArgumentError(action, msg % explicit_arg)
   2027 # if there is no explicit argument, try to match the
   2028 # optional's string arguments with the following strings
   2029 # if successful, exit the loop
   2030 else:

ArgumentError: argument --force/-f: ignored explicit argument '[c:\\Users\\jacta\\AppData\\Roaming\\jupyter\\runtime\\kernel-v2-99168hn1nHFYc8jn.json](file:///C://Users//jacta//AppData//Roaming//jupyter//runtime//kernel-v2-99168hn1nHFYc8jn.json)'

During handling of the above exception, another exception occurred:

SystemExit                                Traceback (most recent call last)
...
-> 1002     source_file = inspect.getsourcefile(etb.tb_frame)
[...]

PR #535, opened 2023-03-28
Fix/#530 - Rename `--force` to `--force-run`

trgiangdo:
https://github.com/Avaiga/taipy-core/issues/530

The `--force` option in Version CLI is conflicted with Anaconda CLI.

### DIFF
--- a/src/taipy/core/_core.py
+++ b/src/taipy/core/_core.py
@@ -65,7 +65,7 @@ class Core:
         Config.check()
         Config.block_update()
 
-    def __manage_version(self, mode, _version_number, override, clean_entities):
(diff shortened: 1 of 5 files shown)


joaoandre-avaiga (dismissed):
LGTM. But I agree with Fabien's challenge. Not sure if theres a way to avoid this problem with future taipy cli keywords, or with future versions of anaconda. I think we should do some research later and try to find a solution for this problem(if there is any)

jrobinAV (dismissed):
I agree with joao.

This PR is ok for a quick fix, but we need to find a better solution in the future. If possible, I now believe we should not use CLI args anymore.

jrobinAV (approved):

PR #538, opened 2023-03-31
change code organization: create new sub package _entity

jrobinAV:
- Creating _entity package to clean a bit common package
- splitting alias into multiple entity_id files

Merging this PR will require to update all repositories sharing the same `__init__.py` file.


src/taipy/core/cycle/_cycle_manager.py
```diff
@@ -57,7 +54,7 @@ def _get_all_by(cls, by, version_number: Optional[str] = "all") -> List[Cycle]:
 
     @classmethod
     def _get_or_create(
-        cls, frequency: Frequency, creation_date: Optional[datetime] = None, display_name: Optional[str] = None
+        cls, frequency: Frequency, creation_date: Optional[datetime] = None, name: Optional[str] = None
```
trgiangdo, line 57: Are we using the `name` property, or exposing it in the documentation in any way?

I know it's different from `label`, but I don't see the purpose of the `name` (yet).

trgiangdo (approved):
A lot of changes 😆 But all looks good.
Just a small question about the `name` property.

src/taipy/core/cycle/_cycle_manager.py
jrobinAV: The name is an optional attribute of some entities. It s for an end user to name the entities he/she creates so he can differentiate them and access them. He/she does not want to use an id for that.
It is documented in the reference manual.

However, the "display name" was never used or documented. So I removed it.

src/taipy/core/cycle/_cycle_manager.py
```diff
@@ -15,26 +15,23 @@
...
 from .cycle import Cycle
+from .cycle_id import CycleId
 
 
 class _CycleManager(_Manager[Cycle]):
     _repository = _CycleRepositoryFactory._build_repository()  # type: ignore
     _ENTITY_NAME = Cycle.__name__
 
     @classmethod
-    def _create(
```
toan-quach, line 31: Hi @jrobinAV 
Is there a reason why you removed the display_name parameters? It's causing some trouble in the enterprise version :/

Issue #544, opened 2023-04-12
BUG-Error raised because PostgreSQL engine for sql_data_node does not support "driver" argument


s-wel:
**Description**

Executing a scenario that uses a SQL data node with PostgreSQL engine causes an error:
`psycopg2.ProgrammingError: invalid dsn: invalid connection option "driver"`

The driver argument is not supported by psycopg2 (it seems). A temporary fix is to remove the driver argument in the abstract_sql.py file:
https://github.com/Avaiga/taipy-core/blob/13e5a19eac8f977785a43258c93860766e22dcd6/src/taipy/core/data/abstract_sql.py#L169 

A proper fix likely requires that "driver" is a regular (optional) extra argument and not mandatory for all engines.

**How to reproduce**

Try to run a sql_data_node with a PostgreSQL configuration.

- Configuration file example:
    ```
    from taipy import Config;

   pg_config = Config.configure_sql_data_node(
    id="pg_tracker_store",
    db_username="postgres",
    db_extra_args={},
    db_password="xxx",
    db_engine="postgresql",
    db_host="localhost",
    db_port="5432",
    db_name="xxx",
    read_query="SELECT * from events;",
    write_query_builder=write_query_builder
)
    ```

**Expected behavior**
The expected behavior is that the Node can connect to PostgreSQL and retrieve data.

**Runtime environment**
Please specify relevant indications.
 - Taipy version: 2.1
 - OS: Ubuntu 20.04 (WSL) 
and any other relevant information.

Issue likely relates to this line where invalid arguments are passed that psycopg2 does not accept:
https://github.com/Avaiga/taipy-core/blob/13e5a19eac8f977785a43258c93860766e22dcd6/src/taipy/core/data/abstract_sql.py#L179

FlorianJacta:
Thank you for reporting this issue and providing a detailed description of your problem with SQL data nodes and the PostgreSQL engine in Taipy.

We appreciate your temporary solution, as it appears to be an appropriate approach we have also seen after your comment on the Taipy GitHub discussion.

We have added this issue to our backlog for future investigation. Our team will prioritize this matter and work on a proper fix as soon as possible. Once we have an update or a resolution, we will let you know.

trgiangdo:
Hello @s-wel, thank you for reporting the bug.

We have addressed the issue, and the fix will be delivered with taipy-core 2.2.2.

s-wel:
Great! Thanks.

PR #545, opened 2023-04-14
Hotfix2.2.2/#544 - Remove default value of "db_driver" on SQLDataNode

trgiangdo:
https://github.com/Avaiga/taipy-core/issues/544


jrobinAV (commented):
should we add a unit test to verify it s working as expected?

jrobinAV (approved):

PR #546, opened 2023-04-14
Fix/#544 - Remove default value of "db_driver" on SQLDataNode

trgiangdo:
https://github.com/Avaiga/taipy-core/issues/544


jrobinAV (dismissed):
I am surprised that there is no test to change. Otherwise that s good.

jrobinAV (approved):
That s awesome. It s so clear that it could be referenced for documentation

PR #547, opened 2023-04-18
feature/#458 rename DefaultCustomDocument to MongoDefaultDocument

toan-quach:


trgiangdo (commented):
The name change is actually a breaking change, so we've targeted it for 3.0.

The base branch should be dev/3.0.

PR #551, opened 2023-04-19
add tests for core event consumer base class

toan-quach:


tests/core/notification/test_core_event_consumer.py
```diff
@@ -0,0 +1,112 @@
...
+    scenario = tp.create_scenario(scenario_config)
+    sleep(3)
+
+    # assert_true_after_time(lambda: len(event_processor_0.event_type_collected) == 5, time=10)
+    # assert_true_after_time(lambda: event_processor_0.event_operation_collected[EventOperation.CREATION] == 5, time=10)
+    # assert_true_after_time(lambda: event_processor_1.scenario_event_collected == 1, time=10)
+    # assert_true_after_time(lambda: event_processor_1.event_operation_collected[EventOperation.CREATION] == 1, time=10)
+    # assert_true_after_time(lambda: len(event_processor_1.event_operation_collected) == 1, time=10)
+    # assert_true_after_time(lambda: event_processor_2.pipeline_event_collected == 1, time=10)
+    # assert_true_after_time(lambda: event_processor_2.creation_event_operation_collected == 1, time=10)
```
jrobinAV: Why don't you use the `assert_true_after_time` at the end? I believe it's cleaner than a `sleep(3)`.

jrobinAV (dismissed):
Neat!

tests/core/notification/test_core_event_consumer.py
toan-quach: I tried as you can see, but it caused failed test in windows and macos github action 😢 It's still causing them :)

jrobinAV (approved):

PR #552, opened 2023-04-20
feat: SQLRepostioryV2

joaoandre-avaiga:


src/taipy/core/_repository/_v2/_sql_repository.py
```diff
@@ -35,28 +32,35 @@ def __init__(self, model: Type[ModelType], converter: Type[Converter], db_table)
...
+    ###############################
     def _save(self, entity: Entity):
         obj = self.converter._entity_to_model(entity)
-        if entry := self.db.query(self.model).filter_by(model_id=obj.id).first():
+        if entry := self.db.query(self.model).filter_by(id=obj.id).first():
             self.__update_entry(entry, obj)
             return
         self.__insert_model(obj)
 
     def _load(self, model_id: Any) -> Optional[ModelType]:
```
jrobinAV: `model.id` and `entity.id` both represent the same concept: the identifier of the entity.

I would remove model_id everywhere and use `entity_id` instead, even if we compare it to the `model.id`. So we don't create confusion, and we know that there is only one id: the entity's id. 

```suggestion
    def _load(self, entity_id: Any) -> Optional[ModelType]:
```

src/taipy/core/_repository/_v2/_sql_repository.py
```diff
@@ -35,28 +32,35 @@ def __init__(self, model: Type[ModelType], converter: Type[Converter], db_table)
...
-        return self.db.query(self.model).filter(self.model.id == model_id).first()
+        if entry := self.db.query(self.model).filter(self.model.id == model_id).first():
```
jrobinAV:
```suggestion
        if entry := self.db.query(self.model).filter(self.model.id == entity_id).first():
```

src/taipy/core/_repository/_v2/db/_sql_base_model.py
```diff
@@ -0,0 +1,29 @@
...
+# specific language governing permissions and limitations under the License.
+
+from typing import Any
+
+from sqlalchemy.ext.declarative import as_declarative, declared_attr
+from sqlalchemy.orm import registry
+
+
+@as_declarative()
+class Base:
```
jrobinAV: I would align the class name with the module name.

src/taipy/core/_repository/_v2/db/_init_db.py
jrobinAV, line 1: I don't get the purpose of this file.

src/taipy/core/_version/_version_model.py
```diff
@@ -9,20 +9,28 @@
...
+
 
 @dataclass
-class _VersionModel:
+class _VersionModel(_BaseModel):
+    __table__ = Table(
+        "version",
+        mapper_registry.metadata,
+        Column("id", String, primary_key=True),
+        Column("config", String),
```
jrobinAV: I guess we will store the Config as a json string. Right?
If so, Can you add a comment here to explain it?

src/taipy/core/cycle/_cycle_model.py
```diff
@@ -9,17 +9,31 @@
...
 @dataclass
-class _CycleModel:
+class _CycleModel(_BaseModel):
+    __table__ = Table(
+        "cycle",
+        mapper_registry.metadata,
+        Column("id", String, primary_key=True),
+        Column("name", String),
+        Column("frequency", String),
+        Column("properties", String),
```
jrobinAV, line 32: What do you have in mind for properties? A json representation of the dict?

jrobinAV (approved):
I probably need some discussion to better understand the `db` folder. Otherwise it s good enough for me. I approved it, even if I listed some possible improvements. Tell me if they make sense to you: 
* _AbstractRepository:
	- [x] Update all docstrings
	- [x] Move Encoder and Decoder into separate modules: _repository._encoder.py and _repository._decoder.py
	- [ ] Add _save_many(entities: Iterable[Entities])
	- [ ] Add _load_many(model_ids: entities: Iterable[str])
	- [ ] Add _load_by(filters: Optional[List[Dict]] = None)
	- [ ] Change _load_all(filters: Optional[List[Dict]] = None) to _load_all()
	- [x] Change _delete_many documentation adding ids param
	- [ ] Change _delete_by(attribute, value) to _delete_by(attribute, value)-> Iterator[EntityId]
	- [x] Change _delete_by documentation adding params
	- [ ] Change _search(attribute, value, filters)-> Optional[Entity] to _search(attribute, value, filters) -> Iterator[Entity]
	- [ ] Change _search(attribute, value)-> Optional[Entity] documentation adding filters param
			
* _FileSystemRepository:
	- [ ] Rename model to model_type
	- [ ] Move converter that can be hold by parent class _AbstractRepository
	- [x] Inline _get(filepath) that is only used in _load. Otherwise it can be renamed to __get(filepath).
	- [x] Remove __migrate_old_entity. It was there for compatibility reasons.
	- [ ] Change _search(attribute, value)-> Optional[Entity] to _search(filters: Optional[List[Dict]]) -> Iterator[Entity]
		The __search method looks for all entities matching and returns an iterator. It s only used by _search method that only returns one entity.
	
* _SQLRepository:
	- [ ] Rename model to model_type
	- [ ] Move converter that can be hold by parent class _AbstractRepository
	- [x] Rename get_by_config to _get_by_config and move to specific or optimized methods
	- [x] Rename get_multi to _get_multi and move to specific or optimized methods
	- [ ] Remove _get_by_config_and_owner_id, _get_by_configs_and_owner_ids, __get_entities_by_config_and_owner methods. They have been designed to optimize FSRepo. They should not be needed by SQLRepo.
	- [x] Rename param __insert_model method from obj to model

src/taipy/core/_version/_version_model.py
joaoandre-avaiga: Yes, we store the config as JSON. I will add a comment.

src/taipy/core/_repository/_v2/db/_init_db.py
joaoandre-avaiga: This just isolates the database creation(if the tables dont exist, they are created). On the old version, the line `Base.metadata.create_all(bind=engine)` was inside the init on SQLRepository class, which maybe cause the issues with the multiple files open(Im not sure thats the cause). 

I'm still not sure where we will put the database initialization, so for now I just created this function and will experiment until we find the best place for it. This can be removed at the end of the refactoring if it's not needed.

src/taipy/core/cycle/_cycle_model.py
joaoandre-avaiga: Yes, a JSON representation of the dict. On the old SQLRepository we would jsonify the entire entity and store on the document column, so now every field that is a dict it is represented as a json string in the database.

src/taipy/core/_repository/_v2/_encoder.py
```diff
@@ -0,0 +1,39 @@
...
+# Unless required by applicable law or agreed to in writing, software distributed under the License is distributed on
+# an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the
+# specific language governing permissions and limitations under the License.
+
+import json
+from datetime import datetime, timedelta
+from enum import Enum
+from typing import Any
+
+from taipy.core.common.typing import Json
```
jrobinAV:
```suggestion
from ...common.typing import Json
```

joaoandre-avaiga:
_AbstractRepository:

 -> Update all docstrings **(Not sure if I understand this, is something wrong with the docstrings?)**
 -> Move Encoder and Decoder into separate modules: _repository._encoder.py and _repository._decoder.py **(Done)**
 -> Add _save_many(entities: Iterable[Entities]) **(whats the use case for this and load many and load by?)**
 -> Add _load_many(model_ids: entities: Iterable[str])
 -> Add _load_by(filters: Optional[List[Dict]] = None)
 -> Change _load_all(filters: Optional[List[Dict]] = None) to _load_all() **(I agree, however this maybe imply in some changes in the manager, where we call load_all filtering by version)**
 -> Change _delete_many documentation adding ids param **(Done)**
 -> Change _delete_by(attribute, value) to _delete_by(attribute, value)-> Iterator[EntityId]
 -> Change _delete_by documentation adding params **(Done)**
 -> Change _search(attribute, value, filters)-> Optional[Entity] to _search(attribute, value, filters) -> Iterator[Entity]
 -> Change _search(attribute, value)-> Optional[Entity] documentation adding filters param **(Done)**

joaoandre-avaiga:
* _FileSystemRepository:
	- [ ] Rename model to model_type **(Not sure about this one, can you elaborate the reason?)**
	- [ ] Move converter that can be hold by parent class _AbstractRepository **(Converter is specific to each entity, why would it be in the Abstract Class?)**
	- [ ] Inline _get(filepath) that is only used in _load. Otherwise it can be renamed to __get(filepath). **(This was removed)**
	- [ ] Remove __migrate_old_entity. It was there for compatibility reasons. **(Done)**
	- [ ] Change _search(attribute, value)-> Optional[Entity] to _search(filters: Optional[List[Dict]]) -> Iterator[Entity]
		The __search method looks for all entities matching and returns an iterator. It s only used by _search method that only returns one entity.
	
* _SQLRepository:
	- [ ] Rename model to model_type
	- [ ] Move converter that can be hold by parent class _AbstractRepository
	- [ ] Rename get_by_config to _get_by_config and move to specific or optimized methods **(Done)**
	- [ ] Rename get_multi to _get_multi and move to specific or optimized methods **(Done)**
	- [ ] Remove _get_by_config_and_owner_id, _get_by_configs_and_owner_ids, __get_entities_by_config_and_owner methods. They have been designed to optimize FSRepo. They should not be needed by SQLRepo. **They can't be removed unless we change DataManager and TaskManger, which calls these methods directly**
	- [ ] Rename param __insert_model method from obj to model **(Done)**

joaoandre-avaiga:
About the changes in the delete methods(to return the deleted IDs), I would propose moving to a different ticket. What do you think?

jrobinAV:
I believe you are right. We should use another ticket/PR for changing the delete methods, so they return the ID(s).

We can also use another ticket for the renaming I proposed.

For me, the PR is mergeable as it is today.

PR #553, opened 2023-04-20
Small improvements for RefMan generation

FabienLelaquais:


src/taipy/core/config/task_config.py
```diff
@@ -23,20 +23,24 @@
 
 class TaskConfig(Section):
     """
-    Holds all the configuration fields needed to create an actual `Task^` from the `TaskConfig`.
+    Configuration fields needed to create an actual `Task^`.
```
jrobinAV:
```suggestion
    Configuration fields needed to instantiate an actual `Task^`.
```

jrobinAV (dismissed):

jrobinAV (dismissed):

src/taipy/core/task/task.py
```diff
@@ -178,7 +178,7 @@ def submit(
         force: bool = False,
         wait: bool = False,
         timeout: Optional[Union[float, int]] = None,
-    ):
+    ) -> "Job":
```
jrobinAV: I am not sure we can do that. 
Each Job references a Task. So the Task class cannot reference the Job class without a circular import.

FabienLelaquais: There is no import at this point...

jrobinAV (dismissed):
Tests should pass now...

jrobinAV (approved):

PR #555, opened 2023-04-20
Feature/test core event consumer

jrobinAV:
In order to merge develop brnach into dev/3.0 branch, we need to move event publication logic from repositories to managers. Plus I believe the design is better.


src/taipy/core/scenario/_scenario_manager.py
```diff
@@ -306,6 +306,7 @@ def _delete_by_version(cls, version_number: str):
             if scenario.cycle and len(cls._get_all_by_cycle(scenario.cycle)) == 1:
                 _CycleManagerFactory._build_manager()._delete(scenario.cycle.id)
             cls._repository._delete(scenario.id)
+            _publish_event(cls._EVENT_ENTITY_TYPE, scenario.id, EventOperation.DELETION, None)
```
toan-quach:
```suggestion
            super()._delete(scenario.id)
```
I think this would be cleaner

toan-quach (dismissed):
LGTM, just minor remark

src/taipy/core/scenario/_scenario_manager.py
jrobinAV: Yes, indeed. But this should be refactored for the 3.0 version anyway.

Issue #570, opened 2023-05-09
Expose files to preserve between two installation of the user app 


jrobinAV:
- Create a hidden method called in Core service that creates a file containing an exhaustive list of paths that must be kept for the next run (".data/" folder, data files like pickles, csv, ...) if an environment variable is provided (TAIPY_CLOUD_PRESERVED_FILES for example, the variable is expected to provide a path to the file that will store the list of file paths to keep)
- Update this file on the fly when needed : 
        - When a datanode with a path attribute is created 
        - When a datanode with a path attribute is deleted
        - When the path attribute of a datanode  is modified

jrobinAV:
@toan-quach I updated the description, adding more details on when to update the preserved_files file.

jrobinAV:
To update the file, we may want to use the notification system. We just need to create an appropriate event consumer.
The advantages are that we already have the events created, plus the file edition will be asynchronous in a dedicated thread so it does not impact so much the performances!

PR #571, opened 2023-05-11
feature/#570 added check and update preserve file provided by env var

toan-quach:
#570


src/taipy/core/data/csv.py
```diff
@@ -132,8 +132,10 @@ def path(self):
 
     @path.setter
     def path(self, value):
+        tmp_old_path = self._path
         self._path = value
         self.properties[self.__PATH_KEY] = value
+        _AbstractFileDataNode._check_and_update_preserve_file(old_path=tmp_old_path, new_path=self._path)
```
jrobinAV:
```suggestion
        _check_and_update_preserve_file(old_path=tmp_old_path, new_path=self._path)
```

src/taipy/core/data/_data_manager.py
```diff
@@ -101,35 +104,46 @@ def __create(
             raise InvalidDataNodeType(data_node_config.storage_type)
 
     @classmethod
-    def _clean_pickle_file(cls, data_node: Union[DataNode, DataNodeId]):
-        data_node = cls._get(data_node) if isinstance(data_node, str) else data_node
+    def _clean_pickle_file(cls, data_node: DataNode):
```
jrobinAV, line 108: What is this signature change about?

src/taipy/core/data/_data_manager.py
```diff
@@ -101,35 +104,46 @@ def __create(
...
+            _AbstractFileDataNode._check_and_update_preserve_file(old_path=data_node.path)
+
+    @classmethod
+    def _update_preserve_file_and_clean_pickle_file(cls, data_node: Union[DataNode, DataNodeId]):
+        data_node = cls._get(data_node) if isinstance(data_node, str) else data_node
+        cls._clean_pickle_file(data_node)
+        cls._update_preserve_files(data_node)
+
+    @classmethod
+    def _update_preserve_files_and_clean_pickle_files(cls, data_nodes: Iterable[Union[DataNode, DataNodeId]]):
```
jrobinAV: I would not put together the preserve file update and the pickle cleaning. These are completely unrelated features. I would split them in the code as well.

src/taipy/core/data/abstract_file.py
```diff
@@ -26,3 +28,19 @@ def _build_path(self, storage_type):
...
+                    lines = f.readlines()
+                    old_path = f"{old_path}\n"
+                    if old_path in lines:
+                        f.seek(0)
+                        f.truncate()
+                        lines.remove(old_path)
+                        f.writelines(lines)
+            if new_path:
+                with open(preserve_file_path, "a") as f:
+                    f.write(f"{new_path}\n")
```
jrobinAV: why is the name prefixed by check?

src/taipy/core/data/_data_manager.py
```diff
@@ -101,35 +104,46 @@ def __create(
...
-        data_node = cls._get(data_node) if isinstance(data_node, str) else data_node
+    def _clean_pickle_file(cls, data_node: DataNode):
         if not isinstance(data_node, PickleDataNode):
             return
         if data_node.is_generated and os.path.exists(data_node.path):
             os.remove(data_node.path)
 
     @classmethod
-    def _clean_pickle_files(cls, data_nodes: Iterable[Union[DataNode, DataNodeId]]):
+    def _update_preserve_files(cls, data_node: DataNode):
```
jrobinAV: why an 's' at the end of the name. There is only one file.

src/taipy/core/data/abstract_file.py
jrobinAV: That is a minor non blocking remark, but I would create 3 different methods to match the three use cases.
- append_preserve_file(path)
- remove_from_preserve_file(oldpath, newpath)
- replace_from_preserve_file(path)

I believe it is a way to self-document the code. you don't need to read the method to understand what it does.

src/taipy/core/data/abstract_file.py
```diff
@@ -26,3 +28,19 @@ def _build_path(self, storage_type):
         if not dir_path.exists():
             dir_path.mkdir(parents=True, exist_ok=True)
         return dir_path / f"{self.id}.{self.__EXTENSION_MAP.get(storage_type)}"
+
+    @staticmethod
+    def _check_and_update_preserve_file(old_path: Optional[str] = None, new_path: Optional[str] = None):
```
jrobinAV: Since it is not the responsibility of the abstract_file_data_node to preserve the preserve_file, I would move this code to a dedicated class/module.

src/taipy/core/data/_data_manager.py
```diff
@@ -101,35 +104,46 @@ def __create(
...
+    def _update_preserve_files_and_clean_pickle_files(cls, data_nodes: Iterable[Union[DataNode, DataNodeId]]):
         for data_node in data_nodes:
-            cls._clean_pickle_file(data_node)
+            data_node = cls._get(data_node) if isinstance(data_node, str) else data_node
+            cls._update_preserve_file_and_clean_pickle_file(data_node)
 
     @classmethod
     def _delete(cls, data_node_id: DataNodeId):
-        cls._clean_pickle_file(data_node_id)
+        cls._update_preserve_file_and_clean_pickle_file(data_node_id)
```
jrobinAV: I would update the preserve_file after calling super.delete(), just in case there is a pb during the deletion.

src/taipy/core/data/abstract_file.py
toan-quach: I thought about moving it to the common module as it's outside of the main logic of Core. I'm only keeping it in the abstract file because only file data nodes need to preserve the path. What do you think?

src/taipy/core/data/_data_manager.py
toan-quach: Ah I was thinking about merging clean pickle files and updating the preserve file so that we don't have to load the entity twice, once for cleaning pickle files and once again for preserving the file path. Though like you said in your next comment, they are unrelated features.

src/taipy/core/data/abstract_file.py
toan-quach: Not sure if I understand what you meant :/

jrobinAV: I would just name the function `_update_preserve_file`.

src/taipy/core/data/abstract_file.py
jrobinAV: I would move it to a new package named `_backup`, in a `_backup.py` module.
And I would rename `_check_and_update_preserve_file` to `_backup_files`.

src/taipy/core/data/abstract_file.py
toan-quach: Ah got it! I agree!

jrobinAV (dismissed):

src/taipy/core/_backup/_backup.py
```diff
@@ -0,0 +1,40 @@
...
+
+
+import os
+import pathlib
+
+__BACKUP_FILE_PATH_ENVIRONMENT_VARIABLE_NAME = "TAIPY_BACKUP_FILE_PATH"
+
+
+def append_to_backup_file(new_file_path: str):
+    if preserve_file_path := os.getenv(__BACKUP_FILE_PATH_ENVIRONMENT_VARIABLE_NAME, None):
```
florian-vuillemot:
```suggestion
    if preserve_file_path := os.getenv(__BACKUP_FILE_PATH_ENVIRONMENT_VARIABLE_NAME):
```

src/taipy/core/_backup/_backup.py
```diff
@@ -0,0 +1,40 @@
...
+    preserve_file_path = os.getenv(__BACKUP_FILE_PATH_ENVIRONMENT_VARIABLE_NAME, None)
+    if preserve_file_path and pathlib.Path(preserve_file_path).is_file():
+        with open(preserve_file_path, "r+") as f:
+            lines = f.readlines()
+            to_remove_file_path = f"{to_remove_file_path}\n"
+            if to_remove_file_path in lines:
+                f.seek(0)
+                f.truncate()
+                lines.remove(to_remove_file_path)
+                f.writelines(lines)
```
florian-vuillemot:
```suggestion
    if preserve_file_path:
       try: # The file will mostly be present, so testing if the file exists will cost you more than a simple test.
          with open(preserve_file_path, "r+") as f:
              old_backup = f.read()  # No parsing needs
              to_remove_file_path = to_remove_file_path + "\n"
              # If to_remove_file_path can be several times in the string
              new_backup = old_backup.replace(to_remove_file_path, '')
              # otherwise
              new_backup = old_backup.replace(to_remove_file_path, '', 1)
              if new_backup is not old_backup:
                  f.seek(0)
                  f.write(new_backup)
                  # To verify, but I think it is better in that way.
                  # The idea is to write new content on top of the current content then resize the file.
                  f.truncate()
         except: ....
```

src/taipy/core/_backup/_backup.py
```diff
@@ -0,0 +1,40 @@
...
+
+
+def replace_in_backup_file(old_file_path: str, new_file_path: str):
+    remove_from_backup_file(old_file_path)
+    append_to_backup_file(new_file_path)
```
florian-vuillemot, line 50: This function makes a lot of system calls. Perhaps you can create a function based on `remove_from_backup_file` that can replace one string with another and thus make only one `open`.

src/taipy/core/_backup/_backup.py
```diff
@@ -0,0 +1,40 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software distributed under the License is distributed on
+# an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the
+# specific language governing permissions and limitations under the License.
+
+
+import os
+import pathlib
+
+__BACKUP_FILE_PATH_ENVIRONMENT_VARIABLE_NAME = "TAIPY_BACKUP_FILE_PATH"
```
florian-vuillemot, line 16: Not sure it is a good idea, but you can do something like:
```
if preserve_file_path := os.getenv(__BACKUP_FILE_PATH_ENVIRONMENT_VARIABLE_NAME):
  def append_to_backup_file(new_file_path: str):
     ....
  def remove_from_backup_file(to_remove_file_path: str):
     ....
  def replace_in_backup_file(old_file_path: str, new_file_path: str):
     ...
else:
   def append_to_backup_file(*args, **kwargs):
     pass
  def remove_from_backup_file(*args, **kwargs):
     pass
  def replace_in_backup_file(*args, **kwargs):
     pass
```
Just because I don't like the 'if' in every function ^^' but it's a matter of taste ;)

florian-vuillemot (commented):
Great job !
Maybe some improvements on disk access. Let me know if this is possible.

src/taipy/core/_backup/_backup.py
florian-vuillemot: after discussion with @jrobinAV 
There is a problem with:
```
to_remove_file_path = to_remove_file_path + "\n"
new_backup = old_backup.replace(to_remove_file_path, '')
```
in case of nested files:
```
foo/data_node.csv
data_node.csv
bar/data_node.csv
```
So a better solution can be:
```
to_remove_file_path = to_remove_file_path + "\n"
if old.backup.startwith(to_remove_file_path):
    new_backup = old_backup.replace(to_remove_file_path, '', 1)
# Others
new_backup = old_backup.replace("\n" + to_remove_file_path, '')
```

src/taipy/core/_backup/_backup.py
toan-quach: @florian-vuillemot 
I'm looking into manipulating this file as a pandas dataframe to reduce time needed to opening files. We can simple open it once (hopefully) as pandas dataframe, update the data, and write the update to a file. This should prevent us from reading the same file multiple times either for read or write purposes. What do you think?

src/taipy/core/_backup/_backup.py
toan-quach: I'm all good!!

toan-quach: Oh I did notice, if we do it this way, should we consider the case where `TAIPY_BACKUP_FILE_PATH` env variable is set during running of taipy? Because if we have the condition in placed like proposed, it will only check once and set for the entire life-time of the taipy project

src/taipy/core/_backup/_backup.py
florian-vuillemot: That is a good idea. But we discuss it with @jrobinAV and he does not want to create a "cache" now for simplicity.

src/taipy/core/_backup/_backup.py
florian-vuillemot: Very good point. Keep it as is :)

src/taipy/core/_backup/_backup.py
toan-quach: if there's a duplication, do we want to remove both dups? Because I can create 2 data nodes that have access to the same file, if I remove 1 data node, it means I'm removing all same file path, and the data node that still exists will be affected as well

florian-vuillemot (dismissed):
Great!

src/taipy/core/_backup/_backup.py
toan-quach: So I think, we can just ignore this problem for now to get a working feature first. Then come back and optimize it later. What do you think?

jrobinAV (approved):

Issue #569, opened 2023-05-09
Remove Pipeline Scope


jrobinAV:
- [x] Remove pipeline scope.
- [ ] Add a migration path if possible
- [ ] Change the Ref manual
- [ ] Change the User manual
- [ ] Change the 2 Getting started if needed
- [ ] Add enter in release notes

PR #576, opened 2023-05-16
feature/#569 added warning for the use of Scope.PIPELINE

toan-quach:
#569

### DIFF
--- a/src/taipy/core/data/_data_repository.py
+++ b/src/taipy/core/data/_data_repository.py
@@ -14,9 +14,12 @@ from datetime import datetime, timedelta
 from pydoc import locate
 from typing import Any, Dict, Iterable, List, Optional, Union
 
+from taipy.config.common.scope import Scope
+
 from .._repository._repository import _AbstractRepository
 from .._repository._repository_adapter import _RepositoryAdapter
 from ..common._utils import _load_fct
+from ..common._warnings import _warn_deprecated
 from ._data_model import _DataNodeModel
 from .data_node import DataNode
 from .generic import GenericDataNode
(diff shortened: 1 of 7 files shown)


trgiangdo (commented):
Because the default Scope now is PIPELINE, will this show too many warnings?

I think we can only show the warning after changing the default scope.

toan-quach:
@trgiangdo 
yes, you're absolutely right that there will be a lot of messages. I put it in the from_model function simply because that's the only "guaranteed" way the message will pop up. We can also put it when accessing the attribute "scope" of dn or task. But the message will only appear when the user tries to access the "scope" attribute.
As for the default Scope value, I agree, we should change it.

@jrobinAV
Do you think this ticket is also the right place to change the default scope value (Imho, it's a yes)? And if yes, what would be the right default value (I'll say Scope.SCENARIO)?

jrobinAV:
> Scope.SCENARIO

The Scope.SCENARIO is already the default value.

src/taipy/core/data/_data_repository.py
```diff
@@ -205,6 +208,10 @@ def _from_model(self, model: _DataNodeModel):
         validity_period = None
         if model.validity_seconds is not None and model.validity_days is not None:
             validity_period = timedelta(days=model.validity_days, seconds=model.validity_seconds)
+
+        if model.scope == Scope.PIPELINE:
+            _warn_deprecated(Scope.PIPELINE, Scope.SCENARIO)
```
jrobinAV: WHy do you use Scope enum object here, and string in data_node warning?

jrobinAV (dismissed):

src/taipy/core/data/_data_repository.py
toan-quach: Oh you're right, I noticed the deprecated value is expecting a str, so I keep it as str in data_node. But you're right, we can simply pass the scope value instead

toan-quach: same for in task.py

toan-quach: I have changed it to ```_warn_deprecated(f"`{Scope.PIPELINE}`", suggest="other `Scope` value")``` to be consistent with the config checker message. Let me know if that makes sense to you

trgiangdo (commented):
LGTM! Let's wait for the test to pass and I believe this is ready to merge.

trgiangdo (approved):

jrobinAV (approved):

Issue #481, opened 2023-02-21
Add the ability to rename a version of the Taipy application


trgiangdo:
***Description***

The idea is to add an option on the CLI to allow renaming a version name.

`--rename` takes 2 arguments, the first one is the old version name, and the second one is the new version name.

```bash
$ python3 main.py --list-version
Version number                         Mode                   Creation date       
1.0                                    Experiment (latest)    2023-02-02 11:42:31 
fcf1ebd2-9d45-4e32-b48f-18ff870c3438   Development            2023-02-02 11:42:05 


$ python3 main.py --rename 1.0 0.1
Successfully rename version 1.0 to version 0.1


$ python3 main.py --list-version
Version number                         Mode                   Creation date       
0.1                                    Experiment (latest)    2023-02-02 11:42:31 
fcf1ebd2-9d45-4e32-b48f-18ff870c3438   Development            2023-02-02 11:42:05
```

TODO:
- [ ] Add the rename functionality to the versioning system (on all repositories).
- [ ] Update relnotes

jrobinAV:
We cannot capture an empty string on CLI. We could use a keyword like `no-version`.

```
$ python3 main.py --rename no-version 1.0
```

trgiangdo:
On MacOS, the CLI can actually capture an empty string.
For example
```
$ python3 main.py --rename '' 1.0
```
and the `args.rename` is actually `['', '1.0']`. The same for `""`.

However, we need to test if this is a consistent behavior across different types of terminal.

trgiangdo:
Legacy version migration should not be done by the renaming option (since it requires to load the Config, and to check for compatibility with legacy entities as well).

PR #578, opened 2023-05-17
Feature/#481 -  Ability to rename a version of the Taipy application

trgiangdo:
https://github.com/Avaiga/taipy-core/issues/481

### DIFF
--- a/src/taipy/core/_version/_cli/_version_cli.py
+++ b/src/taipy/core/_version/_cli/_version_cli.py
@@ -12,10 +12,16 @@
 import sys
 
 from taipy._cli._base_cli import _CLI
+from taipy.config.exceptions.exceptions import InconsistentEnvVariableError
 from taipy.logger._taipy_logger import _TaipyLogger
 
+from ...data._data_manager_factory import _DataManagerFactory
 from ...exceptions.exceptions import VersionIsNotProductionVersion
+from ...job._job_manager_factory import _JobManagerFactory
+from ...pipeline._pipeline_manager_factory import _PipelineManagerFactory
+from ...scenario._scenario_manager_factory import _ScenarioManagerFactory
 from ...taipy import clean_all_entities_by_version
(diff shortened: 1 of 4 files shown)


src/taipy/core/data/data_node.py
```diff
@@ -221,6 +221,10 @@ def name(self, val):
     def version(self):
         return self._version
 
+    @version.setter
+    def version(self, val):
+        self._version = val
+
```
jrobinAV: Here you are exposing the possibility for a user to modify manually the version of an entity. Isn't it too risky? Can we keep it hidden or private?

trgiangdo: Oh, we are exposing it on `dev/3.0` too.

We ofcourse can remove the setter and just use `_version`.

jrobinAV: Or calling a method named `_set_version` that is hidden or private.

Something like:
```
def _set_version(self, val):
    self._version = val
    DataManagerFactory.build_manager().set(self)
```

The purpose is just to avoid exposing it to the user.

trgiangdo: Do we need the `.set()` method?

Set an entity is handled at the `__rename_version()` method and I don't think we need to move it here.

jrobinAV: Oh, right. 
So the method is even simpler.

trgiangdo: What I meant was that there is no need for a method. Just `entity._version`.
Is there any other constrain?

jrobinAV: Absolutely. You are right!

trgiangdo: We will need to keep this in mind to remove the version setter from dev/3.0. I will take note and do this latter.

jrobinAV (approved):

PR #567, opened 2023-05-04
feature/#450 added get entities by their config id

toan-quach:
#450

PR #580, opened 2023-05-18
Feature/ Add `_get_by_config_id()` to repository v2

trgiangdo:
The purpose of this PR is to merge what was done in https://github.com/Avaiga/taipy-core/pull/567 to dev/3.0 branch.

### DIFF
--- a/src/taipy/core/_manager/_manager.py
+++ b/src/taipy/core/_manager/_manager.py
@@ -127,10 +127,3 @@ class _Manager(Generic[EntityType]):
         Export an entity.
         """
         return cls._repository._export(id, folder_path)
-
-    @classmethod
-    def _get_by_config_id(cls, config_id: str) -> List[EntityType]:
-        """
-        Get all entities by its config id.
-        """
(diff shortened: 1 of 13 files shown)


joaoandre-avaiga:
We had removed any specific methods from the base repositories classes on puporse, for instance, version does not have a config_id, so VersionRepository should nor have access to this method. I would propose that instead on changing the repositories, we chance the manager base class, from this:
```
    @classmethod
    def _get_by_config_id(cls, config_id: str) -> List[EntityType]:
        """
        Get all entities by its config id.
        """
        return cls._repository._get_by_config_id(config_id)
```
to this:
```
    @classmethod
    def _get_by_config_id(cls, config_id: str) -> List[EntityType]:
        """
        Get all entities by its config id.
        """
        return cls._repository._load_all({"config_id": config_id})
```

What do you think?

trgiangdo:
I got your point. I will try to propose something in the next commit.

jrobinAV (approved):

joaoandre-avaiga (approved):

Issue #583, opened 2023-05-22
Event: fire an event when scenario name is modified


FredLL-Avaiga:
**Description**
An event should be fired when scenario name is modified.

Issue #582, opened 2023-05-22
BUG: Notify when entity property is updated


jrobinAV:
Use the notification system to publish an event when a custom property is updated.

PR #584, opened 2023-05-24
feature/#582 added notification when entity property is updated

toan-quach:
#582 and #583

### DIFF
--- a/src/taipy/core/scenario/scenario.py
+++ b/src/taipy/core/scenario/scenario.py
@@ -233,7 +233,6 @@ class Scenario(_Entity, _Submittable, _Labeled):
         return self._properties.get("name")
 
     @name.setter  # type: ignore
-    @_self_setter(_MANAGER_NAME)
     def name(self, val):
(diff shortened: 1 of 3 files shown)


src/taipy/core/scenario/scenario.py
```diff
@@ -233,7 +233,6 @@ def name(self) -> Optional[str]:
         return self._properties.get("name")
 
     @name.setter  # type: ignore
-    @_self_setter(_MANAGER_NAME)
```
jrobinAV, line 236: I don't get why you removed the self setter decorator. But the Unit test looks ok.

jrobinAV (approved):

src/taipy/core/scenario/scenario.py
toan-quach: Ah that's because it's not needed! But setting/updating a new value in self._properties, the dictionary would already set the entity. Having @_self_setter will only set the entity again without any changes!

PR #610, opened 2023-06-03
migrate jobs 

jrobinAV:
When renaming scheduler to orchestrator we did not realize that the scheduler module path was stored in some jobs in the subscribers attribute.

This is a migration path


toan-quach (approved):
lgtm. didn't expect that we're storing scheduler module path!

Issue #609, opened 2023-06-02
BUG- Data Node not found when Core().run()


FlorianJacta:
**Description**
Au unexpected warning appears in this console.

`[2023-06-02 18:20:39][Taipy][ERROR] DataNode not found: DATANODE_d1_1d56702c-b8b9-401f-8610-7596e1209312`

**How to reproduce**

Execute twice this code:
```python
from taipy.config import Config
import taipy as tp

def f(i):
    return i

d1_cfg = Config.configure_data_node('d1', default_data=1)
d2_cfg = Config.configure_data_node('d2')
task_cfg = Config.configure_task(id="task", function=f, input=d1_cfg, output=d2_cfg)
s = Config.configure_scenario_from_tasks('s', [task_cfg])

if __name__ == "__main__":
    tp.Core().run()

    sc = tp.create_scenario(s)
    sc.submit()
```

**Expected behavior**
No warning should be shown.

**Runtime environment**
Please specify relevant indications.
 - Taipy version: Draft 2.3 (6/2/23)

PR #611, opened 2023-06-03
fix data node deletion methods

jrobinAV:
https://github.com/Avaiga/taipy-core/issues/609

delete data node entity after cleaning pickles and removing path from backup in various delete methods


trgiangdo (approved):
LGTM!

Should we add a test case to make sure this won't happen again in the future?

toan-quach (approved):
lgtm :D

PR #612, opened 2023-06-04
Fix - Absolute link in reference manual

trgiangdo:
In this PR:
- The absolute links in the docstring are replaced by the relative link.
- Somehow the `replace_in_backup_file()` method appears in the Reference Manual when built, so I make it internal. @toan-quach Please let me know if it's ok.

### DIFF
--- a/src/taipy/core/_backup/_backup.py
+++ b/src/taipy/core/_backup/_backup.py
@@ -11,18 +11,17 @@
 
 
 import os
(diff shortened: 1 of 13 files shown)


toan-quach (approved):

toan-quach:
Didn't expect that as well but I believe it's all good!

jrobinAV (approved):

PR #613, opened 2023-06-05
Add _clean methods to all sections

jrobinAV:


bobbyshermi:
Looks ok to me and solves my issue.

Thx

jrobinAV:
Can you have a look to the failing tests?

src/taipy/core/config/data_node_config.py
```diff
@@ -306,6 +306,12 @@ def default_config(cls):
             cls._DEFAULT_KEY, cls._DEFAULT_STORAGE_TYPE, cls._DEFAULT_SCOPE, cls._DEFAULT_VALIDITY_PERIOD
         )
 
+    def _clean(self):
+        self._storage_type = self._DEFAULT_STORAGE_TYPE
```
toan-quach, line 310: @jrobinAV @trgiangdo 
For cases where we have a global config like JobConfig, clean JobConfig means cleaning all the changes to the global state. Meanwhile, here clean seems for me to be resetting a specific data node config back to the default and doesn't affect the global behavior of Taipy. Idk but the name "clean" feels kinda weird in that context. :/ For me, clean in that case should mean removing the datanode config but Idk if that makes more sense :/

toan-quach (approved):
The pr looks good to me. But I just cant get the term "clean" for entity config like scenario config, pipeline config, etc. out of my head. Clean to me for global state like job config means return things to default (clean all recent changes). While for entity config here, it means to reset the config back to default. Imo, removing the entity config makes more sense. But other than that, the pr is great :D

toan-quach (approved):
The pr looks good to me. But I just cant get the term "clean" for entity config like scenario config, pipeline config, etc. out of my head. Clean to me for global state like job config means return things to default (clean all recent changes). While for entity config here, it means to reset the config back to default. Imo, removing the entity config makes more sense. But other than that, the pr is great :D

src/taipy/core/config/data_node_config.py
jrobinAV: I understand your point. I had the same concern at first. We can see it in a slightly different way: 

Cleaning means removing all information from the section. Or making the section empty. 
But what does that mean to remove information on scope or make the scope empty? It means we remove the information added by the user on it. So we set the default value. Same for the storage_type, etc...

The whole purpose of this is to keep a reference to the section. so when recreating it, it is the same instance.

toan-quach: hmm If you put it that way then I don't have any complaints :D it still bugs me but I'm fine with your point too :D Thanks!!!

trgiangdo:
The tests are failing because we haven't merged PR on taipy-config yet. I will do it now

jrobinAV (commented):
Looks good to me

Issue #617, opened 2023-06-05
Expose an api to know if a scenario is deletable


jrobinAV:
- [x] Please refer to https://github.com/Avaiga/taipy/issues/166

- [x] Add documentation (user man, ref man, release notes)

PR #625, opened 2023-06-07
Implement new API tp.is_deletable

jrobinAV:
#617 
- Must update the documentation + release notes
- Must propagate the change on __init__.py file to other repositories

### DIFF
--- a/src/taipy/__init__.py
+++ b/src/taipy/__init__.py
@@ -18,6 +18,7 @@ if find_spec("taipy"):
     if find_spec("taipy.core"):
         from taipy.core._core import Core
         from taipy.core.cycle.cycle import Cycle
+        from taipy.core.cycle.cycle_id import CycleId
         from taipy.core.data.data_node import DataNode
         from taipy.core.data.data_node_id import DataNodeId
         from taipy.core.job.job import Job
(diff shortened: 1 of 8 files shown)


src/taipy/__init__.py
trgiangdo, line 1: Why are we removing some imports from this file?

src/taipy/core/scenario/_scenario_manager.py
```diff
@@ -280,6 +268,27 @@ def _compare(cls, *scenarios: Scenario, data_node_config_id: Optional[str] = Non
...
+        if scenario.is_primary:
+            if len(cls._get_all_by_cycle(scenario.cycle)) > 1:
+                return False
+        return True
+
+    @classmethod
+    def _delete(cls, scenario_id: ScenarioId):
+        scenario = cls._get(scenario_id)
+        if scenario.is_primary:
+            if len(cls._get_all_by_cycle(scenario.cycle)) > 1:
```
trgiangdo: Maybe we can refactor this part, something like if not _is_deletable then raise error.
And the `_hard_delete()` method as well.
What do you think?

trgiangdo (commented):
I have a question and a minor comment

src/taipy/__init__.py
jrobinAV: Oh... Mypy did it cause they are not used. 
We must not delete the imports. I will revert this part.
Thx!

jrobinAV: Even more, we need to add the new method is_deletable to the imports and propagate it to other repos.

src/taipy/core/scenario/_scenario_manager.py
jrobinAV: Yes, good idea.

trgiangdo (approved):
Neat!

toan-quach (approved):
lgtm :D

PR #629, opened 2023-06-08
feat: Version Repository Interface

joaoandre-avaiga:
JR suggested that maybe some methods from VersionRepo could be in the manager, but the _set(latest, development, production) and _get(latest, development, production) specified in the version repos have in their core accesses and logic that are exclusively to each repository type, so I don't see a way to make them more generic at this momment. 

As a compromise, I've created an interface to be used on the version repositories to ensure that at least the methods signatures are the same across all of them. 

Let me know what do you think.


trgiangdo:
You are right, the `_get` maybe similar but the `_set` methods for each mode is really different.

I'm not sure if we need Interface for this, but I'm ok with it.

jrobinAV (approved):
OK so that s not much added value, but I would keep the interface to avoid futur regression when changing one of the two implementations. Plus it could be useful on enterprise version as well.

jrobinAV (approved):

Issue #537, opened 2023-03-30
Improve the `_Reloader` to a thread-safe singleton


trgiangdo:
**Description**

The `_Reloader` singleton right now is not thread-safe.

This can cause problems when a thread just enter the context (to migrate an entity), when another thread tries to read another entity (which requires the reloader to run).

- Possible solution:
    - An example of thread-safe singleton design in Python: https://refactoring.guru/design-patterns/singleton/python/example#example-1.
    - We can do something similar, maybe one lock for each entity type (need discussion).

trgiangdo:
This is not necessary since:
- In multi-thread runtime, the singleton doesn't have any attribute, so all thread will always share the exact same instance (even when 2 instances are created at the start of the program). Also, Taipy provide a multi-processing environment with standalone mode, not multi-thread. 
- In multi-processing runtime (standalone mode), each process will have its own _Reloader singleton, so there will be no overlap whatsoever.

PR #632, opened 2023-06-09
Fix/#537 - Improve the `_Reloader` to a thread-safe singleton

trgiangdo:
https://github.com/Avaiga/taipy-core/issues/537


trgiangdo:
The lock doesn't work properly since there can be multiple nested _Reloader context when reloading a pipeline of a scenario.

Issue #614, opened 2023-06-05
POC - propose a prototype for Removing PipelineConfig


jrobinAV:
- Implement a ScenarioConfig simplification (without pipelineConfig)
- Implement create_scenario_method from the new ScenarioConfig

More details on specifications:
https://docs.google.com/document/d/1Oa2asZXXbtZ7noJWvvgAbwEAYz1pEy_R6YkCYy__C8g

jrobinAV:
Must be shared with the Studio team.

PR #634, opened 2023-06-12
feature/#614 propose changes to remove pipeline config

toan-quach:
#614

What I did:
- Remove `pipelines` and `pipeline_configs` parameters/attributes in ScenarioConfig and replaced them with `tasks_and_data_nodes` or `task_and_data_node_configs`. (The reason why I think it's better to have a specific attribute rather than have `*tasks_and_data_nodes` is to leverage the type hint to make it easier for the user when configuring the ScenarioConfig)
- Kept get pipeline config in the `_from_dict` function for migration purposes (might have missed some places but will try to improve later). If the `as_dict` attribute has `tasks_and_data_nodes` => ignore `pipelines`, if it doesn't, get the PipelineConfig from the pipeline key, get all the tasks in said pipeline config, and ignore pipeline config


src/taipy/core/config/scenario_config.py
```diff
@@ -66,63 +85,128 @@ def __init__(
...
+            copy(self._tasks),
+            copy(self._additional_data_nodes),
+            self.frequency,
+            copy(comp),
+            **copy(self._properties),
+        )
 
     def __getattr__(self, item: str) -> Optional[Any]:
-        return _tpl._replace_templates(self._properties.get(item))
+        return _tpl._replace_templates(self._properties.get(item))  # type: ignore
```
jrobinAV: Why do we add a type ignore here?

src/taipy/core/config/scenario_config.py
```diff
@@ -66,63 +85,128 @@ def __init__(
...
-    def _to_dict(self):
+    def _to_dict(self) -> Dict[str, Any]:
         return {
             self._COMPARATOR_KEY: self.comparators,
-            self._PIPELINE_KEY: self._pipelines,
+            self._TASKS_KEY: list(self._tasks),
+            self._ADDITIONAL_DATA_NODES_KEY: list(self._additional_data_nodes),
             self._FREQUENCY_KEY: self.frequency,
-            **self._properties,
+            **self._properties,  # type: ignore
```
jrobinAV: Same here. Why do we need to add the type ignore?

src/taipy/core/config/scenario_config.py
```diff
@@ -66,63 +85,128 @@ def __init__(
...
-        pipelines = []
-        if pipeline_ids := as_dict.pop(cls._PIPELINE_KEY, None):
-            pipelines = [p_configs[p_id] for p_id in pipeline_ids if p_id in p_configs]
+
+        tasks = set()
+        additional_data_nodes = set()
+
+        if cls._TASKS_KEY in as_dict or cls._ADDITIONAL_DATA_NODES_KEY in as_dict:
+            if task_ids := as_dict.pop(cls._TASKS_KEY, None):
+                task_configs = config._sections[TaskConfig.name]  # type: ignore
```
jrobinAV: are the type ignore still necessary?

src/taipy/core/config/scenario_config.py
```diff
@@ -66,63 +85,128 @@ def __init__(
...
+                for task_id in task_ids:
+                    if task_config := task_configs.get(task_id, None):
+                        tasks.add(task_config)
```
jrobinAV:
```suggestion
                tasks = [task_configs[t_id] for t_id in tasks_ids if t_id in task_configs]
```

src/taipy/core/config/scenario_config.py
```diff
@@ -66,63 +85,128 @@ def __init__(
...
+            if additional_data_node_ids := as_dict.pop(cls._ADDITIONAL_DATA_NODES_KEY, None):
+                data_node_configs = config._sections[DataNodeConfig.name]  # type: ignore
+                for additional_data_node_id in additional_data_node_ids:
+                    if additional_data_node_config := data_node_configs.get(additional_data_node_id, None):
+                        additional_data_nodes.add(additional_data_node_config)
```
jrobinAV:
```suggestion
                additional_data_nodes = [data_node_configs[dn_id] for dn_id in additional_data_node_ids if dn_id in data_node_configs]

```

src/taipy/core/config/scenario_config.py
```diff
@@ -66,63 +85,128 @@ def __init__(
...
+        else:
+            # Check if pipeline configs exist, if yes, migrate by getting all task configs and ignore pipeline configs
+            if pipeline_ids := as_dict.pop(cls._PIPELINES_KEY, None):
+                pipeline_configs = config._sections[PipelineConfig.name]  # type: ignore
+                for p_id in pipeline_ids:
+                    if pipeline_config := pipeline_configs.get(p_id, None):
+                        tasks.update(pipeline_config.tasks)
```
jrobinAV: smart

src/taipy/core/config/scenario_config.py
```diff
@@ -66,63 +85,128 @@ def __init__(
...
+
         self.comparators = as_dict.pop(self._COMPARATOR_KEY, self.comparators)
         if self.comparators is None and default_section:
             self.comparators = default_section.comparators
-        self._properties.update(as_dict)
+
+        self._properties.update(as_dict)  # type: ignore
         if default_section:
-            self._properties = {**default_section.properties, **self._properties}
+            self._properties = {**default_section.properties, **self._properties}  # type: ignore
```
jrobinAV: do we need to add  type ignore?

src/taipy/core/scenario/_scenario_converter.py
```diff
@@ -43,7 +47,8 @@ def _model_to_entity(cls, model: _ScenarioModel) -> Scenario:
         scenario = Scenario(
             scenario_id=model.id,
             config_id=model.config_id,
-            pipelines=model.pipelines,
+            tasks=set(model.tasks),
+            additional_data_nodes=set(model.additional_data_nodes),
```
jrobinAV: breaking change.

src/taipy/core/scenario/_scenario_model.py
```diff
@@ -30,7 +32,8 @@ class _ScenarioModel(_BaseModel):
         mapper_registry.metadata,
         Column("id", String, primary_key=True),
         Column("config_id", String),
-        Column("pipelines", JSON),
+        Column("tasks", JSON),
+        Column("additional_data_nodes", JSON),
```
jrobinAV, line 36: breaking change

src/taipy/core/scenario/scenario.py
```diff
@@ -83,7 +90,15 @@ def __init__(
         super().__init__(subscribers)
         self.config_id = _validate_id(config_id)
         self.id: ScenarioId = scenario_id or self._new_id(self.config_id)
-        self._pipelines = pipelines
+
+        self._tasks: Union[Set[TaskId], Set[Task], Set] = set(tasks) if tasks else set()
+        if not self._tasks and self._MIGRATED_PIPELINES_KEY in properties:
```
jrobinAV: We should not have _MIGRATED_PIPELINES_KEY in entities. The migration for old entities should be encapsulated at the reposiotry/model level. Not the entity/manager level.

src/taipy/core/scenario/scenario.py
```diff
@@ -96,6 +111,24 @@ def _new_id(config_id: str) -> ScenarioId:
...
+        for pipeline_or_id in pipelines:
+            p = pipeline_manager._get(pipeline_or_id, pipeline_or_id)
+
+            if not isinstance(p, Pipeline):
+                raise NonExistingPipeline(pipeline_or_id)
+
+            tasks.update(p.tasks.values())
+
+        return tasks
+
```
jrobinAV: If needed, should be moved to model/repository level.

src/taipy/core/scenario/scenario.py
```diff
@@ -160,14 +215,14 @@ def _get_set_of_tasks(self) -> Set[Task]:
...
+    @property  # type: ignore
+    @_self_reload(_MANAGER_NAME)
     def data_nodes(self) -> Dict[str, DataNode]:
-        data_nodes = {}
-        list_data_nodes = [pipeline.data_nodes for pipeline in self.pipelines.values()]
-        for data_node in list_data_nodes:
-            for k, v in data_node.items():
-                data_nodes[k] = v
-        return data_nodes
+        # breakpoint()
```
jrobinAV: Don't forget to remove

src/taipy/core/scenario/scenario.py
```diff
@@ -382,24 +437,26 @@ def is_deletable(self) -> bool:
...
-            pipelines[p.config_id] = p
-        return pipelines
+    #         if not isinstance(p, Pipeline):
+    #             raise NonExistingPipeline(pipeline_or_id)
+    #         pipelines[p.config_id] = p
+    #     return pipelines
 
     @classmethod
-    def _to_model(cls, entity) -> _ScenarioModel:
+    def _to_model(cls, entity: Scenario) -> _ScenarioModel:
```
jrobinAV: is the method still used in 3.0?

src/taipy/core/scenario/scenario.py
```diff
@@ -414,7 +471,8 @@ def _from_model(cls, model: _ScenarioModel):
         scenario = Scenario(
```
jrobinAV: is the method still used in 3.0?

tests/core/config/checkers/test_config_id_checker.py
```diff
@@ -23,7 +23,7 @@ def test_check_standalone_mode(self, caplog):
         assert len(Config._collector.errors) == 0
 
         Config.configure_data_node(id="foo", storage_type="in_memory")
-        Config.configure_scenario(id="bar", pipeline_configs=[])
+        Config.configure_scenario(id="bar", task_and_data_node_configs=[])
```
jrobinAV: THe signature is wrong, no?

jrobinAV (commented):
1. can we try to reduce the nb of type ignore ?
2. I still don't know about removing the pipeline model. It seems a good idea, but I am a bit scared it s too disruptive. I would like to run some migration tests (running an app in 2.3, creating scenarios, running the same app in 3.0 and assert the old entities are still accessible)
3. The migration of old entities should be encapsulated in the repository/model/converter level. Not at the Manager/entity level

jrobinAV:
Oh btw, we need to update `taipy/core/config/config.schema.json` as well for Studio

src/taipy/core/scenario/scenario.py
toan-quach: I believe it's a no. But I will leave the decision to Joao when he works on the removal of the old repositories and I can just merge the changes from his part :D

src/taipy/core/scenario/scenario.py
toan-quach: same answer as above

jrobinAV:
Oops I miss clicked and set the PR to 'ready to review' by mistake. SOrry @toan-quach.
It seems I cannot revert it.

src/taipy/core/scenario/scenario.py
toan-quach: I moved it to the scenario_converter.py

tests/core/config/checkers/test_config_id_checker.py
toan-quach: yes, it's the old version I haven't fixed yet

src/taipy/core/config/scenario_config.py
```diff
@@ -66,63 +85,147 @@ def __init__(
...
         self.frequency = None
         self.comparators = dict()
         self._properties.clear()
 
-    def _to_dict(self):
+    def _to_dict(self) -> Dict[str, Any]:
         return {
             self._COMPARATOR_KEY: self.comparators,
-            self._PIPELINE_KEY: self._pipelines,
+            self._TASKS_KEY: list(self._tasks),
```
jrobinAV: set instead of list?

src/taipy/core/config/scenario_config.py
```diff
@@ -66,63 +85,147 @@ def __init__(
...
+            self._ADDITIONAL_DATA_NODES_KEY: list(self._additional_data_nodes),
```
jrobinAV: set instead of list?

src/taipy/core/config/scenario_config.py
```diff
@@ -66,63 +85,147 @@ def __init__(
...
-    def _from_dict(cls, as_dict: Dict[str, Any], id: str, config: Optional[_Config]):  # type: ignore
+    def _from_dict(cls, as_dict: Dict[str, Any], id: str, config: Optional[_Config] = None) -> "ScenarioConfig":  # type: ignore
         as_dict.pop(cls._ID_KEY, id)
-        p_configs = config._sections.get(PipelineConfig.name, None) or []  # type: ignore
-        pipelines = []
-        if pipeline_ids := as_dict.pop(cls._PIPELINE_KEY, None):
-            pipelines = [p_configs[p_id] for p_id in pipeline_ids if p_id in p_configs]
+
+        if cls._TASKS_KEY in as_dict:
+            task_config_ids = as_dict.pop(cls._TASKS_KEY, list())
```
jrobinAV, line 169: set instead of list?

src/taipy/core/config/scenario_config.py
```diff
@@ -66,63 +85,147 @@ def __init__(
...
+            tasks = cls.__get_task_configs(task_config_ids, config)
+        else:
+            # Check if pipeline configs exist, if yes, migrate by getting all task configs and ignore pipeline configs
+            pipeline_config_ids = as_dict.pop(cls._PIPELINES_KEY, list())
+            tasks = cls.__get_task_configs_from_pipeline_configs(pipeline_config_ids, config)
+
+        additional_data_node_ids = as_dict.pop(cls._ADDITIONAL_DATA_NODES_KEY, list())
```
jrobinAV, line 179: set instead of list?

src/taipy/core/cycle/_cycle_manager.py
```diff
@@ -111,18 +111,20 @@ def _get_children_entity_ids(cls, cycle: Cycle) -> _EntityIds:
...
+            # for pipeline in scenario.pipelines.values():
+            owner_ids = {scenario.id, cycle.id}
+            #     if pipeline.owner_id in owner_ids:
+            #         entity_ids.pipeline_ids.add(pipeline.id)
+            for task in scenario.tasks.values():
+                if task.owner_id in owner_ids:
+                    entity_ids.task_ids.add(task.id)
+                for data_node in task.data_nodes.values():
+                    if data_node.owner_id in owner_ids:
+                        entity_ids.data_node_ids.add(data_node.id)
```
jrobinAV:
```suggestion
            for data_node in scenario.data_nodes.values():
                if data_node.owner_id in owner_ids:
                    entity_ids.data_node_ids.add(data_node.id)
```

src/taipy/core/scenario/_scenario_converter.py
```diff
@@ -40,10 +44,18 @@ def _entity_to_model(cls, scenario: Scenario) -> _ScenarioModel:
 
     @classmethod
     def _model_to_entity(cls, model: _ScenarioModel) -> Scenario:
+        tasks = None
+        if model.tasks:
+            tasks = set(model.tasks)
+        else:
+            if model.pipelines:
+                tasks = Scenario._get_set_of_tasks_from_pipelines(model.pipelines)
```
jrobinAV: please move this part to the scenario model module. And isolate it in a dedicated method.
You have an example of migration functions in `data_node_model` module.

src/taipy/core/scenario/_scenario_manager.py
```diff
@@ -105,46 +105,55 @@ def _create(
...
         is_primary_scenario = len(cls._get_all_by_cycle(cycle)) == 0 if cycle else False
         props = config._properties.copy()
         if name:
             props["name"] = name
         version = cls._get_latest_version()
+
         scenario = Scenario(
             str(config.id),
-            pipelines,
+            set(tasks),
```
jrobinAV: Please create a new issue in the product backlog to make `_task_manager._bulk_get_or_create` and `_data_manager._bulk_get_or_create` methods return a set directly. This could avoid converting the list into set.

src/taipy/core/scenario/_scenario_manager.py
```diff
@@ -105,46 +105,55 @@ def _create(
...
             props,
+            set(additional_data_nodes),
```
jrobinAV: Please create a new issue in the product backlog to make _task_manager._bulk_get_or_create and _data_manager._bulk_get_or_create methods return a set directly. This could avoid converting the list into set.

src/taipy/core/scenario/_scenario_manager.py
```diff
@@ -345,17 +354,14 @@ def _delete_by_version(cls, version_number: str):
...
-                    if data_node.owner_id in (pipeline.id, scenario.id):
-                        entity_ids.data_node_ids.add(data_node.id)
+        for task in scenario.tasks.values():
+            if not isinstance(task, Task):
+                task = _TaskManagerFactory._build_manager()._get(task)
+            if task.owner_id == scenario.id:
+                entity_ids.task_ids.add(task.id)
+            for data_node in task.data_nodes.values():
+                if data_node.owner_id == scenario.id:
+                    entity_ids.data_node_ids.add(data_node.id)
```
jrobinAV:
```suggestion
        for data_node in scenario.data_nodes.values():
            if data_node.owner_id == scenario.id:
                entity_ids.data_node_ids.add(data_node.id)
```

src/taipy/core/scenario/_scenario_model.py
```diff
@@ -30,37 +32,43 @@ class _ScenarioModel(_BaseModel):
...
     cycle: Optional[CycleId] = None
 
     @staticmethod
     def from_dict(data: Dict[str, Any]):
         return _ScenarioModel(
             id=data["id"],
             config_id=data["config_id"],
-            pipelines=data["pipelines"],
+            tasks=data.get("tasks", None),
+            additional_data_nodes=data.get("additional_data_nodes", None),
```
jrobinAV, line 65: Please make the migration at the lowest possible level, i.e., here.
You can have a look at the data node model module to have an example.

src/taipy/core/scenario/scenario.py
```diff
@@ -30,25 +30,31 @@
...
+from ..task.task_id import TaskId
+from ._scenario_model import _ScenarioModel
 from .scenario_id import ScenarioId
 
 
 class Scenario(_Entity, _Submittable, _Labeled):
     """Instance of a Business case to solve.
 
-    A scenario holds a list of pipelines (instances of `Pipeline^` class) to submit for execution
+    A scenario holds a list of tasks (instances of `Task^` class) to submit for execution
```
jrobinAV: Can you add a few words on additional data nodes as well?

src/taipy/core/config/scenario_config.py
```diff
@@ -319,3 +336,17 @@ def _configure_default(
         )
         Config._register(section)
         return Config.sections[ScenarioConfig.name][_Config.DEFAULT_KEY]
+
+    def add_sequences(self, sequences: Dict[str, List[TaskConfig]]):
+        # TODO: How about user just provides task config id
```
jrobinAV: I would not accept ids. It just adds some complexity to me, for very little added value.

tests/core/config/test_config_serialization.py
```diff
@@ -101,6 +101,9 @@ def config_test_scenario():
         comparators={test_json_dn_cfg.id: compare_function},
         frequency=Frequency.DAILY,
     )
+    test_scenario_cfg.add_sequences({"sequence1": [test_task_cfg]})
+
+    # Config._register(test_scenario_cfg)
```
jrobinAV: ?

src/taipy/core/scenario/scenario.py
jrobinAV: I believe it should be moved to scenario model file.

tests/core/config/test_config_serialization.py
toan-quach: ah just for testing, I'll remove it

src/taipy/core/scenario/scenario.py
toan-quach: oh? Since it's the transition from scenarioModel to Scenario so I thought scenario_converter would fit the job best no?

jrobinAV: I was thinking of making the convertion even before. Between the json dict and the scenarioModel. I believe there is some similar code in datamodel.

toan-quach: For the data model, we're migrating from the repo data to the model, I'm assuming this is what you meant. The migration here for scenario, requires us to retrieve the Pipeline entity as well and not just its id to get the Task entities. So I think it's more complicated in the end to get Pipeline -> Task -> TaskId for ScenarioModel, the in ScenarioModel to Scenario, we still need to get the Task entities again, so I think it should be more efficient to do it once in the converter. Of course it's not complicated to do it in ScenarioModel if you still want to keep it that way :D

jrobinAV: You are right. That s a very good point. Then the converter seems to be the right location then.

src/taipy/core/config/checkers/_scenario_config_checker.py
```diff
@@ -34,6 +34,8 @@ def _check(self) -> IssueCollector:
                 self._check_task_configs(scenario_config_id, scenario_config)
                 self._check_addition_data_node_configs(scenario_config_id, scenario_config)
                 self._check_comparators(scenario_config_id, scenario_config)
+
+        # TODO: check sequences use tasks in tasks
```
jrobinAV: Can you create a separate ticket to implement the checker, so this one does not grow too much?

src/taipy/core/config/checkers/_scenario_config_checker.py
```diff
@@ -77,3 +92,36 @@ def _check_comparators(self, scenario_config_id: str, scenario_config: ScenarioC
...
+        data_node_configs = set()
+        for task_config in scenario_config.task_configs:
+            if isinstance(task_config, TaskConfig):
+                input_dn_configs = task_config.input_configs if task_config.input_configs else []
+                output_dn_configs = task_config.output_configs if task_config.output_configs else []
+                data_node_configs.update({*input_dn_configs, *output_dn_configs})
+
+        for additional_data_node_config in scenario_config.additional_data_node_configs:
+            if additional_data_node_config in data_node_configs:
+                self._error(
```
jrobinAV: I believe it should be a warning. No?

src/taipy/core/config/scenario_config.py
```diff
@@ -31,31 +32,54 @@ class ScenarioConfig(Section):
 
     Attributes:
         id (str): Identifier of the scenario config. It must be a valid Python variable name.
-        pipeline_configs (Union[PipelineConfig, List[PipelineConfig]]): List of pipeline configs.<br/>
-            The default value is [].
+        tasks (Optional[Union[TaskConfig, List[TaskConfig]]]): List of task configs.<br/>
+            The default value is None.
+        additional_data_nodes (Optional[Union[DataNodeConfig, List[DataNodeConfig]]]): <br/>
+            List of additional data node configs. The default value is None.
+        frequency (Optional[Frequency]): The frequency of the scenario's cycle. The default value is None.
```
jrobinAV, line 39:
```suggestion
        frequency (Optional[Frequency]): The frequency of the scenario's cycle. The default value is None.
        sequences (Dict[str, List[TaskConfig]]): List of sequences.<br/> The default value is [].
```

src/taipy/core/scenario/_scenario_manager.py
```diff
@@ -105,36 +108,63 @@ def _create(
...
+        additional_data_nodes = (
+            _data_manager._bulk_get_or_create(config.additional_data_node_configs, cycle_id, scenario_id)
+            if config.additional_data_node_configs
+            else {}
+        )
         pipelines = [
-            _PipelineManagerFactory._build_manager()._get_or_create(p_config, cycle.id if cycle else None, scenario_id)
-            for p_config in config.pipeline_configs
+            _PipelineManagerFactory._build_manager()._get_or_create(
+                PipelineConfig(p_config_id, task_configs), cycle.id if cycle else None, scenario_id
```
jrobinAV, line 132: OH I did not realize until now that we have to build this PipelineConfig to create the Pipeline.
That s OK for now, but I created a ticket to remove this. 

Basically, we should create a Pipeline from a list of tasks or task ids. It should be enough. No need for a PipelineConfig to be instantiated.

Here is the ticket : https://github.com/Avaiga/taipy-core/issues/687

What do you think?

src/taipy/core/scenario/scenario.py
```diff
@@ -30,25 +30,31 @@
...
 from .scenario_id import ScenarioId
 
 
 class Scenario(_Entity, _Submittable, _Labeled):
     """Instance of a Business case to solve.
 
-    A scenario holds a list of pipelines (instances of `Pipeline^` class) to submit for execution
-    in order to solve the Business case.
+    A scenario holds a list of tasks (instances of `Task^` class) to submit for execution in order to
+    solve the Business case. , It also holds a list of additional data nodes (instances of `DataNode` class)
```
jrobinAV:
```suggestion
    solve the Business case. It also holds a list of additional data nodes (instances of `DataNode` class)
```

src/taipy/core/scenario/scenario.py
```diff
@@ -30,25 +30,31 @@
...
+    for extra data related to the scenario.
 
     Attributes:
         config_id (str): The identifier of the `ScenarioConfig^`.
-        pipelines (List[Pipeline^]): The list of pipelines.
+        tasks (Set[Task^]): The set of tasks.
+        additional_data_nodes (Set[DataNode^]): The set of additional data nodes.
         properties (dict[str, Any]): A dictionary of additional properties.
```
jrobinAV, line 60:
```suggestion
        pipelines (List[Pipeline^]): The list of pipelines: subsets of tasks that can be submitted together independently from the rest of the scenario's tasks.
        properties (dict[str, Any]): A dictionary of additional properties.
```

src/taipy/core/scenario/scenario.py
```diff
@@ -112,59 +145,105 @@ def __getattr__(self, attribute_name):
...
-                if task_config_id in tasks:
-                    if task not in tasks[task_config_id]:
-                        tasks[task_config_id].append(task)
-                else:
-                    tasks[task_config_id] = [task]
-        return tasks
+    def add_pipelines(self, pipelines: Union[List[PipelineId], List[Pipeline]]):
+        _pipelines = list(self.pipelines.values())
+        _pipelines.extend(pipelines)
+        self.pipelines = _pipelines  # type: ignore
```
jrobinAV, line 173: We must check if the tasks of the pipelines to add belongs to the scenario's tasks.
Can you create a ticket for that in the product backlog?

tests/core/_orchestrator/test_orchestrator.py
```diff
@@ -763,51 +730,51 @@ def test_blocked_task():
...
-        assert _DataManager._get(task_1.bar.id).is_ready_for_reading  # bar becomes ready
-        assert _DataManager._get(task_1.bar.id).read() == 2  # the data is computed and written
-        assert_true_after_time(job_2.is_running)  # And job 2 can start running
-        assert_true_after_time(lambda: len(_OrchestratorFactory._dispatcher._dispatched_processes) == 1)
-        assert len(_Orchestrator.blocked_jobs) == 0
-    assert_true_after_time(job_2.is_completed)  # job 2 unlocked so it can complete
-    assert _DataManager._get(task_2.baz.id).is_ready_for_reading  # baz becomes ready
-    assert _DataManager._get(task_2.baz.id).read() == 6  # the data is computed and written
-    assert_true_after_time(lambda: len(_OrchestratorFactory._dispatcher._dispatched_processes) == 0)
+# def test_blocked_pipeline():
```
jrobinAV: Why removing it? It passes on my laptop.
It s very similar to `test_blocked_scenario`, but still it tests the pipeline submission works fine when there are blocked jobs.

tests/core/cycle/test_cycle_manager.py
```diff
@@ -199,19 +199,24 @@ def test_hard_delete_shared_entities():
...
+    scenario_config_1.add_sequences(
+        {
+            "pipeline_config_1": [task_config_1, task_config_2],
+            "pipeline_config_2": [task_config_1, task_config_2],
+            "pipeline_config_3": [task_config_3],
+        }
+    )
     scenario_config_2 = Config.configure_scenario(
-        "scenario_config_2", [pipeline_config_3]
+        "scenario_config_2", [task_config_2, task_config_3]
```
jrobinAV, line 217: Here you changed the use case. you should have added only task_config_3 to mathc the previous use case with pipeline_config_3.
Is it on purpose?

tests/core/cycle/test_cycle_manager_with_sql_repo.py
```diff
@@ -209,19 +209,24 @@ def test_hard_delete_shared_entities():
...
+    scenario_config_1.add_sequences(
+        {
+            "pipeline_config_1": [task_config_1, task_config_2],
+            "pipeline_config_2": [task_config_1, task_config_2],
+            "pipeline_config_3": [task_config_3],
+        }
+    )
     scenario_config_2 = Config.configure_scenario(
-        "scenario_config_2", [pipeline_config_3]
+        "scenario_config_2", [task_config_2, task_config_3]
```
jrobinAV, line 227: same remark. Is this change on purpose?

tests/core/data/test_data_manager.py
```diff
@@ -539,7 +539,6 @@ def test_create_dn_from_loaded_config_modified_default_config(self):
         Config.set_default_data_node_configuration(storage_type="csv")
         scenario = tp.create_scenario(Config.scenarios["my_scenario"])
 
-        # assert isinstance(scenario.input, CSVDataNode) TODO Replace the next line by the commented one.
```
jrobinAV: This is an old bug, no?
we change the default data node storage type to CSV. When creating the data node, I expect to have a CSVDataNode and not a PickleDataNode.

New ticket: https://github.com/Avaiga/taipy-core/issues/688

tests/core/data/test_data_node.py
```diff
@@ -273,7 +271,9 @@ def test_do_not_recompute_data_node_up_to_date_but_continue_pipeline_execution(s
         assert scenario.C.read() == "C"
         assert scenario.D.read() == "D"
 
-        assert len(tp.get_jobs()) == 4
```
jrobinAV, line 276: It should not have been 4 before, but 3. No? 
Did we fix a bug ?

tests/core/notification/test_core_event_consumer.py
```diff
@@ -43,12 +43,12 @@ def process_event(self, event: Event):
 
 class PipelineCreationCoreEventConsumerProcessor(CoreEventConsumerBase):
     def __init__(self, registration_id: str, queue: SimpleQueue):
-        self.pipeline_event_collected = 0
```
jrobinAV, line 46: You changed the `PipelineCreationCoreEventConsumerProcessor` to make it a task creation event consumer.

1/ We should rename the consumer.
2/ Maybe we should have both to have a better tests coverage.

tests/core/scenario/test_scenario_converter.py
```diff
@@ -0,0 +1,453 @@
...
+from src.taipy.core.pipeline._pipeline_manager_factory import _PipelineManagerFactory
+from src.taipy.core.pipeline.pipeline import Pipeline, PipelineId
+from src.taipy.core.scenario._scenario_converter import _ScenarioConverter
+from src.taipy.core.scenario.scenario import Scenario, ScenarioId, _ScenarioModel
+from src.taipy.core.task._task_manager_factory import _TaskManagerFactory
+from src.taipy.core.task.task import Task, TaskId
+from taipy.config.common.scope import Scope
+
+
+def test_preserve_tasks_and_data_nodes_old_scenario_with_pipeline():
```
jrobinAV, line 26: Can we split this test.
It s very hard to review it.

tests/core/scenario/test_scenario.py
jrobinAV, line 1: I could not review this file yet. There are a lot of changes. I will have more time tomorrow to do it.

src/taipy/core/config/checkers/_scenario_config_checker.py
toan-quach: it won't grow too much and I have actually already added it before seeing your comment...

src/taipy/core/config/checkers/_scenario_config_checker.py
toan-quach: well... I don't think it will hurt the execution or anything, does feel weird to have an input or output dn as an additional dn and vice versa though. I doesn't represent correctly what we have in mind for additional dn so I put it as an error. What do you think?

src/taipy/core/scenario/_scenario_manager.py
toan-quach: Absolutely! Thanks for pointing this out I almost forgot to raise this. So yes we should have a function that can create a Pipeline from a list of Tasks or TaskIds, the current implementation using PipelineConfig is just to reuse what's already available without causing even more changes, it's not the ideal way by any means!

tests/core/notification/test_core_event_consumer.py
toan-quach: I believe it won't be necessary to have both :D

tests/core/scenario/test_scenario_converter.py
toan-quach: It's mostly a migration of tests from `test_scenario.py`. I agree it's quite a lot, I will see what I can do :D

src/taipy/core/config/checkers/_scenario_config_checker.py
jrobinAV: OK no pb

src/taipy/core/config/checkers/_scenario_config_checker.py
jrobinAV: Raising an error prevents the user from running its application. Since the pb does not hurt the execution, I would use the WARNING level.

src/taipy/core/scenario/_scenario_manager.py
jrobinAV: Now we have a ticket. We won't forget. 😄

src/taipy/core/config/checkers/_scenario_config_checker.py
toan-quach: that makes perfect sense! I'll update that :D

tests/core/data/test_data_manager.py
toan-quach: not sure, I didn't even notice I removed that line...

jrobinAV: In any case we have a ticket to investigate it.

jrobinAV (dismissed):

jrobinAV (dismissed):

src/taipy/core/scenario/scenario.py
toan-quach: #695

tests/core/_orchestrator/test_orchestrator.py
toan-quach: ah I suspect it was before the time we decided not to remove Pipeline entity anymore, I have also uncommented this test so it should be available now :D

tests/core/cycle/test_cycle_manager.py
toan-quach: Yep it was on purpose :D by adding task_config_2, we can also see how dn_2 (Scope.Scenario) behaves in different scenario configs :D task_3 only has 1 scope.cycle and 1 scope.global dn

jrobinAV (dismissed):

tests/core/test_core_cli.py
```diff
@@ -165,14 +165,14 @@ def test_dev_mode_clean_all_entities_of_the_latest_version():
     # Assert number of entities of the latest version only
     assert len(_DataManager._get_all(version_number="latest")) == 2
     assert len(_TaskManager._get_all(version_number="latest")) == 1
-    assert len(_PipelineManager._get_all(version_number="latest")) == 1
+    assert len(_PipelineManager._get_all(version_number="all")) == 2
```
trgiangdo:
```suggestion
    assert len(_PipelineManager._get_all(version_number="latest")) == 2
```
If `version_number="latest"`, how many number of pipeline entities exist?

tests/core/test_core_cli.py
```diff
@@ -165,14 +165,14 @@ def test_dev_mode_clean_all_entities_of_the_latest_version():
...
     assert len(_ScenarioManager._get_all(version_number="latest")) == 1
     assert len(_JobManager._get_all(version_number="latest")) == 1
 
     # Assert number of entities of the development version only
     assert len(_DataManager._get_all(version_number="development")) == 2
     assert len(_TaskManager._get_all(version_number="development")) == 1
-    assert len(_PipelineManager._get_all(version_number="development")) == 1
+    assert len(_PipelineManager._get_all(version_number="all")) == 2
```
trgiangdo: same question here

tests/core/test_core_cli.py
```diff
@@ -319,7 +318,7 @@ def test_production_mode_load_all_entities_from_previous_production_version():
     scenario = _ScenarioManager._create(scenario_config)
     _ScenarioManager._submit(scenario)
 
-    assert len(_DataManager._get_all()) == 4
+    assert len(_DataManager._get_all()) == 6
```
trgiangdo: This seems like a bug to me. I will check it again.

tests/core/version/test_production_version_migration.py
```diff
@@ -246,7 +245,7 @@ def test_migrate_compatible_version():
         _ScenarioManager._submit(scenario_v2)
 
         assert scenario_v2.d2.read() == 2
-        assert len(_DataManager._get_all(version_number="all")) == 4
+        assert len(_DataManager._get_all(version_number="all")) == 6
```
trgiangdo: I'm not sure why would the number of data nodes increases here. I will investigate

tests/core/version/test_production_version_migration.py
```diff
@@ -264,7 +263,7 @@ def test_migrate_compatible_version():
         _ScenarioManager._submit(scenario_v2_1)
 
     assert scenario_v2_1.d2.read() == 6
-    assert len(_DataManager._get_all(version_number="all")) == 6
+    assert len(_DataManager._get_all(version_number="all")) == 10
```
trgiangdo: I'm not sure why would the number of data nodes increases here. I will investigate

trgiangdo (changes requested):
There are some weird behavior of the versioning tests.

I will block the merge of this PR for the time being.

trgiangdo (dismissed):

jrobinAV (approved):

PR #639, opened 2023-06-13
Bug/add missing clean method

toan-quach:
Fix:
- The latest merge from develop to dev/3.0 has caused broken unit tests due to lack of clean method in migration config

### DIFF
--- a/tests/core/_backup/test_backup.py
+++ b/tests/core/_backup/test_backup.py
@@ -63,7 +63,7 @@ def test_no_new_entry_when_file_is_in_storage_folder():
 
     backup_files = read_backup_file(backup_file_path)
     assert backup_files == [f"{dn_1.path}\n"]
-    
+
     os.remove("dn_1.pickle")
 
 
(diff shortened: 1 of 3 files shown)


trgiangdo (approved):
LGTM! Thank you

trgiangdo:
And also, there are quite a few linting errors.
But I think those are not caused by this PR but are on `dev/3.0` branch.

toan-quach:
@trgiangdo 
Yep you're right. What I'm thinking is that, we merge this PR first so all the tests are passing and deal with the linter in a separate PR. The amount of complains from the linter is huge and non-related to this PR so here wouldn't be the right place to fix it.

tests/core/_backup/test_backup.py
```diff
@@ -64,6 +64,8 @@ def test_no_new_entry_when_file_is_in_storage_folder():
     backup_files = read_backup_file(backup_file_path)
     assert backup_files == [f"{dn_1.path}\n"]
 
+    os.remove("dn_1.pickle")
```
jrobinAV, line 67: 👍🏻

jrobinAV (approved):

jrobinAV:
#642 I created this ticket for solving the linter issues so we can merge this PR.

PR #645, opened 2023-06-14
bug/ added delay publishing event when entity is in context manager

toan-quach:
The property class still posts some pain when updating it in context manager. I tried many different ways but the problem lies down to the object references and the fact that we're storing `_is_in_context` (the attribute used to determine the existence of a context manager) in the memory. And as we have multiple places where we would reload the entity, this `_is_in_context` can be affected. I have checked multiple location and it seems that we're good (fingers-crossed), aside for places like properties.

### DIFF
--- a/src/taipy/core/_entity/_properties.py
+++ b/src/taipy/core/_entity/_properties.py
@@ -18,7 +18,6 @@ class _Properties(UserDict):
     def __init__(self, entity_owner, **kwargs):
         super().__init__(**kwargs)
         self._entity_owner = entity_owner
-        self._current_in_context_entity_owner = None
 
     def __setitem__(self, key, value):
         super(_Properties, self).__setitem__(key, value)
--- a/src/taipy/core/_entity/_properties.py
+++ b/src/taipy/core/_entity/_properties.py
@@ -18,18 +18,35 @@ class _Properties(UserDict):
     def __init__(self, entity_owner, **kwargs):
         super().__init__(**kwargs)
         self._entity_owner = entity_owner
+        self._current_in_context_entity_owner = None
 
     def __setitem__(self, key, value):
         super(_Properties, self).__setitem__(key, value)
         from ... import core as tp
 
-        if hasattr(self, "_entity_owner"):
-            tp.set(self._entity_owner)
-            _publish_event(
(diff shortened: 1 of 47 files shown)


jrobinAV (changes requested):
I think we can merge this part and create a new PR for the TODOS. 
However, I prefer to block it until the 2.3.0 is out.

toan-quach:
@jrobinAV Agree! 😃

src/taipy/core/_entity/_properties.py
```diff
@@ -18,19 +18,24 @@ class _Properties(UserDict):
     def __init__(self, entity_owner, **kwargs):
         super().__init__(**kwargs)
         self._entity_owner = entity_owner
+        self._current_in_context_entity_owner = None
```
jrobinAV: Should it be part of the other PR?

src/taipy/core/cycle/cycle.py
```diff
@@ -132,6 +132,7 @@ def name(self, val):
     @property
     def properties(self):
         self._properties = _reload(self._MANAGER_NAME, self)._properties
+        self._properties._entity_owner = self
```
jrobinAV: I don't like this. It looks like a code smell to me.

src/taipy/core/_entity/_properties.py
toan-quach: oh my bad yes!!

src/taipy/core/cycle/cycle.py
toan-quach: I won't argue with that 😅 but other than that, I'm not sure how to solve object references problem

trgiangdo: From what I saw, all the
```python
@property
    def properties(self):
```
are the same for all entities.
I think it makes sense to move it into the _Entity class.

And about the `_entity_owner` reference, maybe it is cleaner if we can do something inside the `_reload()` method. But I'm not sure if it is possible or not.

jrobinAV (approved):

jrobinAV (approved):
Sounds good to me

trgiangdo (approved):

Issue #648, opened 2023-06-18
BUG: the parameter --taipy-force is broken


jrobinAV:
The  --taipy-force option does not work.

``` bash
jeanrobin@LAPTOP-RPTO2SFK MINGW64 ~/workspace/demos/scenario
$ python main.py --production 2.0 --taipy-force
[2023-06-18 17:00:06][Taipy][INFO] No scenario comparators defined for ScenarioConfig `scenario_configuration`.
[2023-06-18 17:00:06][Taipy][INFO] There are non-conflicting changes between the version 1.0 Configuration and the current Configuration:
        CORE "version_number" was modified: 1.0 -> 2.0
        CORE "force" was modified: False:bool -> True:bool
[2023-06-18 17:00:06][Taipy][ERROR] The version 1.0 Configuration is conflicted with the current Configuration:
        DATA_NODE "replacement_type" has attribute "default_value" added: DEFAULT
[2023-06-18 17:00:06][Taipy][ERROR] To force running the application with the changes, run your application with --taipy-force option.
The application is stopped. Please check the error log for more information.
```

- First the "force" parameter should not appear in the config changes.
- Second, I should be able to run passing the --taipy-force parameter

trgiangdo:
1. The "force" parameter appears under INFO section, which means that it is a "non-conflicting change". If I get your message correctly, you want to keep the "version_number" change in the INFO, but remove "force" right? That would require an exception since both are on the CoreSection.
2. My guess is that the --taipy-force is failing here because the production version 1.0 is conflict with the current Configuration. On 2.3, we do not provide any migration method yet, so it is supposed to fail.

trgiangdo:
To resolve the second point, I added 2 PRs to improve the error message. It will be something like
```console
$ python3 main.py --production "2.0"
[2023-06-21 10:41:57][Taipy][INFO] No scenario comparators defined for ScenarioConfig `s`.
[2023-06-21 10:41:57][Taipy][INFO] There are non-conflicting changes between the version 1.0 Configuration and the current Configuration:
        CORE "mode" was modified: experiment -> production
        CORE "version_number" was modified: 1.0 -> 2.0
[2023-06-21 10:41:57][Taipy][ERROR] The version 1.0 Configuration is conflicted with the current Configuration:
        DATA_NODE "b" has attribute "default_data" added: b
[2023-06-21 10:41:57][Taipy][ERROR] Please remove version 1.0 from production to avoid conflict.
```

[...]

PR #649, opened 2023-06-21
Fix - Misleading conflict error message on production version

trgiangdo:
https://github.com/Avaiga/taipy-core/issues/648

### DIFF
--- a/src/taipy/core/_version/_version_manager.py
+++ b/src/taipy/core/_version/_version_manager.py
@@ -54,7 +54,9 @@ class _VersionManager(_Manager[_Version]):
             comparator_result = Config._comparator._find_conflict_config(version.config, Config._applied_config, id)
             if comparator_result.get(_ComparatorResult.CONFLICTED_SECTION_KEY):
                 if force:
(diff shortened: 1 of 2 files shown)


jrobinAV (dismissed):

src/taipy/core/_version/_version_manager.py
```diff
@@ -170,7 +184,7 @@ def _replace_version_number(cls, version_number: Optional[str]):
             return cls._get_latest_version()
         if version_number in cls.__DEVELOPMENT_VERSION:
             return cls._get_development_version()
-        if version_number in cls.__PRODUCTION_VERSION:
+        if version_number == cls.__PRODUCTION_VERSION:
```
jrobinAV, line 187: why ?

jrobinAV (approved):

src/taipy/core/_version/_version_manager.py
trgiangdo: Both are string. I fix it in dev/3.0, but in 2.3 we are not using it yet, so there is no problem yet

PR #653, opened 2023-06-22
feature/#161 added is submittable to primary scenario in scenario manager

toan-quach:
https://app.zenhub.com/workspaces/taipy-613ef328affde6000ff77fb3/issues/gh/avaiga/taipy-enterprise/161

### DIFF
--- a/src/taipy/core/pipeline/_pipeline_manager.py
+++ b/src/taipy/core/pipeline/_pipeline_manager.py
@@ -129,6 +129,12 @@ class _PipelineManager(_Manager[Pipeline]):
             task_manager._set(i)
 
     @classmethod
+    def _is_submittable(cls, pipeline: Union[Pipeline, PipelineId]) -> bool:
+        if isinstance(pipeline, str):
+            pipeline = cls._get(pipeline)
+        return True if isinstance(pipeline, Pipeline) else False
+
+    @classmethod
     def _submit(
         cls,
         pipeline: Union[PipelineId, Pipeline],
--- a/src/taipy/core/pipeline/_pipeline_manager.py
+++ b/src/taipy/core/pipeline/_pipeline_manager.py
@@ -132,7 +132,7 @@ class _PipelineManager(_Manager[Pipeline]):
     def _is_submittable(cls, pipeline: Union[Pipeline, PipelineId]) -> bool:
         if isinstance(pipeline, str):
             pipeline = cls._get(pipeline)
-        return True if isinstance(pipeline, Pipeline) else False
+        return isinstance(pipeline, Pipeline)
 
     @classmethod
     def _submit(
--- a/src/taipy/core/scenario/_scenario_manager.py
+++ b/src/taipy/core/scenario/_scenario_manager.py
@@ -143,7 +143,7 @@ class _ScenarioManager(_Manager[Scenario]):
     def _is_submittable(cls, scenario: Union[Scenario, ScenarioId]) -> bool:
         if isinstance(scenario, str):
             scenario = cls._get(scenario)
-        return True if isinstance(scenario, Scenario) else False
+        return isinstance(scenario, Scenario)
 
     @classmethod
     def _submit(
--- a/src/taipy/core/scenario/_scenario_manager.py
+++ b/src/taipy/core/scenario/_scenario_manager.py
@@ -140,6 +140,12 @@ class _ScenarioManager(_Manager[Scenario]):
             pipeline_manager._set(i)
 
     @classmethod
+    def _is_submittable(cls, scenario: Union[Scenario, ScenarioId]) -> bool:
+        if isinstance(scenario, str):
+            scenario = cls._get(scenario)
+        return True if isinstance(scenario, Scenario) else False
+
+    @classmethod
     def _submit(
         cls,
         scenario: Union[Scenario, ScenarioId],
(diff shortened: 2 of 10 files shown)


src/taipy/core/taipy.py
```diff
@@ -64,6 +64,20 @@ def set(entity: Union[DataNode, Task, Pipeline, Scenario, Cycle]):
...
+    """Indicate if an entity can be submitted.
+    Returns:
+        True if the given entity can be submitted. False otherwise.
+    """
+    if isinstance(entity, Scenario) or (isinstance(entity, str) and entity.startswith(Scenario._ID_PREFIX)):
+        return _ScenarioManagerFactory._build_manager()._is_submittable(entity)  # type: ignore
+    if isinstance(entity, Pipeline) or (isinstance(entity, str) and entity.startswith(Pipeline._ID_PREFIX)):
+        return _PipelineManagerFactory._build_manager()._is_submittable(entity)  # type: ignore
+    if isinstance(entity, Task) or (isinstance(entity, str) and entity.startswith(Task._ID_PREFIX)):
+        return _TaskManagerFactory._build_manager()._is_submittable(entity)  # type: ignore
```
trgiangdo, line 78: why do we need the `# type: ignore` in this section?

src/taipy/core/taipy.py
```diff
@@ -64,6 +64,20 @@ def set(entity: Union[DataNode, Task, Pipeline, Scenario, Cycle]):
         return _DataManagerFactory._build_manager()._set(entity)
 
 
+def is_submittable(entity: Union[Scenario, ScenarioId, Pipeline, PipelineId, Task, TaskId]) -> bool:
+    """Indicate if an entity can be submitted.
+    Returns:
```
trgiangdo:
```suggestion
    """Indicate if an entity can be submitted.

    Returns:
```
I think in the convention, there should be an empty line here.

src/taipy/core/pipeline/_pipeline_manager.py
```diff
@@ -128,6 +128,12 @@ def __save_tasks(cls, tasks):
         for i in tasks:
             task_manager._set(i)
 
+    @classmethod
+    def _is_submittable(cls, pipeline: Union[Pipeline, PipelineId]) -> bool:
+        if isinstance(pipeline, str):
+            pipeline = cls._get(pipeline)
+        return True if isinstance(pipeline, Pipeline) else False
```
trgiangdo:
```suggestion
        return isinstance(pipeline, Pipeline)
```
This should be enough, right?

src/taipy/core/scenario/_scenario_manager.py
```diff
@@ -139,6 +139,12 @@ def __save_pipelines(cls, pipelines):
         for i in pipelines:
             pipeline_manager._set(i)
 
+    @classmethod
+    def _is_submittable(cls, scenario: Union[Scenario, ScenarioId]) -> bool:
+        if isinstance(scenario, str):
+            scenario = cls._get(scenario)
+        return True if isinstance(scenario, Scenario) else False
```
trgiangdo:
```suggestion
        return isinstance(scenario, Scenario)
```
Same comment as above

src/taipy/core/task/_task_manager.py
```diff
@@ -140,6 +140,12 @@ def _get_children_entity_ids(cls, task: Task):
                 entity_ids.job_ids.add(job.id)
         return entity_ids
 
+    @classmethod
+    def _is_submittable(cls, task: Union[Task, TaskId]) -> bool:
+        if isinstance(task, str):
+            task = cls._get(task)
+        return True if isinstance(task, Task) else False
```
trgiangdo:
```suggestion
        return isinstance(task, Task)
```
Same comment as above

tests/core/pipeline/test_pipeline_manager.py
```diff
@@ -119,6 +119,16 @@ def test_set_and_get_pipeline():
     assert _TaskManager._get(task_2.id).id == task_2.id
 
 
+def test_is_submittable():
+    assert len(_PipelineManager._get_all()) == 0
+    pipeline_config = Config.configure_pipeline("pipeline", [])
+    pipeline = _PipelineManager._get_or_create(pipeline_config)
+
+    assert len(_PipelineManager._get_all()) == 1
+    assert _PipelineManager._is_submittable(pipeline)
```
trgiangdo, line 128: Can we test using the PipelineId here as well?
Something like `assert _PipelineManager._is_submittable("PIPELINE_pipeline...)`

trgiangdo (commented):
I have some comments.

tests/core/pipeline/test_pipeline_manager.py
jrobinAV: assert _PipelineManager._is_submittable(pipeline.id)

jrobinAV (dismissed):
Agreed with Giang remarks

src/taipy/core/taipy.py
toan-quach: Ah because the signature of is_submittable is entity and it can be Task, Scenario or Pipeline, so it doesn't match fully with _is_submittable for the managers :D

trgiangdo (dismissed):

trgiangdo (approved):

jrobinAV (approved):

PR #654, opened 2023-06-23
 bug/ fix properties not delay set in context manager

toan-quach:
TODO:
- [x] Added unit tests to capture all expected cases
- [x] Solve failing cases


jrobinAV (changes requested):
Looks exhaustive

jrobinAV (dismissed):

src/taipy/core/_entity/_reload.py
```diff
@@ -71,7 +72,9 @@ def _do_set_entity(self, *args, **kwargs):
             ]
 
             if not self._is_in_context:
-                entity_manager._set(self)
+                entity = entity_manager._get(self, self)
```
toan-quach, line 75: the purpose of this snippet is to solve the scenario where we have 2 Tasks t1 and t2:
```
assert task_1.function == print
assert task_2.function == print

task_1.function = sum
assert task_1.function == sum
assert task_2.function == sum

task_2.function = mock_func
assert task_1.function == mock_func
assert task_2.function == mock_func

assert not task_1.skippable
assert not task_2.skippable

task_1.skippable = True
assert task_1.skippable
assert task_2.skippable

task_2.skippable = False
assert not task_1.skippable
assert not task_2.skippable

assert task_1.function == mock_func
assert task_2.function == mock_func
```
Why does this matter:
Let say the initial value of `function` for both task_1 and task_2 is print
When setting task_1.function = sum => assert task_1._function == sum is True
But task_2._function == print is still True
And if we set task_2.skippable = False => auto_setter will auto set task_2 where task_2._function == print
==>> the Task entity in the repository will have its function go from print -> sum -> print

src/taipy/core/_entity/_entity.py
```diff
@@ -30,6 +30,9 @@ def __exit__(self, exc_type, exc_value, exc_traceback):
         self._is_in_context = False
         while self._in_context_attributes_changed_collector:
             _publish_event(*self._in_context_attributes_changed_collector.pop(0))
+        if hasattr(self, "_properties"):
+            self._properties.data = self._properties._previous_data
+
```
jrobinAV: Don't we want to publish event after setting the properties and the entity?

src/taipy/core/_entity/_properties.py
```diff
@@ -18,6 +18,7 @@ class _Properties(UserDict):
     def __init__(self, entity_owner, **kwargs):
         super().__init__(**kwargs)
         self._entity_owner = entity_owner
+        self._previous_data = {}
```
jrobinAV: Why previous data?

This corresponds to the new values to be set at the context exit.
why not naw_data or pending_changes, or something else.

src/taipy/core/_entity/_reload.py
```diff
@@ -71,7 +72,9 @@ def _do_set_entity(self, *args, **kwargs):
             ]
 
             if not self._is_in_context:
-                entity_manager._set(self)
+                entity = _Reloader()._reload(manager, self)
+                fct(entity, *args, **kwargs)
+                entity_manager._set(entity)
```
toan-quach, line 75: the purpose of this snippet is to solve the scenario where we have 2 Tasks t1 and t2:
```
assert task_1.function == print
assert task_2.function == print

task_1.function = sum
assert task_1.function == sum
assert task_2.function == sum

task_2.function = mock_func
assert task_1.function == mock_func
assert task_2.function == mock_func

assert not task_1.skippable
assert not task_2.skippable

task_1.skippable = True
assert task_1.skippable
assert task_2.skippable

task_2.skippable = False
assert not task_1.skippable
assert not task_2.skippable

assert task_1.function == mock_func
assert task_2.function == mock_func
```
Why does this matter:
Let say the initial value of `function` for both task_1 and task_2 is print
When setting task_1.function = sum => assert task_1._function == sum is True
But task_2._function == print is still True
And if we set task_2.skippable = False => auto_setter will auto set task_2 where task_2._function == print
==>> the Task entity in the repository will have its function go from print -> sum -> print

And so instead of setting the current "`self`", we can load the "up-to-date" value from the repo as an entity, apply the changes to the new entity and set it instead of `self`

src/taipy/core/data/abstract_sql.py
```diff
@@ -120,7 +120,7 @@ def __init__(
         )
         self._engine = None
         if not self._last_edit_date:
-            self.last_edit_date = datetime.now()  # type: ignore
+            self._last_edit_date = datetime.now()
```
jrobinAV: So we don't auto-set the entity anymore after this change. Since it s on the __init__ method, I don't understand the impact of this change. Any clue?

src/taipy/core/_entity/_reload.py
```diff
@@ -71,7 +72,9 @@ def _do_set_entity(self, *args, **kwargs):
             ]
 
             if not self._is_in_context:
-                entity_manager._set(self)
+                entity = _Reloader()._reload(manager, self)
+                fct(entity, *args, **kwargs)
```
jrobinAV, line 74: fct is applied twice. We should remove the previous call and move it to the else block.

src/taipy/core/_entity/_entity.py
toan-quach: you're right!!

src/taipy/core/_entity/_properties.py
toan-quach: I was thinking in the term of data from previous entity but since the purpose is to delay changes in context, `pending_changes` would make much more sense :D

src/taipy/core/data/abstract_sql.py
toan-quach: ah yep, without the underscore, this line will trigger _self_setter, we have just created the entity without setting it, and to trigger _self_setter before even the manager does that is kinda strange no? Also it caused some recursion problem :D

src/taipy/core/_entity/_reload.py
toan-quach: it's actually applied to different entities, the 1st fct called on line 64 applied to self `fct(self, *args, **kwargs)`, the 2nd on line 76 applied to the new up-to-date entity `fct(entity, *args, **kwargs)`

jrobinAV: YEs but the first one is useless if we set the second one, right?

toan-quach: hmmm I believe we still need the 1st one to set the underscore value of the current entity, but what you said make sense too since we're saving directly to the repo, whatever is stored in the memory is now obsolete?

toan-quach: so I tried putting the 1st fct called to the else block, and some tests failed :D apparently we do use _ attributes in some places and it needs to be updated

jrobinAV: Can we have an overview of all the _attributes of an entity used outside the entity itself?
That should clarify the impact of such change.

jrobinAV (dismissed):

jrobinAV (approved):

Issue #641, opened 2023-06-13
Store the submitted entity id at job's creation


jrobinAV:
The purpose is to be able to retrieve what is the submitted entity that triggered the job s creation and submission.

Will be used in the job selector gui component.

PR #658, opened 2023-06-26
feature/#641 added submit_entity_id to job entity

toan-quach:
#641


src/taipy/core/_orchestrator/_orchestrator.py
```diff
@@ -72,7 +72,7 @@ def submit(
         with cls.lock:
             for ts in tasks:
                 for task in ts:
-                    res.append(cls._submit_task(task, submit_id, callbacks=callbacks, force=force))
+                    res.append(cls._submit_task(task, submit_id, submittable.id, callbacks=callbacks, force=force))  # type: ignore
```
jrobinAV: Why type ignore?

src/taipy/core/job/_job_converter.py
```diff
@@ -42,7 +43,11 @@ def _model_to_entity(cls, model: _JobModel) -> Job:
         task_repository = task_manager._repository
 
         job = Job(
-            id=model.id, task=task_repository._load(model.task_id), submit_id=model.submit_id, version=model.version
+            id=model.id,
+            task=task_repository._load(model.task_id),
+            submit_id=model.submit_id,
+            submit_entity_id=model.submit_entity_id,
```
jrobinAV, line 49: Maybe we can hold the `submit_entity` in the `Job` and only store the `submit_entity_id` in the model. Exactly like we did for the job.task.

toan-quach: I believe because the `id` attribute isn't available in the Submittable class

src/taipy/core/_orchestrator/_orchestrator.py
toan-quach: I believe because the id attribute isn't available in the Submittable class

src/taipy/core/job/_job_converter.py
toan-quach: sure I don't see why not :D

src/taipy/core/_orchestrator/_orchestrator.py
jrobinAV: Oh probably. Thx.

Issue #665, opened 2023-07-02
Remove useless info log message


jrobinAV:
- [x] Remove the following log
[2023-07-02 20:42:21][Taipy][INFO] No scenario comparators defined for ScenarioConfig `scenario_configuration`.
- [x] Update checker documentation in user manual

PR #670, opened 2023-07-06
Fix/#665 - Improve the scenario comparator checker

trgiangdo:
https://github.com/Avaiga/taipy-core/issues/665

Beside removing the "[INFO] No scenario comparators" log, this PR also improves the checker.

The scenario comparator checker now check:
- If the comparators field is a dictionary or not.
- If keys in the comparators dictionary are valid data node ids.
- If values in the comparators dictionary are callable objects, or lists or callable objects.


jrobinAV (approved):
neat!

Issue #655, opened 2023-06-23
Create exists API


jrobinAV:
Expose an `exists` method to check if an entity id exists.

Expected signature: `def tp.exists(entityId) -> bool:`

For performance reasons, the method should not load the entity but just check if it exists.

- [x] create the method in _Managers
- [x] expose it in taipy.core.taipy
- [x] expose it in taipy._init.py
- [ ] implement the enterprise version checking the roles
- [x] add doc string for ref manual
- [x] add it to taipy-doc in FORCE_PACKAGE list in `tools/_setup_generation/step_refman.py`
- [x] add new entry in release notes

PR #673, opened 2023-07-07
Feature/#655 - Create exists API

trgiangdo:
https://github.com/Avaiga/taipy-core/issues/655

There are a lot of added test cases, but most of them are similar to each other.

### DIFF
--- a/src/taipy/core/_repository/_sql_repository.py
+++ b/src/taipy/core/_repository/_sql_repository.py
@@ -52,7 +52,7 @@ class _SQLRepository(_AbstractRepository[ModelType, Entity]):
         self.__insert_model(obj)
 
     def _exists(self, entity_id: str):
-        return bool(self.db.query(self.model_type).filter_by(id=entity_id).first())
+        return bool(self.db.query(self.model_type.id).filter_by(id=entity_id).first())  # type: ignore
 
     def _load(self, entity_id: str) -> Entity:
         if entry := self.db.query(self.model_type).filter(self.model_type.id == entity_id).first():  # type: ignore
--- a/src/taipy/core/_repository/_sql_repository.py
+++ b/src/taipy/core/_repository/_sql_repository.py
@@ -51,7 +51,10 @@ class _SQLRepository(_AbstractRepository[ModelType, Entity]):
             return
         self.__insert_model(obj)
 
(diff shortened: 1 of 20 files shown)


src/taipy/core/_repository/_sql_repository.py
```diff
@@ -51,7 +51,10 @@ def _save(self, entity: Entity):
             return
         self.__insert_model(obj)
 
-    def _load(self, entity_id: Any) -> Entity:
+    def _exists(self, entity_id: str):
+        return bool(self.db.query(self.model_type).filter_by(id=entity_id).first())
```
jrobinAV: Instead of using a filter method, did you try to use the exists method? It should be faster since it s designed for that.
https://docs.sqlalchemy.org/en/14/orm/query.html#sqlalchemy.orm.Query.exists

jrobinAV (changes requested):

src/taipy/core/_repository/_sql_repository.py
trgiangdo: In https://stackoverflow.com/questions/32938475/flask-sqlalchemy-check-if-row-exists-in-table, there is also an example that are using `exists()` method.

However, that requires a subquery, which (slightly) worsen the performance.
It goes something like this
```suggestion
        return self.db.query(
            self.db.query(self.model_type).filter_by(id=entity_id).exists()
        ).scalar()
```

trgiangdo: If we want to improve the performance a bit, instead of querying the entire model, we can only query the id.
```suggestion
        return bool(self.db.query(self.model_type.id).filter_by(id=entity_id).first()) # type: ignore
```

jrobinAV: And what about just that ?
```
return self.db.query(self.model_type.id).filter_by(id=entity_id).exists()
```

trgiangdo: That will returns a query. To execute that query, we need to wrap it inside another `self.db.query(...)`.

trgiangdo: I'm not an expert in SQL, so maybe we can have a third opinion about this.

jrobinAV (approved):

PR #674, opened 2023-07-08
Fix/ Inconsistent failing CLI test on SQL repo

trgiangdo:
The problem was that in `test_modify_job_configuration_dont_stop_application()` test, child processes open SQL session.
And on the next `test_modify_config_properties_without_force()` test, trying to execute an SQL query sometimes can fail because the child processes are not killed yet.

In this PR:
- `test_modify_job_configuration_dont_stop_application()` is moved to be the last test.

### DIFF
--- a/tests/core/test_core_cli_with_sql_repo.py
+++ b/tests/core/test_core_cli_with_sql_repo.py
@@ -551,10 +551,12 @@ def test_modify_job_configuration_dont_stop_application(caplog, tmp_sqlite):
 
     with patch("sys.argv", ["prog", "--experiment", "1.0"]):
         Config.configure_job_executions(mode="development")
-        Core().run(force_restart=True)
+        core = Core()
+        core.run(force_restart=True)
     scenario = _ScenarioManager._create(scenario_config)
     jobs = _ScenarioManager._submit(scenario)
     assert all([job.is_finished() for job in jobs])
+    core.stop()
 
     init_config()
     Config.configure_global_app(repository_type="sql", repository_properties={"db_location": tmp_sqlite})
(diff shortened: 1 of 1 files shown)


tests/core/test_core_cli_with_sql_repo.py
```diff
@@ -573,6 +544,35 @@ def test_modify_config_properties_without_force(caplog, tmp_sqlite):
...
+    assert all([job.is_finished() for job in jobs])
+
+    init_config()
+    Config.configure_global_app(repository_type="sql", repository_properties={"db_location": tmp_sqlite})
+
+    scenario_config = config_scenario()
+
+    with patch("sys.argv", ["prog", "--experiment", "1.0"]):
+        Config.configure_job_executions(mode="standalone", max_nb_of_workers=2)
+        Core().run(force_restart=True)
```
jrobinAV:
```suggestion
        core = Core()
        core.run(force_restart=True)
```

tests/core/test_core_cli_with_sql_repo.py
```diff
@@ -573,6 +544,35 @@ def test_modify_config_properties_without_force(caplog, tmp_sqlite):
...
+    scenario = _ScenarioManager._create(scenario_config)
+
+    jobs = _ScenarioManager._submit(scenario)
+    assert_true_after_time(lambda: all(job.is_finished() for job in jobs))
+    error_message = str(caplog.text)
+    assert 'JOB "mode" was modified' in error_message
+    assert 'JOB "max_nb_of_workers" was modified' in error_message
+
```
jrobinAV, line 579:
```suggestion
core.stop()
```
this could help as well, no?
At the end of the test, we can even assert that dispatcher.is_running is False.

jrobinAV (dismissed):

tests/core/test_core_cli_with_sql_repo.py
trgiangdo: True, it just that it's not the main cause.

jrobinAV (approved):

Issue #662, opened 2023-06-28
ensure methods have the right visibility on converters


jrobinAV:
Some methods are hidden but could be private. We need to verify:
_deserialize_sql_dn_model_properties, _serialize_edits, and many more

trgiangdo:
In `src/taipy/core/data/_data_converter.py`:
- `_serialize_sql_dn_properties()` should be private
- `_serialize_mongo_collection_dn_model_properties()` should be private
- `_serialize_edits()` should be private
- `_deserialize_sql_dn_model_properties()` should be private
- `_deserialize_sql_dn_model_properties` should be private
- `_deserialize_edits` should be private

In `src/taipy/core/data/_abstract_tabular.py`:
- `_serialize_exposed_type()` is not used anywhere
- `_deserialize_exposed_type()` is not used anywhere

In `src/taipy/core/data/data_node.py`:
- `_serialize_datanode_properties()` should be removed
- `_deserialize_datanode_model_properties()` should be removed

In `src/taipy/core/job/_job_converter.py` and `src/taipy/core/job/job.py`:
- `_serialize_subscribers()` should be private

jrobinAV:
Sounds good. I suspect the unused methods were used before the repository refactoring. @joaoandre Do you know if I am right?

PR #675, opened 2023-07-10
Feature/#662 - Ensure some methods have the right visibility on converters

trgiangdo:
https://github.com/Avaiga/taipy-core/issues/662

In this PR:
- `_abstract_file.py` and `_abstract_sql.py` are now internal modules.
- `_serialize_exposed_type()` and `_deserialize_exposed_type()` are now moved to the DataConverter since they are only used here.
- Some methods in job converter and data converter are now private.


toan-quach (approved):

src/taipy/core/data/_abstract_file.py
jrobinAV, line 1: On this documentation 
[page]https://docs.taipy.io/en/develop/manuals/reference/taipy.core.data.CSVDataNode/),
The `_AbstractFileDataNode` is referred to. I am unsure, but renaming its module from `abstract_file` to `_abstract_file` may break the doc. Can you test it?

src/taipy/core/data/_abstract_sql.py
jrobinAV, line 1: same comment here abt renaming it adding _

src/taipy/core/data/_abstract_file.py
trgiangdo: The link in the doc page is already not working because _AbstractFileDataNode is internal class.
This change will not affect it

jrobinAV (approved):

Issue #676, opened 2023-07-10
BUG- `taipy.get_...()` methods return entities from all versions


trgiangdo:
**Description**

For now, all`taipy.get_...()` methods return a list of entities from all version.
This is not good when we have a lot of versions.

The desired behavior:
- In development mode and experiment mode, the user should only be able to get entities from the current version.
- In production mode, the user should be able to get all entities from all production versions.

PR #678, opened 2023-07-11
Fix/#676 - tp.get methods return entities from all versions

trgiangdo:
https://github.com/Avaiga/taipy-core/issues/676

In this PR:
- The default values of "version_number" of `_get_all()` method on managers are set to "None" to get the value from the current version only.
- The "filters" in SQL and FS repositories work a bit different. For example:
  - If `filters = [{"version": "1.0"}, {"version": "2.0"}]`, the `_get_all_by()` methods will return entities where
  (version = 1.0 OR version = 2.0)
  - If `filters = [{"version": "1.0", "config_id": "dn1"}, {"version": "2.0", "config_id": "dn1"}]`, the `_get_all_by()` methods will return entities where
  ( (version "1.0" AND config_id = "dn1") OR (version "2.0" AND config_id = "dn1") )

Please review the logic carefully since there maybe side effect.


tests/core/data/test_data_manager.py
```diff
@@ -251,6 +251,31 @@ def test_get_all(self):
...
+                _DataManager._set(
+                    InMemoryDataNode(
+                        f"config_id_{i+version}",
+                        Scope.SCENARIO,
+                        id=DataNodeId(f"id{i}_v{version}"),
+                        version=f"{version}.0",
+                    )
+                )
+
+        _VersionManager._set_development_version("1.0")
```
jrobinAV, line 274: I don't get why we set a development version here. Don't we want to test the feature on production and experiment versions?

trgiangdo: This test is for a multiple-version environment. Experiment or development mode is not really the point here.
We have separate test cases for experiment and multiple production versions on `test_core_cli.py`.

jrobinAV (commented):
Hi, 

@trgiangdo I have plenty of questions. Can we have a call to discuss them?

jrobinAV (dismissed):

jrobinAV (approved):

Issue #672, opened 2023-07-06
add new APIs on submittable entities


jrobinAV:
- [x] Rename submittable._get_inputs to make it public (get_inputs). It returns the set of input data nodes.
- [x] Add submittable.get_outputs. It should return the set of output data nodes.
- [x] Add submittable.get_intermediates. It should return the set of data nodes that are not an input nor an output. 
- [x] Add submittable.is_ready_to_run.
```
def is_ready_to_run(self):
    return all([d.is_ready_for_reading for d in submittable.get_inputs()])
```
- [x] Add submittable.execution_in_progress method that returns the set (eventually empty) of data nodes of the submittable that are in progress.
- [x] Add documentation (user man and ref man)
- [x] Add release notes

toan-quach:
@jrobinAV 
For `submittable.execution_in_progress`, I don't think this will be feasible at least in development mode. In standalone mode, it should be doable. What do you think?

jrobinAV:
Since the development mode is synchronous, when a user calls `submittable.execution_in_progress`, no data node will be in progress. You are right. Maybe there will be some specific cases where the application stops during an execution. When it restarts, some data nodes could be in progress while no job is running. It s kind of a pathological case.

In any case, I believe we can implement it regardless of the execution mode. It should work for all cases.

toan-quach:
I believe you're right!

PR #692, opened 2023-07-20
feature/#672 added new api for submittable

toan-quach:
#672

### DIFF
--- a/src/taipy/core/_entity/_submittable.py
+++ b/src/taipy/core/_entity/_submittable.py
@@ -37,9 +37,58 @@ class _Submittable:
     ):
         raise NotImplementedError
 
-    def _get_inputs(self) -> Set[DataNode]:
-        dag = self._build_dag()
+    def __get_all_data_nodes_in_dag(self, dag: Optional[nx.DiGraph] = None) -> Set[DataNode]:
+        dag = self._build_dag() if not dag else dag
+        return {node for node in dag.nodes if isinstance(node, DataNode)}
+
+    def get_inputs(self, dag: Optional[nx.DiGraph] = None) -> Set[DataNode]:
+        """Return the set of input data nodes of the submittable entity.
+
+        Returns:
+            The set of input data nodes.
+        """
+        dag = self._build_dag() if not dag else dag
         return {node for node, degree in dict(dag.in_degree).items() if degree == 0 and isinstance(node, DataNode)}
+
+    def get_outputs(self, dag: Optional[nx.DiGraph] = None) -> Set[DataNode]:
+        """Return the set of output data nodes of the submittable entity.
+
+        Returns:
+            The set of output data nodes.
+        """
+        dag = self._build_dag() if not dag else dag
+        return {node for node, degree in dict(dag.out_degree).items() if degree == 0 and isinstance(node, DataNode)}
+
+    def get_intermediate(self) -> Set[DataNode]:
+        """Return the set of intermediate data nodes of the submittable entity.
+
+        Returns:
+            The set of intermediate data nodes.
+        """
+        dag = self._build_dag()
+        all_data_nodes_in_dag = self.__get_all_data_nodes_in_dag(dag)
+        return all_data_nodes_in_dag - self.get_inputs(dag) - self.get_outputs(dag)
+
+    def is_ready_to_run(self) -> bool:
+        """Indicate if the entity is ready to be run.
+
+        Returns:
+            True if the given entity is ready to be run. False otherwise.
+        """
+        return all(dn.is_ready_for_reading for dn in self.get_inputs())
+
+    def execution_in_progress(self) -> Set[DataNode]:
+        """Return the set of data nodes of the submittable entity that are being executed.
+
+        Returns:
+            The set of data nodes that are being executed.
+        """
+        all_data_nodes_in_dag = self.__get_all_data_nodes_in_dag()
+        data_nodes_in_progress = set()
+        for dn in all_data_nodes_in_dag:
+            if dn.edit_in_progress:
+                data_nodes_in_progress.add(dn)
+        return data_nodes_in_progress
 
     @abc.abstractmethod
     def subscribe(self, callback: Callable[[_Submittable, Job], None], params: Optional[List[Any]] = None):
--- a/src/taipy/core/_entity/submittable.py
+++ b/src/taipy/core/_entity/submittable.py
@@ -23,8 +23,16 @@ from ..task.task import Task
 from ._dag import _DAG
 
 
-class _Submittable:
-    def __init__(self, subscribers=None):
+class Submittable:
+    """Instance of an entity that can be submitted for execution.
+
+    A submittable holds functions that can be used to build the executional directed acyclic graph.
+
+    Attributes:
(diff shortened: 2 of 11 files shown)


src/taipy/core/_entity/_submittable.py
```diff
@@ -37,10 +37,59 @@ def submit(
...
+
+    def is_ready_to_run(self) -> bool:
+        """Indicate if the entity is ready to be run.
+
+        Returns:
+            True if the given entity is ready to be run. False otherwise.
+        """
+        return all(dn.is_ready_for_reading for dn in self.get_inputs())
+
+    def execution_in_progress(self) -> Set[DataNode]:
```
trgiangdo: For example, when calling `scenario.execution_in_progress`, I would expect a boolean in return.
But this methods is returning a set of data nodes.

Maybe we should find a better name. Probably `executing_data_nodes()`, `data_nodes_in_execution()`, ...

src/taipy/core/_entity/_submittable.py
```diff
@@ -37,10 +37,59 @@ def submit(
...
+    def get_outputs(self, dag: Optional[nx.DiGraph] = None) -> Set[DataNode]:
+        """Return the set of output data nodes of the submittable entity.
+
+        Returns:
+            The set of output data nodes.
+        """
+        dag = self._build_dag() if not dag else dag
+        return {node for node, degree in dict(dag.out_degree).items() if degree == 0 and isinstance(node, DataNode)}
+
+    def get_intermediate(self) -> Set[DataNode]:
```
trgiangdo, line 72:
```suggestion
    def get_intermediate(self, dag: Optional[nx.DiGraph] = None) -> Set[DataNode]:
```
Shouldn't we also accept a dag parameter here? Like get_inputs and get_output

trgiangdo (commented):
I have a few comments.

src/taipy/core/_entity/_submittable.py
```diff
@@ -37,10 +37,59 @@ def submit(
     ):
         raise NotImplementedError
 
-    def _get_inputs(self) -> Set[DataNode]:
-        dag = self._build_dag()
+    def __get_all_data_nodes_in_dag(self, dag: Optional[nx.DiGraph] = None) -> Set[DataNode]:
+        dag = self._build_dag() if not dag else dag
+        return {node for node in dag.nodes if isinstance(node, DataNode)}
+
+    def get_inputs(self, dag: Optional[nx.DiGraph] = None) -> Set[DataNode]:
```
jrobinAV: Since `get_inputs` is exposed to the end user, I don't think we should accept an `nx.DiGraph` as a parameter.

Maybe you can have something like:
``` python
def get_inputs(self) -> Set[DataNode]:
    dag = self._build_dag()
    return self.__get_inputs(dag)

 def __get_inputs(self, dag: Optional[nx.DiGraph]) -> Set[DataNode]:
    return {node for node, degree in dict(dag.in_degree).items() if degree == 0 and isinstance(node, DataNode)}
```

src/taipy/core/_entity/_submittable.py
```diff
@@ -37,10 +37,59 @@ def submit(
...
+        """Return the set of input data nodes of the submittable entity.
+
+        Returns:
+            The set of input data nodes.
+        """
+        dag = self._build_dag() if not dag else dag
         return {node for node, degree in dict(dag.in_degree).items() if degree == 0 and isinstance(node, DataNode)}
 
+    def get_outputs(self, dag: Optional[nx.DiGraph] = None) -> Set[DataNode]:
```
jrobinAV: Since `get_outputs` is exposed to the end user, I don't think we should accept an `nx.DiGraph` as a parameter.

Maybe you can have something like:

``` python
def get_outputs(self) -> Set[DataNode]:
    dag = self._build_dag()
    return self.__get_outputs(dag)

 def __get_outputs(self, dag: Optional[nx.DiGraph]) -> Set[DataNode]:
    return {node for node, degree in dict(dag.out_degree).items() if degree == 0 and isinstance(node, DataNode)}

File: src/taipy/core/_entity/_submittable.py (hunk 2 above)
Comment (jrobinAV) [reply src/taipy/core/_entity/_submittable.py:72]: Since `get_intermediate` is exposed to the end user, I don't think we should accept an `nx.DiGraph` as a parameter.

Another way of doing it is:

``` python
def get_intermediate(self) -> Set[DataNode]:
    dag = self._build_dag()
    return {node for node, degree in dict(dag.degree).items() if degree != 0 and isinstance(node, DataNode)}
```

With this implementation, you simultaneously parse the dag and build the new set. You don't need three loops (`__get_all_data_nodes_in_dag`, `__get_inputs`, and `__get_outputs`).

src/taipy/core/_entity/_submittable.py
jrobinAV: Execution refers to a task, so it s strange to return data nodes.

What about `data_nodes_being_edited`?

src/taipy/core/_entity/_submittable.py
```diff
@@ -37,10 +37,59 @@ def submit(
...
+
+        Returns:
+            The set of data nodes that are being executed.
+        """
+        all_data_nodes_in_dag = self.__get_all_data_nodes_in_dag()
+        data_nodes_in_progress = set()
+        for dn in all_data_nodes_in_dag:
+            if dn.edit_in_progress:
+                data_nodes_in_progress.add(dn)
+        return data_nodes_in_progress
```
jrobinAV:
```suggestion
        dag = self._build_dag()
        return {node for node in dag.nodes if isinstance(node, DataNode) and node.edit_in_progress}
```

With this implementation, you simultaneously parse the dag and build the new set. You don't need two loops.

src/taipy/core/_entity/_submittable.py
jrobinAV, line 1: Since we now have public APIs in this class. we need to make it public and rename it Submittable.

toan-quach: Ah hah you're right! I forgot about that!!

src/taipy/core/_entity/_submittable.py
toan-quach: I'd go with `data_node_being_edited` 😀

src/taipy/core/_entity/_submittable.py
toan-quach: this will actually get all of the data nodes...

jrobinAV: if the degree (in + out) is != from 0 ?
Maybe the degree API does not exist or is not what I am thinking about then.

toan-quach: never mind, it should be `return {node for node, degree in dict(dag.degree).items() if degree > 1 and isinstance(node, DataNode)}` :D for data nodes that have more than 1 edges connecting to it.

src/taipy/core/_entity/_submittable.py
toan-quach: aside from that, I believe `_submittable.py` -> `submittable.py`

jrobinAV (approved):

trgiangdo (approved):

Issue #688, opened 2023-07-18
BUG: set_default_data_node seems broken


jrobinAV:
We change the default data node config, changing the storage type to CSV. When instantiating a datanode without specifying the storage type, we expect to have a CSVDataNode. We do have a PickleDataNode.

The following test seems wrong:
`test_data_manager.test_create_dn_from_loaded_config_modified_default_config()`

PR #696, opened 2023-07-21
Fix/#688 - Fail test on set_default_data_node

trgiangdo:
https://github.com/Avaiga/taipy-core/issues/688

### DIFF
--- a/tests/core/data/test_data_manager.py
+++ b/tests/core/data/test_data_manager.py
@@ -571,12 +571,11 @@ class TestDataManager:
         )
         from src.taipy import core as tp
 
-        Config.override(file_config.filename)
         Config.set_default_data_node_configuration(storage_type="csv")
+        Config.override(file_config.filename)
         scenario = tp.create_scenario(Config.scenarios["my_scenario"])
 
-        # assert isinstance(scenario.input, CSVDataNode) TODO Replace the next line by the commented one.
-        assert isinstance(scenario.input, PickleDataNode)
+        assert isinstance(scenario.input, CSVDataNode)
         assert isinstance(scenario.output, InMemoryDataNode)
 
     def test_get_tasks_by_config_id(self):


tests/core/data/test_data_manager.py
```diff
@@ -571,12 +571,11 @@ def test_create_dn_from_loaded_config_modified_default_config(self):
         )
         from src.taipy import core as tp
 
-        Config.override(file_config.filename)
         Config.set_default_data_node_configuration(storage_type="csv")
+        Config.override(file_config.filename)
```
jrobinAV, line 575: Well, OK, I understand the problem. It is less important than what we thought at first. 
Your change exhibits the issue but does not fix it. You changed the use case to hide the bug 🤣.

The problem is that default configs must be set before any configuration. I would expect to be able to set default values after configuring sections as well.

In the use case, we are
- Building `file_config` using override.
- Compiling `applied_config`
- Setting the default storage type of a data node to CSV.
- Compiling `applied_config`

I expect the second compilation to understand that the default storage type is now CSV even if the default storage type has been set after creating the `file_config`.

The problem is that the `file_config` comprises the attributes in the TOML file plus the default attributes set in the default data node. So when changing the default attributes, the `file_config` attributes created from the previous default values are not updated.

The problem is the same for `_default_config`, `python_config` and `env_file_config`.

I believe it s not a top priority. I propose to postpone the fix. What do you think?

trgiangdo: I think the logic of "set default value before configure anything" makes sense for the user.
That's why I wouldn't change the behavior of set_default_configuration.

trgiangdo: However, if we really need to, I believe we need to discuss some use-cases and open a ticket on taipy-config side.
And I agree that it's not urgent.

jrobinAV (approved):

Issue #671, opened 2023-07-06
Expose an api to know if a job is deletable


jrobinAV:
- [x] re-use the is_deletable(sceanrio) API
- [x] add entries in release notes
- [x] Add documentation (user man and ref man)

PR #699, opened 2023-07-24
feature/#671 added api for job deletable

toan-quach:
#671

### DIFF
--- a/src/taipy/core/job/job.py
+++ b/src/taipy/core/job/job.py
@@ -343,7 +343,7 @@ class Job(_Entity, _Labeled):
         return self._get_simple_label()
 
     def is_deletable(self) -> bool:
-        """Indicate if the scenario can be deleted.
+        """ Indicate if the job can be deleted.
 
         Returns:
             True if the scenario can be deleted. False otherwise.
--- a/src/taipy/core/job/job.py
+++ b/src/taipy/core/job/job.py
@@ -343,7 +343,7 @@ class Job(_Entity, _Labeled):
         return self._get_simple_label()
 
     def is_deletable(self) -> bool:
-        """ Indicate if the job can be deleted.
+        """Indicate if the job can be deleted.
 
         Returns:
             True if the job can be deleted. False otherwise.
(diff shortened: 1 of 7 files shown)


jrobinAV (dismissed):
Neat.

trgiangdo (commented):
The PR looks good.

I only have a few minor comments:
1. Should we expose the `is_deletable()` method at Job level as well? Just like scenario, so the user can do something like `job.is_deletable()`.
2. Should the `test_is_deletable()` be duplicated to the `test_job_manager_with_sql_repo.py` as well?

jrobinAV:
> The PR looks good.
> 
> I only have a few minor comments:
> 
> 1. Should we expose the `is_deletable()` method at Job level as well? Just like scenario, so the user can do something like `job.is_deletable()`.
> 2. Should the `test_is_deletable()` be duplicated to the `test_job_manager_with_sql_repo.py` as well?

Yes, good ideas.

toan-quach:
@trgiangdo you're right! lemme add that as well!

tests/core/job/test_job_manager_with_sql_repo.py
```diff
@@ -207,6 +208,56 @@ def test_force_deleting_unfinished_job(init_sql_repo):
     assert_true_after_time(lambda: reference_last_edit_date != dn_3.last_edit_date)
 
 
+def test_is_deletable():
```
trgiangdo:
```suggestion
def test_is_deletable(init_sql_repo):
```

trgiangdo (dismissed):
Just a minor comment to initialize the SQL repo on the test.

src/taipy/core/job/job.py
```diff
@@ -341,3 +341,13 @@ def get_simple_label(self) -> str:
             The simple label of the job as a string.
         """
         return self._get_simple_label()
+
+    def is_deletable(self) -> bool:
+        """Indicate if the scenario can be deleted.
```
jrobinAV:
```suggestion
        """ Indicate if the job can be deleted.
```

src/taipy/core/job/job.py
```diff
@@ -341,3 +341,13 @@ def get_simple_label(self) -> str:
...
+
+        Returns:
+            True if the scenario can be deleted. False otherwise.
```
jrobinAV:
```suggestion
            True if the job can be deleted. False otherwise.
```

trgiangdo (approved):

Issue #697, opened 2023-07-21
BUG-Performance issue when creating a new scenario


FredLL-Avaiga:
**Description**
When creating a scenario from the Gui Core elements, getting the list of scenario takes a few seconds
This is new with the latest version in develop branch

**How to reproduce**
With taipy Core Gui Elements:

- start the app,
- in the browser, add a new scenario 
- wait for the list to refresh

- code:
``` python
from taipy.gui_core.GuiCoreLib import _GuiCore

from taipy.gui import Gui, State, notify
from taipy import Scope, Config, Frequency, Core, Scenario, Pipeline, Job


def replace(input_msg, template, name):
    return input_msg.replace(template, name)


selected_scenario = None
selected_data_node = None
selected_jobs = None

def on_change(state: State, var_name: str, var_value: Scenario):
    print(f"on_change(state: State, var_name: {var_name}, var_value: {var_value})")


page = """
## Demo Core Gui

<|layout|columns=1 1|

<|
### Scenario Selector
<|{selected_scenario}|scenario_selector|height=20vh|>
|>

<|
### DataNode Selector
<|{selected_data_node}|data_node_selector|>
|>

<|
### Visualizer
<|{selected_scenario}|scenario|>
|>

<|
### Selected DataNode
<|{'None' if selected_data_node is None else selected_data_node.get_label()}|>
<|{selected_data_node}|data_node|scenario={selected_scenario}|>
|>


### DAG
<|{selected_scenario}|scenario_dag|>

|>

### Job Selector
<|{selected_jobs}|job_selector|>


"""

input_msg_cfg = Config.configure_data_node(id="input_msg", scope=Scope.GLOBAL, default_data="Hello, today it's <TPL>!")
template_cfg = Config.configure_data_node(id="template", scope=Scope.GLOBAL, default_data="<TPL>")
date_cfg = Config.configure_data_node(id="date")
message_cfg = Config.configure_data_node(id="msg")

replace_template_cfg = Config.configure_task("replace", replace, [input_msg_cfg, template_cfg, date_cfg], message_cfg)
scenario_daily_cfg = Config.configure_scenario_from_tasks(
    id="scenario_daily_cfg", task_configs=[replace_template_cfg], frequency=Frequency.DAILY
)
scenario_cfg = Config.configure_scenario_from_tasks(id="scenario_cfg", task_configs=[replace_template_cfg])

Core().run()

Gui(page).run()
```

**Expected behavior**
Good perf

[...]

PR #702, opened 2023-07-25
Fix/#697 - Performance issue when creating a new scenario

trgiangdo:
https://github.com/Avaiga/taipy-core/issues/697

The issue is that when creating the first scenario on FS repository, several necessary files are not created yet, which makes the `_load()` method raises the `ModelNotFound` exception.

And since the default `read_entity_retry` value is 3, it would take about 2-3 seconds to finish creating the scenario, based on how many exception raised.

However, when creating the second scenario, it will be just fine.

Test code:
```python
from taipy import Scope, Config, Frequency
from src.taipy.core import taipy as tp
from src.taipy.core import Core
import time


def replace(input_msg, template, name):
    return input_msg.replace(template, name)


input_msg_cfg = Config.configure_data_node(id="input_msg", scope=Scope.GLOBAL, default_data="Hello, today it's <TPL>!")
template_cfg = Config.configure_data_node(id="template", scope=Scope.GLOBAL, default_data="<TPL>")
date_cfg = Config.configure_data_node(id="date")
message_cfg = Config.configure_data_node(id="msg")

replace_template_cfg = Config.configure_task("replace", replace, [input_msg_cfg, template_cfg, date_cfg], message_cfg)
scenario_daily_cfg = Config.configure_scenario_from_tasks(
    id="scenario_daily_cfg", task_configs=[replace_template_cfg], frequency=Frequency.DAILY
)
scenario_cfg = Config.configure_scenario_from_tasks(id="scenario_cfg", task_configs=[replace_template_cfg])

Core().run()

start = time.time()
scenario = tp.create_scenario(scenario_cfg)
print(f"Create scenario time: {time.time() - start}")
start = time.time()
tp.submit(scenario)
print(f"Submit scenario time: {time.time() - start}")

start = time.time()
scenario_2 = tp.create_scenario(scenario_cfg)
print(f"Create scenario 2 time: {time.time() - start}")
start = time.time()
tp.submit(scenario_2)
print(f"Submit scenario 2 time: {time.time() - start}")
```

Result:
```console
[2023-07-25 11:36:35][Taipy][INFO] Development mode: Clean all entities of version a4b4a09f-ecad-465b-bbdf-ceda3077f2f4
[2023-07-25 11:36:38][Taipy][ERROR] DataNode not found: DATANODE_input_msg_c9ca66fe-5a3e-46cc-a7ef-379c5dff64e5
[2023-07-25 11:36:40][Taipy][ERROR] DataNode not found: DATANODE_template_a544c333-bf3c-42a3-b8ea-06414ebb5183
Create scenario time: 3.0631818771362305
[2023-07-25 11:36:40][Taipy][WARNING] DATANODE_date_a6bf9c3f-697a-402a-a825-d00245cfe784 cannot be read because it has never been written. Hint: The data node may refer to a wrong path : .data/pickles/DATANODE_date_a6bf9c3f-697a-402a-a825-d00245cfe784.p 
Submit scenario time: 0.0188140869140625
Create scenario 2 time: 0.03878903388977051
[2023-07-25 11:36:40][Taipy][WARNING] DATANODE_date_3fee7bdd-300f-4edd-94dd-4564bd339ed3 cannot be read because it has never been written. Hint: The data node may refer to a wrong path : .data/pickles/DATANODE_date_3fee7bdd-300f-4edd-94dd-4564bd339ed3.p 
Submit scenario 2 time: 0.014927148818969727
```

You can see that in the result, the time to create scenario_2 is 0.03 second, which is just fine.

There are 2 possible solutions that I can think of.
1. Instead of retry on all Exception, we only retry on `json.JSONDecodeError`, which was raised by Fred a while ago.
(The exception that is causing the performance issue is `ModelNotFound`, which is catch at the `_Manager._get()` method).
This is also what I implemented.
2. Reduce the default `read_entity_retry` to 1, and maybe reduce the sleep time (0.5s maybe too long, idk).
This will let the creation performance of the first scenario be worse than the next.
It won't be too long anyway.

### DIFF
--- a/src/taipy/core/_repository/_filesystem_repository.py
+++ b/src/taipy/core/_repository/_filesystem_repository.py
@@ -19,7 +19,7 @@ from taipy.config.config import Config
 
 from ..common._utils import _retry_read_entity
 from ..common.typing import Converter, Entity, Json, ModelType
-from ..exceptions import InvalidExportPath, ModelNotFound
+from ..exceptions import FileCannotBeRead, InvalidExportPath, ModelNotFound
 from ._abstract_repository import _AbstractRepository
 from ._decoder import _Decoder
 from ._encoder import _Encoder
@@ -39,7 +39,7 @@ class _FileSystemRepository(_AbstractRepository[ModelType, Entity]):
         dir_name (str): Folder that will hold the files for this dataclass model.
     """
 
-    __EXCEPTIONS_TO_RETRY = (json.JSONDecodeError,)
+    __EXCEPTIONS_TO_RETRY = (FileCannotBeRead,)
 
     def __init__(self, model_type: Type[ModelType], converter: Type[Converter], dir_name: str):
         self.model_type = model_type
@@ -57,6 +57,7 @@ class _FileSystemRepository(_AbstractRepository[ModelType, Entity]):
     ###############################
     # ##   Inherited methods   ## #
     ###############################
+
     def _save(self, entity: Entity):
         self.__create_directory_if_not_exists()
         model = self.converter._entity_to_model(entity)  # type: ignore
@@ -67,16 +68,16 @@ class _FileSystemRepository(_AbstractRepository[ModelType, Entity]):
     def _exists(self, entity_id: str) -> bool:
         return self.__get_path(entity_id).exists()
 
-    @_retry_read_entity(__EXCEPTIONS_TO_RETRY)
     def _load(self, entity_id: str) -> Entity:
+        path = pathlib.Path(self.__get_path(entity_id))
+
         try:
-            with pathlib.Path(self.__get_path(entity_id)).open(encoding="UTF-8") as source:
-                file_content = json.load(source)
-            return self.__file_content_to_entity(file_content)
-        except FileNotFoundError:
+            file_content = self.__read_file(path)
+        except (FileNotFoundError, FileCannotBeRead):
             raise ModelNotFound(str(self.dir_path), entity_id)
 
-    @_retry_read_entity(__EXCEPTIONS_TO_RETRY)
+        return self.__file_content_to_entity(file_content)
+
     def _load_all(self, filters: Optional[List[Dict]] = None) -> List[Entity]:
         entities = []
         try:
@@ -173,7 +174,7 @@ class _FileSystemRepository(_AbstractRepository[ModelType, Entity]):
     #############################
     # ##   Private methods   ## #
     #############################
-    @_retry_read_entity(__EXCEPTIONS_TO_RETRY)
+
     def __filter_files_by_config_and_owner_id(
         self, config_id: str, owner_id: Optional[str], filters: Optional[List[Dict]] = None
     ):
@@ -192,7 +193,6 @@ class _FileSystemRepository(_AbstractRepository[ModelType, Entity]):
             pass
         return None
 
-    @_retry_read_entity(__EXCEPTIONS_TO_RETRY)
     def __match_file_and_get_entity(self, filepath, config_and_owner_ids, filters):
         if match := [(c, p) for c, p in config_and_owner_ids if c.id in filepath.name]:
             for config, owner_id in match:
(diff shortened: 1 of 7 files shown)


src/taipy/core/_repository/_filesystem_repository.py
```diff
@@ -64,7 +66,7 @@ def _save(self, entity: Entity):
     def _exists(self, entity_id: str) -> bool:
         return self.__get_path(entity_id).exists()
 
-    @_retry_read_entity((Exception,))
+    @_retry_read_entity(__EXCEPTIONS_TO_RETRY)
     def _load(self, entity_id: str) -> Entity:
         try:
             with pathlib.Path(self.__get_path(entity_id)).open(encoding="UTF-8") as source:
```
toan-quach:
```suggestion
    path = pathlib.Path(self.__get_path(entity_id))
    if path.is_file():
        try:
            with pathlib.Path(self.__get_path(entity_id)).open(encoding="UTF-8") as source:
```

toan-quach (changes requested):
I think we can check if the file exists first, if it doesn't exist, there's no point in trying to read it anyway. This can also solve the performance issue.
0.5s is also kinda long IMO, for decent computer storage should be pretty fast, 0.2s seem more reasonable for me for almost "instant" feedback.

src/taipy/core/_repository/_filesystem_repository.py
trgiangdo: The file-not-exists case is already caught by the `except FileNotFound:` right?

trgiangdo:
> 0.5s is also kinda long IMO, for decent computer storage should be pretty fast, 0.2s seem more reasonable for me for almost "instant" feedback.

Agree. A 0.1 or 0.2 seconds sleep should be enough

src/taipy/core/_repository/_filesystem_repository.py
toan-quach: short answer is Yes 😃 long answer is, you will still need to try and "read" it to know it doesn't exist. Maybe we should introduce a new Exception here "FileCannotBeRead" for example. 
- If path == False: Raise "FileNotFoundError" -> don't retry (no point trying to read a non-existing file again and again anyway) 
- If path == True: Try read the file, if cant -> raise"FileCannotBeRead" and retry

What do you think? @trgiangdo @jrobinAV

trgiangdo: Interesting idea. That would help us having more control on when to retry.

jrobinAV: I believe it s a great idea !

jrobinAV:
There is something I don't understand.

We changed the default Core config to set the retry nb to 3. Right?
Before the change, it was not by default in Core but it was set to 3 in Taipy Core viz elements library. 

So, in the end, I don't understand why the behavior is different. It should be the exact same behavior from the Core viz elements standpoint.

trgiangdo:
As I understand, the `read_entity_retry` was set to 3 in `src/taipy/gui_core/GuiCoreLib.py` in taipy.

In `src/taipy/__init__.py`, Core is initialized before GuiCoreLib, so probably setting the `read_entity_retry` did not take effect at all (because the value of Config.global_config.read_entity_retry is already passed to the wrapper).
Which is really weird

trgiangdo:
@toan-quach @jrobinAV I implemented Toan's idea, however, there was an issue.

The `_load()` method always need to raise the ModelNotFound exception when there is an error (for the `_Manager_get()` method).
If we modify the catch on manager level, it can cause hidden bug (since we have multiple repository types).

Instead, I notice that we only want to retry_read_entity when reading a file. So I split the responsibility to a separate private method, and only retry at that level.

There is 1 more question remains: Do we only retry when the file can not be read, or do we also retry when the file doesn't exist?

Please let me know what you think.

src/taipy/core/_repository/_filesystem_repository.py
```diff
@@ -159,14 +163,18 @@ def _get_by_config_and_owner_id(
...
-            filters.append({"owner_id": owner_id})
+            for fil in filters:
+                fil.update({"owner_id": owner_id})
         return self.__filter_files_by_config_and_owner_id(config_id, owner_id, filters)
 
     #############################
     # ##   Private methods   ## #
     #############################
-    @_retry_read_entity((Exception,))
+
```
jrobinAV, line 177:
```suggestion
```

jrobinAV:
> There is 1 more question remains: Do we only retry when the file can not be read, or do we also retry when the file doesn't exist?

I believe only when the file exists and cannot be read. There is no point retrying when the file does not exist yet.

Thx,

jrobinAV (approved):

toan-quach (approved):
I agree with @jrobinAV remark :D

Issue #700, opened 2023-07-24
BUG-Write of Excel data node not working properly


FlorianJacta:
**Description**
An error occurs when trying to read a Excel Data Node with a particular sheet name that has been written by Taipy.

```
  File "...\site-packages\pandas\io\excel\_openpyxl.py", line 569, in get_sheet_by_name
    self.raise_if_bad_sheet_by_name(name)
  File "...\site-packages\pandas\io\excel\_base.py", line 597, in raise_if_bad_sheet_by_name
    raise ValueError(f"Worksheet named '{name}' not found")
ValueError: Worksheet named 'one' not found
```

**How to reproduce**

```python
import taipy as tp 

in_ = tp.Config.configure_data_node(id="input", default_data={"x": [1, 2], "y": [3, 4]})

out = tp.Config.configure_excel_data_node(id="output", sheet_name="one")

def id(i):
    return i 

t = tp.Config.configure_task(id="task", input=[in_], function=id, output=[out])

s_cfg = tp.Config.configure_scenario_from_tasks(id="scenario", task_configs=[t])

tp.Core().run()

s = tp.create_scenario(s_cfg)
s.submit()

print(s.output.read())

```

**Expected behavior**
The above code should work. The Excel file should contain the sheet.

**Runtime environment**
Taipy 2.3.1

toan-quach:
@jrobinAV 
should we also include the fix for this in 2.3?

jrobinAV:
Yes, we should merge the fix into 3 branches:
- release/2.3
- dev/2.4
- develop

jrobinAV:
Can we move the ticket to done?
@FlorianJacta did you test it?

FlorianJacta:
Not working on 2.4 with the same error

jrobinAV:
Hum 🤔. 
Thank you. 
We are going to check that.

toan-quach:
@FlorianJacta I just tested using branch `release/2.4` on Github (the fix hasn't been published yet so please use branch realease/2.4 instead) 

Here is my output:
![image](https://github.com/Avaiga/taipy-core/assets/93168955/1ba4cbb1-53dd-41e6-8240-441ae948723e)

Here is the code I used: 
```
import taipy as tp 

in_ = tp.Config.configure_data_node(id="input", default_data={"x": [1, 2], "y": [3, 4]})

out = tp.Config.configure_excel_data_node(id="output", sheet_name="one")

def id(i):
    return i

[...]

PR #707, opened 2023-07-27
[2.3] hotfix/#700 excel dn behavior inconsistent with pandas

toan-quach:
#700

### DIFF
--- a/src/taipy/core/data/excel.py
+++ b/src/taipy/core/data/excel.py
@@ -24,7 +24,12 @@ from taipy.config.common.scope import Scope
 from .._backup._backup import _replace_in_backup_file
 from .._entity._reload import _self_reload
 from .._version._version_manager_factory import _VersionManagerFactory
-from ..exceptions.exceptions import ExposedTypeLengthMismatch, InvalidExposedType, NonExistingExcelSheet
+from ..exceptions.exceptions import (
+    ExposedTypeLengthMismatch,
+    InvalidExposedType,
+    NonExistingExcelSheet,
+    SheetNameLengthMismatch,
+)
 from ..job.job_id import JobId
 from .abstract_file import _AbstractFileDataNode
 from .data_node import DataNode
@@ -287,29 +292,41 @@ class ExcelDataNode(DataNode, _AbstractFileDataNode):
         except pd.errors.EmptyDataError:
             return modin_pd.DataFrame()
 
-    def __write_excel_with_sheet_name(self, write_excel_fct, *args, **kwargs):
+    def __write_excel_with_single_sheet(self, write_excel_fct, *args, **kwargs):
         sheet_name = self.properties.get(self.__SHEET_NAME_PROPERTY)
         if sheet_name:
-            sheet_name = sheet_name if isinstance(sheet_name, str) else sheet_name[0]
+            if not isinstance(sheet_name, str):
+                if len(sheet_name) > 1:
+                    raise SheetNameLengthMismatch
+                else:
+                    sheet_name = sheet_name[0]
             write_excel_fct(*args, **kwargs, sheet_name=sheet_name)
         else:
             write_excel_fct(*args, **kwargs)
+
+    def __write_excel_with_multiple_sheets(self, data: Any, columns: List[str] = None):
+        with pd.ExcelWriter(self._path) as writer:
+            # Each key stands for a sheet name
+            for key in data.keys():
+                if isinstance(data[key], np.ndarray):
+                    df = pd.DataFrame(data[key])
+                else:
+                    df = data[key]
+
+                if columns:
+                    data[key].columns = columns
+
+                df.to_excel(writer, key, index=False)
 
     def _write(self, data: Any):
         if isinstance(data, Dict) and all(
             [isinstance(x, (pd.DataFrame, modin_pd.DataFrame, np.ndarray)) for x in data.values()]
         ):
-            with pd.ExcelWriter(self._path) as writer:
-                # Write to excel file with multiple sheets (each key stands for a sheet name)
-                for key in data.keys():
-                    if isinstance(data[key], np.ndarray):
-                        pd.DataFrame(data[key]).to_excel(writer, key, index=False)
-                    else:
-                        data[key].to_excel(writer, key, index=False)
+            self.__write_excel_with_multiple_sheets(data)
         elif isinstance(data, (pd.DataFrame, modin_pd.DataFrame)):
-            self.__write_excel_with_sheet_name(data.to_excel, self._path, index=False)
+            self.__write_excel_with_single_sheet(data.to_excel, self._path, index=False)
         else:
-            self.__write_excel_with_sheet_name(pd.DataFrame(data).to_excel, self._path, index=False)
+            self.__write_excel_with_single_sheet(pd.DataFrame(data).to_excel, self._path, index=False)
 
     def write_with_column_names(self, data: Any, columns: List[str] = None, job_id: Optional[JobId] = None):
         """Write a set of columns.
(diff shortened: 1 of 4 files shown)


tests/core/data/test_excel_data_node.py
```diff
@@ -320,12 +352,8 @@ def test_read_multi_sheet_with_header(self):
         excel_data_node_as_pandas_no_sheet_name = ExcelDataNode("bar", Scope.PIPELINE, properties={"path": path})
 
         data_pandas_no_sheet_name = excel_data_node_as_pandas_no_sheet_name.read()
-        assert isinstance(data_pandas_no_sheet_name, Dict)
-        assert len(data_pandas_no_sheet_name) == 2
-        assert data_pandas.keys() == data_pandas_no_sheet_name.keys()
-
-        for sheet_name in sheet_names:
-            assert data_pandas[sheet_name].equals(data_pandas_no_sheet_name[sheet_name])
+        assert isinstance(data_pandas_no_sheet_name, pd.DataFrame)
```
toan-quach: Here, the default behavior of pandas when reading an Excel file is that if nothing is provided for the sheet_name parameter, it will try to load "Sheet1" with "Sheet1" is the default value.
In our case, if no sheet_name is provided, we will pass None to the sheet_name parameter. This will load all sheets in the Excel file.
This fix will have a side effect, it stays true to the default behavior of pandas, but if user passes in None and expect it to return all sheets, it will go back to the default and only returns "Sheet1"

src/taipy/core/data/excel.py
```diff
@@ -241,11 +241,17 @@ def _read_as_numpy(self):
         return sheets.to_numpy()
 
     def _do_read_excel(self, engine, sheet_names, kwargs) -> pd.DataFrame:
```
joaoandre-avaiga: What if we change the default value of sheet_names? From this:
```
def _read_as_pandas_dataframe(self, sheet_names=None)
```
To this:
```
def _read_as_pandas_dataframe(self, sheet_names="")
```

And on _do_read_excel, do this:
if sheet_names != "":
            df = pd.read_excel(
                self._path,
                sheet_name=sheet_names,
                **kwargs,
            )
        else:
            df = pd.read_excel(
                self._path,
                **kwargs,
            )

Could that work so we maintain all the options available? If no name is provided, sheet_names is `""` and we just call read_excel without parameters. If the user passes None, them a dict is returned because we would pass `sheet_names=None` to read_excel. And any other name or list of names would alsso work.

The only thing Im questioning is if empty string is a valid excel sheet name 👀 . What do you guys think? @toan-quach  @jrobinAV

toan-quach: Actually, I just remembered, when initializing the ExcelDataNode, if sheet_name is not provided in the property, we will set it to None. 
image

We could set it to the default value of pandas instead ("Sheet1"). This I think is a much simpler solution, reducing a lot of painstaking if-else. The only drawback I can think of atm is we will have to monitor this default value. Though I doubt they have any plans of changing it soon. So then your proposal can also be "Sheet1" instead of "". What do you think? @joaoandre-avaiga @jrobinAV

tests/core/data/test_excel_data_node.py
```diff
@@ -263,6 +263,44 @@ def test_write(self, excel_file, default_data_frame, content, columns):
...
+
+        assert np.array_equal(excel_dn.read().values, default_data_frame.values)
+        if not columns:
+            excel_dn.write(content)
+            df = pd.DataFrame(content)
+        else:
+            excel_dn.write_with_column_names(content, columns)
+            df = pd.DataFrame(content, columns=columns)
+
+        # breakpoint()
```
jrobinAV: to remove

src/taipy/core/data/excel.py
```diff
@@ -287,20 +287,29 @@ def _read_as_modin_dataframe(
         except pd.errors.EmptyDataError:
             return modin_pd.DataFrame()
 
+    def __write_excel_with_sheet_name(self, write_excel_fct, *args, **kwargs):
```
jrobinAV: I would rename the method `__do_write` since we don't necessarily write with a sheet name

src/taipy/core/data/excel.py
```diff
@@ -314,5 +323,6 @@ def write_with_column_names(self, data: Any, columns: List[str] = None, job_id:
             df = pd.DataFrame(data)
         else:
             df = pd.DataFrame(data, columns=columns)
-        df.to_excel(self.path, index=False)
+
```
jrobinAV: I suspect this method only works with a single sheet Excel file. Either we should make it explicit or we should make it compatible with multiple sheets.

src/taipy/core/data/excel.py
```diff
@@ -287,20 +287,29 @@ def _read_as_modin_dataframe(
         except pd.errors.EmptyDataError:
             return modin_pd.DataFrame()
 
+    def __write_excel_with_sheet_name(self, write_excel_fct, *args, **kwargs):
+        sheet_name = self.properties.get(self.__SHEET_NAME_PROPERTY)
+        if sheet_name:
+            sheet_name = sheet_name if isinstance(sheet_name, str) else sheet_name[0]
```
jrobinAV: This line looks suspicious to me. 

Why do we only write the first one if there are multiple sheet names? 
The number of data frames to write should be the same as the nb of sheet names. No?
Since this method is only called when the data is not a dictionary of data frames, that means it is only used with single-sheet Excel files. This is inconsistent with multiple sheet names. No?
In other words, do we want to raise an error when writing data that does not match the number of sheet names?

tests/core/data/test_excel_data_node.py
```diff
@@ -263,6 +263,44 @@ def test_write(self, excel_file, default_data_frame, content, columns):
...
+            ([[11, 22, 33], [44, 55, 66]], ["e", "f", "g"], "sheet_name"),
+        ],
+    )
+    def test_write_with_sheet_name(self, excel_file_with_sheet_name, default_data_frame, content, columns, sheet_name):
+        if sheet_name:
+            excel_dn = ExcelDataNode(
+                "foo", Scope.PIPELINE, properties={"path": excel_file_with_sheet_name, "sheet_name": sheet_name}
+            )
+        else:
+            excel_dn = ExcelDataNode("foo", Scope.PIPELINE, properties={"path": excel_file_with_sheet_name})
```
jrobinAV: Useless regarding the set of parameters

tests/core/data/test_excel_data_node.py
```diff
@@ -263,6 +263,44 @@ def test_write(self, excel_file, default_data_frame, content, columns):
...
         assert len(excel_dn.read()) == 0
 
+    @pytest.mark.parametrize(
+        "content,columns,sheet_name",
+        [
+            ([{"a": 11, "b": 22, "c": 33}, {"a": 44, "b": 55, "c": 66}], None, "sheet_name"),
+            ([[11, 22, 33], [44, 55, 66]], ["e", "f", "g"], "sheet_name"),
+        ],
+    )
+    def test_write_with_sheet_name(self, excel_file_with_sheet_name, default_data_frame, content, columns, sheet_name):
```
jrobinAV: This test looks messy to me. 
In some cases, we are testing the `write` function; in some other, we are testing the `write_with_column_names`. It should be split into multiple tests, and the name of the test should be clear enough to understand exactly what is tested.

jrobinAV:
Please, before merging wait for https://github.com/Avaiga/taipy-core/pull/709 to be approved to propagate changes.

src/taipy/core/data/excel.py
toan-quach: I have renamed it to `__write_excel_with_single_sheet` :D

src/taipy/core/data/excel.py
toan-quach: I think you're right, we should raise an error instead, let me fix that to add check if sheet name length is more than 1, we will raise an error as this is single sheet writing

tests/core/data/test_excel_data_node.py
toan-quach: hmmm I dont think I get what you mean?

jrobinAV: The else condition is never used. The `sheet_name` variable always equals `"sheet_name"`.

tests/core/data/test_excel_data_node.py
toan-quach: I can only say I have tried my best in the most recent commit 😃

src/taipy/core/data/excel.py
jrobinAV: I love it

tests/core/data/test_excel_data_node.py
jrobinAV: 🤣
My comment is not specific to your PR or even this test. And that's not a blocking point.
We all (in the back-end team) tend to write too-long unit tests testing multiple things. Sometimes, they are not unitary.

I should create a dedicated issue to spend a couple of days reviewing all the unit tests and trying to improve the overall readability by shortening them, splitting some of them, eventually finding better names, and/or better organizing them in multiple modules and classes...

Let's forget about this comment if that's ok for you. It should be on another issue.

toan-quach: too late, already made changes to that 🤣

jrobinAV: Haha, that s useful anyway. Thx.

jrobinAV (approved):

PR #709, opened 2023-07-28
[3.0] bug/#700 write of excel dn not working properly

toan-quach:
#700

### DIFF
--- a/src/taipy/core/data/excel.py
+++ b/src/taipy/core/data/excel.py
@@ -241,17 +241,11 @@ class ExcelDataNode(DataNode, _AbstractFileDataNode, _AbstractTabularDataNode):
         return sheets.to_numpy()
 
     def _do_read_excel(self, engine, sheet_names, kwargs) -> pd.DataFrame:
-        if sheet_names:
-            df = pd.read_excel(
-                self._path,
-                sheet_name=sheet_names,
-                **kwargs,
-            )
-        else:
-            df = pd.read_excel(
-                self._path,
-                **kwargs,
-            )
+        df = pd.read_excel(
+            self._path,
+            sheet_name=sheet_names,
+            **kwargs,
+        )
         # We are using pandas to load modin dataframes because of a modin issue
         # https://github.com/modin-project/modin/issues/4924
         if engine == "modin":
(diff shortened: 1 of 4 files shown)


tests/core/data/test_excel_data_node.py
```diff
@@ -263,6 +259,38 @@ def test_write(self, excel_file, default_data_frame, content, columns):
...
+            "foo", Scope.SCENARIO, properties={"path": excel_file_with_sheet_name, "sheet_name": sheet_name}
+        )
+        excel_dn.read()
+        assert np.array_equal(excel_dn.read().values, default_data_frame.values)
+        if not columns:
+            excel_dn.write(content)
+            df = pd.DataFrame(content)
+        else:
+            excel_dn.write_with_column_names(content, columns)
+            df = pd.DataFrame(content, columns=columns)
```
trgiangdo: In this test, the column is always None. So the `else` will not be called at all, right?

Should we add another test with defined `columns` in the `@pytest.mark.parametrize`?

tests/core/data/test_excel_data_node.py
```diff
@@ -263,6 +259,38 @@ def test_write(self, excel_file, default_data_frame, content, columns):
...
+
+        assert np.array_equal(excel_dn.read().values, df.values)
+
+        sheet_names = pd.ExcelFile(excel_file_with_sheet_name).sheet_names
+        if sheet_name:
+            expected_sheet_name = sheet_name[0] if isinstance(sheet_name, list) else sheet_name
+        else:
+            expected_sheet_name = "Sheet1"
```
trgiangdo: Since the `excel_file_with_sheet_name` always has a sheet_names, is this if else a bit redundant?

src/taipy/core/data/excel.py
```diff
@@ -80,6 +80,7 @@ class ExcelDataNode(DataNode, _AbstractFileDataNode, _AbstractTabularDataNode):
     __DEFAULT_PATH_KEY = "default_path"
     __HAS_HEADER_PROPERTY = "has_header"
     __SHEET_NAME_PROPERTY = "sheet_name"
+    __PANDAS_DEFAULT_SHEET_NAME = "Sheet1"
```
trgiangdo: Should we clarify this in the doc, so the user know that the default value is?

trgiangdo (commented):
Just a few minor comments

src/taipy/core/data/excel.py
toan-quach: as discussed, we will revert to the previous version :D

tests/core/data/test_excel_data_node.py
toan-quach: as discussed, we will revert to the previous version :D

jrobinAV:
I did comment on the PR https://github.com/Avaiga/taipy-core/pull/707 instead... Sorry I should have started with the one to be merged into develop.

jrobinAV (dismissed):

jrobinAV (approved):

PR #711, opened 2023-08-01
chore: update GH actions access token

joaoandre-avaiga:


trgiangdo (approved):
The benchmark workflow is failing on develop, so I guess this will fix that. Thank you Joao

jrobinAV (approved):

Issue #687, opened 2023-07-18
Remove config_id from Pipeline entity.


jrobinAV:
- [x] Remove config_id attribute from pipeline entity, and propagate the change in the repository and the manager.
- [x] In particular change the PipelineManager._get_or_create method to take as parameter a list of tasks or ids instead of the PipelineConfig.

PR #715, opened 2023-08-02
feature/#687 remove config_id from pipeline

toan-quach:
#687

### DIFF
--- a/src/taipy/core/pipeline/_pipeline_manager.py
+++ b/src/taipy/core/pipeline/_pipeline_manager.py
@@ -87,7 +87,7 @@ class _PipelineManager(_Manager[Pipeline], _VersionMixin):
         _publish_event(cls._EVENT_ENTITY_TYPE, pipeline.id, EventOperation.UPDATE, "subscribers")
 
     @classmethod
-    def _get_or_create(
+    def _create(
         cls,
         pipeline_name: str,
         tasks: List[Task],
@@ -107,7 +107,7 @@ class _PipelineManager(_Manager[Pipeline], _VersionMixin):
 
         version = cls._get_latest_version()
         pipeline = Pipeline(
-            {},
+            {"name": pipeline_name},
             tasks,
             pipeline_id,
             owner_id,
@@ -115,7 +115,6 @@ class _PipelineManager(_Manager[Pipeline], _VersionMixin):
             version=version,
         )
         for task in tasks:
-            task._parent_ids = task.parent_ids
             task._parent_ids.update([pipeline_id])
         cls.__save_tasks(tasks)
         cls._set(pipeline)
--- a/src/taipy/core/pipeline/_pipeline_manager.py
+++ b/src/taipy/core/pipeline/_pipeline_manager.py
@@ -89,14 +89,12 @@ class _PipelineManager(_Manager[Pipeline], _VersionMixin):
     @classmethod
     def _get_or_create(
         cls,
-        pipeline_config: PipelineConfig,
+        pipeline_name: str,
+        tasks: List[Task],
         cycle_id: Optional[CycleId] = None,
         scenario_id: Optional[ScenarioId] = None,
     ) -> Pipeline:
-        pipeline_id = Pipeline._new_id(str(pipeline_config.id))
-
-        task_manager = _TaskManagerFactory._build_manager()
-        tasks = task_manager._bulk_get_or_create(pipeline_config.task_configs, cycle_id, scenario_id)
+        pipeline_id = Pipeline._new_id(pipeline_name)
 
         scope = min(task.scope for task in tasks) if len(tasks) != 0 else Scope.GLOBAL
         owner_id: Union[Optional[PipelineId], Optional[ScenarioId], Optional[CycleId]]
@@ -107,16 +105,9 @@ class _PipelineManager(_Manager[Pipeline], _VersionMixin):
         else:
             owner_id = None
 
-        filters = cls._build_filters_with_version(None)
-        if pipelines_from_owner := cls._repository._get_by_config_and_owner_id(  # type: ignore
-            str(pipeline_config.id), owner_id, filters
-        ):
-            return pipelines_from_owner
-
         version = cls._get_latest_version()
         pipeline = Pipeline(
-            str(pipeline_config.id),
-            dict(**pipeline_config._properties),
+            {},
             tasks,
             pipeline_id,
             owner_id,
@@ -124,6 +115,7 @@ class _PipelineManager(_Manager[Pipeline], _VersionMixin):
             version=version,
         )
         for task in tasks:
+            task._parent_ids = task.parent_ids
             task._parent_ids.update([pipeline_id])
         cls.__save_tasks(tasks)
         cls._set(pipeline)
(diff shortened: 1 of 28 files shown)


src/taipy/core/pipeline/pipeline.py
```diff
@@ -57,27 +56,27 @@ class Pipeline(_Entity, Submittable, _Labeled):
 
     def __init__(
         self,
-        config_id: str,
         properties: Dict[str, Any],
```
trgiangdo, line 59: Should we also remove the `properties`?

Since the PipelineConfig is gone, there will be no way to configure any properties. So it is kinda useless now.

src/taipy/core/scenario/_scenario_model.py
```diff
@@ -53,7 +53,7 @@ class _ScenarioModel(_BaseModel):
     subscribers: List[Dict]
     tags: List[str]
     version: str
-    pipelines: Optional[List[PipelineId]] = None
+    pipelines: Optional[Union[Dict[str, PipelineId], Dict]] = None
```
trgiangdo, line 56: This type hint seems weird.
Can it be `Optional[Dict[str, PipelineId]]`?
Or `Optional[Union[Dict[str, PipelineId], Dict[str, Pipeline]]]`

src/taipy/core/scenario/scenario.py
```diff
@@ -100,7 +100,7 @@ def __init__(
         self._primary_scenario = is_primary
         self._tags = tags or set()
         self._properties = _Properties(self, **properties)
-        self._pipelines: Union[List[Pipeline], List] = pipelines or []
+        self._pipelines: Union[Dict[str, Pipeline], Dict[str, PipelineId], Dict] = pipelines or {}
```
trgiangdo, line 103:
```suggestion
        self._pipelines: Union[Dict[str, Pipeline], Dict[str, PipelineId]] = pipelines or {}
```
Again, there is an additional `Dict` type hint. Not sure if it's necessary/

src/taipy/core/pipeline/pipeline.py
toan-quach: at first it will be set to empty, but I keep it there for users to use if they need it

src/taipy/core/scenario/scenario.py
toan-quach: mypy is complaining about it 🤣  as the default value is { }

trgiangdo: Ah ... mypy again

src/taipy/core/taipy.py
```diff
@@ -579,18 +579,6 @@ def create_scenario(
     return _ScenarioManagerFactory._build_manager()._create(config, creation_date, name)
 
 
-def create_pipeline(config: PipelineConfig) -> Pipeline:
```
jrobinAV, line 582: Must be documented in the release notes.

src/taipy/core/pipeline/_pipeline_manager.py
```diff
@@ -89,14 +89,12 @@ def __remove_subscriber(cls, callback, params, pipeline):
     @classmethod
     def _get_or_create(
```
jrobinAV: I would rename it to `_create` since we don't try to get an existing one anymore.

src/taipy/core/pipeline/_pipeline_manager.py
```diff
@@ -107,23 +105,17 @@ def _get_or_create(
...
-            dict(**pipeline_config._properties),
+            {},
             tasks,
             pipeline_id,
             owner_id,
             {scenario_id} if scenario_id else None,
             version=version,
         )
         for task in tasks:
+            task._parent_ids = task.parent_ids
```
jrobinAV: I don't get this. Can't we just do:
```
 task.parent_ids.update([pipeline_id])
```

src/taipy/core/scenario/_scenario_manager.py
```diff
@@ -161,9 +164,9 @@ def _create(
             dn._parent_ids.update([scenario_id])
             _data_manager._set(dn)
 
-        for pipeline in pipelines:
+        for pipeline in pipelines.values():
             pipeline._parent_ids.update([scenario_id])
-        cls.__save_pipelines(pipelines)
+        cls.__save_pipelines(pipelines.values())
```
jrobinAV, line 169: Since scenarios do not share any pipeline anymore, a pipeline has only one parent that can be set at its creation. Am I right?

src/taipy/core/scenario/scenario.py
jrobinAV: if we replace {} by Dict(), is it better?

src/taipy/core/pipeline/_pipeline_manager.py
```diff
@@ -107,23 +105,17 @@ def _get_or_create(
...
-        if pipelines_from_owner := cls._repository._get_by_config_and_owner_id(  # type: ignore
-            str(pipeline_config.id), owner_id, filters
-        ):
-            return pipelines_from_owner
-
         version = cls._get_latest_version()
         pipeline = Pipeline(
-            str(pipeline_config.id),
-            dict(**pipeline_config._properties),
+            {},
```
jrobinAV: Please add the name as a property:
```suggestion
            {"name": pipeline_name},
```

tests/core/_entity/test_labelled.py
```diff
@@ -101,10 +101,10 @@ def test_get_label_complex_case():
...
     assert scenario.get_label() == "Today > My Name"
     assert scenario.get_simple_label() == "My Name"
-    assert pipeline_C.get_label() == "Today > pipeline_C"
-    assert pipeline_C.get_simple_label() == "pipeline_C"
-    assert pipeline_S.get_label() == "Today > My Name > pipeline_S"
-    assert pipeline_S.get_simple_label() == "pipeline_S"
+    assert "Today > PIPELINE_pipeline_C" in pipeline_C.get_label()
+    assert "PIPELINE_pipeline_C" in pipeline_C.get_simple_label()
+    assert "Today > My Name > PIPELINE_pipeline_S" in pipeline_S.get_label()
+    assert "PIPELINE_pipeline_S" in pipeline_S.get_simple_label()
```
jrobinAV: That's an undesired consequence...
We should avoid as much as possible to have a label that a human cannot read. So we want to avoid having an ID as label. 
By adding the name property to the Pipeline, it should solve the pb.

src/taipy/core/scenario/_scenario_manager.py
toan-quach: yep :D

src/taipy/core/pipeline/_pipeline_manager.py
toan-quach: fixed it back to ` task._parent_ids.update([pipeline_id])`. In short, there was a failed test, the reason for failing was correct, but the test itself was wrong 🤣 so I updated the test and revert to the previous version. It's a bit difficult to explain in text but in short, it's related to 2 same entities but load to different address and not in total sync. The test was failing for the correct reason, but what it was testing was a wrong use case, so fixing the test is enough.

src/taipy/core/scenario/scenario.py
toan-quach: I tried but it's still complaining...

trgiangdo (approved):

Issue #718, opened 2023-08-04
Stop storing PipelineModels


jrobinAV:
- Remove PipelineModel, PipelineRepository (FS, SQL), PipelineConverter
- Make PipelineManager methods work directly with ScenarioRepository
- Make PipelineManagerFactory build PipelineManager "injecting" scenario repository

PR #722, opened 2023-08-09
feature/#718 stop storing PipelineModel

toan-quach:
#718 

What this PR does:
- Remove dependency on pipeline repositories in PipelineManager
- Rewrite all _Manager functions for PipelineManger
- Remove initializing of Pipeline repositories in PipelineManagerFactory
- Store Pipeline information in ScenarioModel
- Rewrite pipeline manipulation via Scenario entity
- Add unit tests where applicable

What this PR doesn't do:
- Remove Pipeline repositories, PipelineModel, PipelineConverter (for safety purposes, I want to guarantee that everything works fine without depending on the repositories first and remove them in a separate PR, this will help make the changes smaller and easier to review)


src/taipy/core/pipeline/_pipeline_manager.py
```diff
@@ -10,41 +10,43 @@
...
 class _PipelineManager(_Manager[Pipeline], _VersionMixin):
 
     _ENTITY_NAME = Pipeline.__name__
-    _repository: _AbstractRepository
     _EVENT_ENTITY_TYPE = EventEntityType.PIPELINE
 
     @classmethod
     def _get_all(cls, version_number: Optional[str] = None) -> List[Pipeline]:
         """
         Returns all entities.
```
jrobinAV:
```suggestion
        Returns all pipeline entities.
```

jrobinAV (commented):
sounds good to me at this point.

src/taipy/core/pipeline/_pipeline_manager.py
```diff
@@ -9,42 +9,60 @@
...
-        return cls._repository._load_all(filters)
+        pipelines = []
+        scenarios = _ScenarioManagerFactory._build_manager()._get_all(version_number)
+        for scenario in scenarios:
+            pipelines.extend(list(scenario.pipelines.values()))
+
+        return pipelines
+
+    @classmethod
+    def _get_all_by(cls, filters: Optional[List[Dict]] = None) -> List[Pipeline]:
```
trgiangdo: This method is weird imo.

We filter the scenario, then get all pipelines.
But what we should actually do is get all pipelines, then filter those pipelines.

However, this method did not exist before. Why do we need to add it?

src/taipy/core/pipeline/pipeline.py
```diff
@@ -126,7 +114,6 @@ def data_nodes(self) -> Dict[str, DataNode]:
         return data_nodes
 
     @property  # type: ignore
```
```suggestion
    @property
```
I think we can safely remove the `# type: ignore` here since there is only 1 decorator now.
Not sure though

src/taipy/core/pipeline/_pipeline_manager.py
```diff
@@ -90,42 +108,96 @@ def __remove_subscriber(cls, callback, params, pipeline):
...
+                Scenario._PIPELINE_PROPERTIES_KEY: pipeline._properties.data,
+            }
+            scenario._pipelines[pipeline_name] = pipeline_data
+            scenario_manager._set(scenario)
+        else:
+            cls._logger.error(f"Pipeline {pipeline.id} belongs to a non-existing Scenario {scenario_id}.")
+            raise PipelineBelongsToNonExistingScenario(pipeline.id, scenario_id)
+
+    @classmethod
+    def _breakdown_pipeline_id(cls, pipeline_id: str) -> Tuple[str, str]:
```
trgiangdo:
```suggestion
    def __breakdown_pipeline_id(cls, pipeline_id: str) -> Tuple[str, str]:
```
I think this should be a private method

src/taipy/core/pipeline/_pipeline_manager.py
trgiangdo, line 1: The order of the methods in this file is messy.

We should group all `get` methods, `create` method, `delete` methods, ... like other manager, then put all private methods at the end of the file.
What do you think?

src/taipy/core/pipeline/_pipeline_manager.py
```diff
@@ -164,13 +236,109 @@ def _submit(
...
+        """
+        pipeline_name, scenario_id = cls._breakdown_pipeline_id(pipeline_id)
+
+        if scenario := _ScenarioManagerFactory._build_manager()._get(scenario_id):
+            if pipeline_name in scenario._pipelines.keys():
+                scenario.remove_pipelines([pipeline_name])
+                if hasattr(cls, "_EVENT_ENTITY_TYPE"):
+                    _publish_event(cls._EVENT_ENTITY_TYPE, pipeline_id, EventOperation.DELETION, None)
+                return
+        raise ModelNotFound(cls._dir_name, pipeline_id)
```
trgiangdo: I think the way we raise `ModelNotFound` in SQLRepository is different, without the dir name

src/taipy/core/pipeline/_pipeline_manager.py
```diff
@@ -164,13 +236,109 @@ def _submit(
...
+        pipeline_name, scenario_id = cls._breakdown_pipeline_id(id)
+        pipeline = {"id": id, "owner_id": scenario_id, "parent_ids": [scenario_id], "name": pipeline_name}
+
+        scenario = _ScenarioManagerFactory._build_manager()._get(scenario_id)
+        if pipeline_data := scenario._pipelines.get(pipeline_name, None):
+            pipeline.update(pipeline_data)
+            with open(export_path, "w", encoding="utf-8") as export_file:
+                export_file.write(json.dumps(pipeline))
+        else:
+            raise ModelNotFound(cls._dir_name, id)
```
trgiangdo: I think the way we raise `ModelNotFound` in SQLRepository is different, without the dir name.

src/taipy/core/pipeline/_pipeline_manager.py
```diff
@@ -164,13 +236,109 @@ def _submit(
...
+                for pipeline_name in pipeline_names:
+                    del scenario._pipelines[pipeline_name]
+                scenario_manager._set(scenario)
+
+            if hasattr(cls, "_EVENT_ENTITY_TYPE"):
+                for pipeline_id in pipeline_ids:
+                    _publish_event(cls._EVENT_ENTITY_TYPE, pipeline_id, EventOperation.DELETION, None)
+        except (ModelNotFound, KeyError):
+            cls.__log_error_entity_not_found(pipeline_id)
+            raise ModelNotFound(cls._dir_name, pipeline_id)
```
trgiangdo: I think the way we raise `ModelNotFound` in SQLRepository is different, without the dir name

src/taipy/core/scenario/_scenario_manager.py
```diff
@@ -127,15 +124,19 @@ def _create(
             if config.additional_data_node_configs
             else {}
         )
+
         pipelines = {}
+        tasks_and_config_id_maps = {task.config_id: task for task in tasks}
         for sequence_name, sequence_task_configs in config.sequences.items():
```
trgiangdo, line 130: We've already used the term `sequence` in the Scenario config? I did not know that

tests/core/job/test_job.py
```diff
@@ -80,17 +81,17 @@ def _error():
     raise RuntimeError("Something bad has happened")
 
 
-def test_create_job(pipeline, task, job):
-    from src.taipy.core.pipeline._pipeline_manager_factory import _PipelineManagerFactory
+def test_create_job(scenario, task, job) -> None:
```
trgiangdo:
```suggestion
def test_create_job(scenario, task, job):
```
This is SUS. Copilot? 🤣

trgiangdo (commented):
I have some minor comments

src/taipy/core/pipeline/_pipeline_manager.py
toan-quach: haha I understand your confusion as I did too. The logic behind this is because we're storing Pipeline in ScenarioModel now, so we will need to filter ScenarioModel for data related to Pipeline and that's the reason behind why we filter the scenario 1st. 
As for the 2nd point, yep I don't think it's used anywhere aside from the test. But just to be sure that we stay consistent with the _Manager class, I think it'd have more benefit to implement them also instead of removing it.

src/taipy/core/pipeline/pipeline.py
toan-quach: ah agree thanks!

src/taipy/core/pipeline/_pipeline_manager.py
toan-quach: Actually kept is as _breakdown_pipeline_id so that I can import it for testing 😅  but let me see if we can do that and mypy will still be happy

src/taipy/core/pipeline/_pipeline_manager.py
toan-quach: that's a great suggestion! let me check that!

src/taipy/core/pipeline/_pipeline_manager.py
toan-quach: just checked, it's `ModelNotFound(str(self.model_type.__name__), entity_id)` I think the end result would still be the same or closely the same. I'll have a look to see if we can stay consistent or not

src/taipy/core/scenario/_scenario_manager.py
toan-quach: yep, but only in ScenarioConfig, so we will need to update it on other places as well after this ticket

tests/core/job/test_job.py
toan-quach: hahaha thanks! I think it's the Sonarlint suggestion extension :)

src/taipy/core/pipeline/_pipeline_manager.py
trgiangdo: If some pipelines have a custom attribute named `description` (which should not, but just assuming), and no scenario has any `description`, then `PipelineManager._get_all_by({"description": "foo"})` would return empty.

I don't think this is a safe behavior.

trgiangdo: Can we use the `PipelineManager._get_all()` method, and then filter from all the pipelines?

toan-quach: we can but that'd hurt the performance actually 😅

trgiangdo: We are not using it, so the performance is not a problem atm.

And we need to make sure that if we do use it, the behavior is correct.

toan-quach: hmmm, but we also have some share data of pipeline that is being stored in scenariomodel, so without filtering by the scenario, I'm not sure we can get the correct data either, like version for example is being shared with scenario. maybe the best solution is to filter by scenario first, then filter again by pipeline?

src/taipy/core/pipeline/_pipeline_manager.py
jrobinAV: that s fine to keep it with a single _

src/taipy/core/pipeline/_pipeline_manager.py
```diff
@@ -92,26 +93,33 @@ def _create(
...
-            {scenario_id} if scenario_id else None,
+            properties=properties,
+            tasks=_tasks,
+            pipeline_id=pipeline_id,
+            owner_id=scenario_id,
+            parent_ids={scenario_id} if scenario_id else None,
+            subscribers=subscribers,
             version=version,
         )
         for task in _tasks:
```
jrobinAV, line 172: We should ensure `pipeline._is_consistent` returns `True` after its creation.

src/taipy/core/pipeline/_pipeline_manager.py
jrobinAV: Here are my thoughts: 
The Pipeline manager is meant to deal with pipelines, no scenario. So the filter should apply to pipelines only. 
`PipelineManager._get_all_by({"description": "foo"})` should return all pipelines with a description equals to `foo`, whatever the scenario it belongs.

If we need to filter by both scenario attributes and pipeline attributes, it should be done either with another explicit method or in two steps.

tests/core/pipeline/test_pipeline_manager.py
```diff
@@ -39,10 +48,33 @@
...
+        _PipelineManager._breakdown_pipeline_id("PIPELINE_pipeline_id")
+    with pytest.raises(InvalidPipelineId):
+        _PipelineManager._breakdown_pipeline_id("SCENARIO_scenario_id")
+    with pytest.raises(InvalidPipelineId):
+        _PipelineManager._breakdown_pipeline_id("pipeline_SCENARIO_scenario_id")
+    with pytest.raises(InvalidPipelineId):
+        _PipelineManager._breakdown_pipeline_id("PIPELINE_pipeline_scenario_id")
+    pipeline_name, scenario_id = _PipelineManager._breakdown_pipeline_id("PIPELINE_pipeline_SCENARIO_scenario")
+    assert pipeline_name == "pipeline" and scenario_id == "SCENARIO_scenario"
+    pipeline_name, scenario_id = _PipelineManager._breakdown_pipeline_id("PIPELINEpipelineSCENARIO_scenario")
```
jrobinAV, line 66: I don't get why it is a valid id. We must have a separator between pipeline prefix and pipeline name + a separator between pipeline name and scenario ID. No?

src/taipy/core/exceptions/exceptions.py
```diff
@@ -125,13 +125,34 @@ class InvalidSubscriber(RuntimeError):
...
 
 
+class PipelineBelongsToNonExistingScenario(Exception):
+    """Raised if a Pipeline does not belong to an existing Scenario."""
+
+    def __init__(self, pipeline_id: str, scenario_id: str):
+        self.message = f"Pipeline: {pipeline_id} belongs to a non-existing Scenario: {scenario_id}."
+
+
+class PipelineTaskDoesNotExistInSameScenario(Exception):
```
jrobinAV, line 149: I would rename it: `PipelineTaskDoesNotExistInScenario`

tests/core/notification/test_notifier.py
```diff
@@ -250,9 +250,10 @@ def test_publish_event():
 
     scenario = tp.create_scenario(scenario_config)
     cycle = scenario.cycle
-    pipeline = scenario.pipelines["pipeline_config"]
     task = scenario.tasks[task_config.id]
     dn = scenario.data_nodes[dn_config.id]
+    # NOTE: pipeline is created only when accessed via scenario.pipelines
```
jrobinAV, line 255: The note seems not True anymore. No? We are checking the pipeline creation event later.

jrobinAV (commented):
We must propagate the changes in rest, taipy and enterprise before merging.

tests/core/pipeline/test_pipeline_manager.py
toan-quach: for the breakdown function, it didn't use the separator as a way to breakdown the id, so you can say that it ignores such cases

src/taipy/core/pipeline/_pipeline_manager.py
toan-quach: I have updated it to the solution proposed by Giang, to get all pipelines and check by the filter provided with getattr :D

src/taipy/core/pipeline/_pipeline_manager.py
toan-quach: This will be done in another ticket (https://app.zenhub.com/workspaces/taipy-613ef328affde6000ff77fb3/issues/gh/avaiga/taipy-core/734)

tests/core/notification/test_notifier.py
toan-quach: it's still True actually. When we call the getter `scenario.pipelines`, it will trigger the `__get_pipelines` function which then create all pipelines of the scenario and trigger the CREATE event

src/taipy/core/exceptions/exceptions.py
toan-quach: I'll update the naming in PR https://github.com/Avaiga/taipy-core/pull/731 since it's the PR for check if tasks exist in scenario

tests/core/notification/test_notifier.py
jrobinAV: AH ok!
Got it

jrobinAV (approved):

trgiangdo (approved):

Issue #449, opened 2023-01-27
New API to create Global Data Nodes


FlorianJacta:
**What would that feature address**
Creating Global Data Nodes and using them in a Taipy Core application will be simplified.

***Description of the ideal solution***
`tp.create_global_data_node(<Data Node Config>)` should create the Data Node.
The method should raise a dedicated error if the data node config is not global.

***Documentation***
- Add new entry in the release notes
- Add description in the user manual
- Add method to the Ref manual

jrobinAV:
Requires some design discussion

PR #730, opened 2023-08-15
Feature#449 - Add taipy.create_global_data_node() API

trgiangdo:
https://github.com/Avaiga/taipy-core/issues/449

### DIFF
--- a/src/taipy/core/exceptions/exceptions.py
+++ b/src/taipy/core/exceptions/exceptions.py
@@ -96,7 +96,7 @@ class DataNodeConfigIsNotGlobal(Exception):
     """Raised if a DataNode is not global."""
 
     def __init__(self, data_node_config_id: str):
-        self.message = f"Data node config `{data_node_config_id}` does not GLOBAL scope."
+        self.message = f"Data node config `{data_node_config_id}` is not GLOBAL scope."
 
 
 class MissingReadFunction(Exception):
--- a/src/taipy/core/exceptions/exceptions.py
+++ b/src/taipy/core/exceptions/exceptions.py
@@ -96,7 +96,7 @@ class DataNodeConfigIsNotGlobal(Exception):
     """Raised if a DataNode is not global."""
 
     def __init__(self, data_node_config_id: str):
-        self.message = f"Data node config `{data_node_config_id}` is not GLOBAL scope."
+        self.message = f"Data node config `{data_node_config_id}` does not have GLOBAL scope."
 
 
 class MissingReadFunction(Exception):
--- a/src/taipy/core/exceptions/exceptions.py
+++ b/src/taipy/core/exceptions/exceptions.py
@@ -92,6 +92,13 @@ class InvalidCustomDocument(Exception):
     """Raised if an invalid custom document class is provided to a `MongoCollectionDataNode`."""
 
 
+class DataNodeConfigIsNotGlobal(Exception):
+    """Raised if a DataNode is not global."""
+
+    def __init__(self, data_node_config_id: str):
+        self.message = f"Data node config `{data_node_config_id}` does not GLOBAL scope."
+
+
 class MissingReadFunction(Exception):
     """Raised if no read function is provided for the GenericDataNode."""
 
--- a/src/taipy/core/taipy.py
+++ b/src/taipy/core/taipy.py
@@ -14,11 +14,13 @@ import shutil
 from datetime import datetime
 from typing import Any, Callable, Dict, List, Optional, Set, Union, overload
 
+from taipy.config.common.scope import Scope
 from taipy.logger._taipy_logger import _TaipyLogger
 
 from ._entity._entity import _Entity
 from ._version._version_manager_factory import _VersionManagerFactory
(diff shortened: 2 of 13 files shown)


src/taipy/core/exceptions/exceptions.py
```diff
@@ -92,6 +92,13 @@ class InvalidCustomDocument(Exception):
     """Raised if an invalid custom document class is provided to a `MongoCollectionDataNode`."""
 
 
+class DataNodeConfigIsNotGlobal(Exception):
+    """Raised if a DataNode is not global."""
+
+    def __init__(self, data_node_config_id: str):
+        self.message = f"Data node config `{data_node_config_id}` does not GLOBAL scope."
```
toan-quach:
```suggestion
        self.message = f"Data node config `{data_node_config_id}` is not GLOBAL scope."
```

toan-quach: or `does not have` 😅

toan-quach (dismissed):
lgtm, minor remark

src/taipy/core/exceptions/exceptions.py
```diff
@@ -92,6 +92,13 @@ class InvalidCustomDocument(Exception):
     """Raised if an invalid custom document class is provided to a `MongoCollectionDataNode`."""
 
 
+class DataNodeConfigIsNotGlobal(Exception):
+    """Raised if a DataNode is not global."""
+
+    def __init__(self, data_node_config_id: str):
+        self.message = f"Data node config `{data_node_config_id}` is not GLOBAL scope."
```
trgiangdo:
```suggestion
        self.message = f"Data node config `{data_node_config_id}` does not have GLOBAL scope."
```

toan-quach (dismissed):

src/taipy/core/taipy.py
```diff
@@ -577,6 +584,23 @@ def create_scenario(
...
+    Returns:
+        The new global data node.
+    Raises:
+        DataNodeConfigIsNotGlobal^: If the data node configuration does not have GLOBAL scope.
+    """
+    # Check if the data node config has GLOBAL scope
+    if config.scope is not Scope.GLOBAL:
+        raise DataNodeConfigIsNotGlobal(config.id)  # type: ignore
+
+    return _DataManagerFactory._build_manager()._create_and_set(config, None, None)
```
jrobinAV, line 603: Instead of using `_create_and_set` should we use `_bulk_get_or_create` method?
Because we don't want to create the data node twice if it already exists.

```suggestion
    return _DataManagerFactory._build_manager()._bulk_get_or_create([config])
```

tests/test_taipy.py
```diff
@@ -435,6 +434,20 @@ def test_cycle_exists(self):
...
+            mck.assert_called_once_with(dn_cfg, None, None)
+
+        dn = tp.create_global_data_node(dn_cfg)
+        assert dn.scope == Scope.GLOBAL
+        assert dn.config_id == dn_cfg.id
+
+        dn_cfg.scope = Scope.SCENARIO
+        with pytest.raises(DataNodeConfigIsNotGlobal):
+            tp.create_global_data_node(dn_cfg)
+
```
jrobinAV, line 455: What happens when we create two data nodes from the same GLOBAL data node config ?

jrobinAV (changes requested):

src/taipy/core/taipy.py
trgiangdo: You are right, I will update this and add a unittest

trgiangdo: Actually @jrobinAV, I think since it is a global data node, the user can have 2 data nodes from the same configuration if he want.

We don't need to restrict this.

jrobinAV: But that's not the meaning of `GLOBAL` scope data node config. The meaning is that it is unique and shared by all scenarios. Two scenarios from 2 different cycles (or not) should share the same data node if it is `GLOBAL`.
Plus, that's not how it works when creating the data nodes from scenario creation. When creating the scenario with a `GLOBAL` data node, we look for the data node if it exists. If the data node is created before the scenario, the scenario will re-use it. Which one should we select if multiple `GLOBAL` data nodes from the same config exist?

I would keep the same behavior whatever we create the data node from `create_global_data_node` or from `create_scenario`.

trgiangdo: Right, that makes sense. I will push the fix shortly.

trgiangdo:
I added 2 new commits.
1. [First commit](https://github.com/Avaiga/taipy-core/pull/730/commits/2d2a99fe51382876a52e5d08647587c0274a5413) resolves the case that we don't want to create the data node twice if it already exists. I did not use `_bulk_get_or_create()` because we are only creating 1 entity, not bulk.
2. [Second commit](https://github.com/Avaiga/taipy-core/pull/730/commits/0fdc4561e487b699e95269860049b50b8fc64f6a) fixes the issue in https://github.com/Avaiga/taipy-core/issues/738

jrobinAV (approved):

Issue #695, opened 2023-07-20
Verifying tasks exist in Scenario entity when adding Sequence through add_sequence


toan-quach:
**Description**
When adding sequences/pipelines to a scenario entity through the function add_sequences/add_pipelines, check if the Task entities used here exists in the scenario.tasks

PR #731, opened 2023-08-17
feature/#695 added check for pipeline task exist in scenario

toan-quach:
#695 

This PR is set to Draft only and please don't merge as it will merge directly to branch `feature/#718-stop-storing-PipelineModel` instead of `develop`. I will create a new PR later to merge directly to develop after PR [#722](https://github.com/Avaiga/taipy-core/pull/722) has been merged

### DIFF
--- a/src/taipy/core/exceptions/exceptions.py
+++ b/src/taipy/core/exceptions/exceptions.py
@@ -10,6 +10,9 @@
 # specific language governing permissions and limitations under the License.
 
 
+from typing import List, Union
+
+
 class CycleAlreadyExists(Exception):
     """Raised if it is trying to create a Cycle that has already exists."""
 
@@ -147,10 +150,17 @@ class PipelineBelongsToNonExistingScenario(Exception):
 
 
 class PipelineTaskDoesNotExistInSameScenario(Exception):
-    """Raised if the Task of a Pipeline does not exist in the same Scenario."""
+    """Raised if the Tasks of a Pipeline does not exist in the same Scenario."""
 
-    def __init__(self, task_id: str, pipeline_id: str, scenario_id: str):
-        self.message = f"Task {task_id} of Pipeline {pipeline_id} does not exist in Scenario {scenario_id}."
+    def __init__(self, task_ids: List[Union[str, None]], pipeline_name: str, scenario_id: str):
+        self.message = f"Tasks {task_ids} of Pipeline {pipeline_name} does not exist in Scenario {scenario_id}."
+
+
+class PipelineTaskConfigDoesNotExistInSameScenarioConfig(Exception):
+    """Raised if the TaskConfigs of a Pipeline does not exist in the same ScenarioConfig."""
+
+    def __init__(self, task_config_ids: List[Union[str, None]], pipeline_name: str, scenario_config_id: str):
+        self.message = f"TaskConfig {task_config_ids} of Pipeline name {pipeline_name} does not exist in ScenarioConfig {scenario_config_id}."
 
 
 class NonExistingPipelineConfig(Exception):
--- a/src/taipy/core/exceptions/exceptions.py
+++ b/src/taipy/core/exceptions/exceptions.py
@@ -10,7 +10,7 @@
 # specific language governing permissions and limitations under the License.
 
 
-from typing import List, Union
+from typing import List, Optional
 
 
 class CycleAlreadyExists(Exception):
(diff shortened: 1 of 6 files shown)


src/taipy/core/exceptions/exceptions.py
```diff
@@ -147,10 +150,17 @@ def __init__(self, pipeline_id: str, scenario_id: str):
 
 
 class PipelineTaskDoesNotExistInSameScenario(Exception):
-    """Raised if the Task of a Pipeline does not exist in the same Scenario."""
+    """Raised if the Tasks of a Pipeline does not exist in the same Scenario."""
```
trgiangdo:
```suggestion
    """Raised if Tasks of a Pipeline do not exist in the same Scenario that the Pipeline belongs to."""
```
Or
```suggestion
    """Raised if there is any Task of a Pipeline does not exist in the same Scenario that the Pipeline belongs to."""
```

src/taipy/core/exceptions/exceptions.py
```diff
@@ -147,10 +150,17 @@ def __init__(self, pipeline_id: str, scenario_id: str):
...
+
+    def __init__(self, task_ids: List[Union[str, None]], pipeline_name: str, scenario_id: str):
```
trgiangdo:
```suggestion
    def __init__(self, task_ids: List[str], pipeline_name: str, scenario_id: str):
```
The type hint here is weird, it should not be a List of None. Is this what you meant?

src/taipy/core/exceptions/exceptions.py
```diff
@@ -147,10 +150,17 @@ def __init__(self, pipeline_id: str, scenario_id: str):
...
+        self.message = f"Tasks {task_ids} of Pipeline {pipeline_name} does not exist in Scenario {scenario_id}."
+
+
+class PipelineTaskConfigDoesNotExistInSameScenarioConfig(Exception):
+    """Raised if the TaskConfigs of a Pipeline does not exist in the same ScenarioConfig."""
```
trgiangdo:
```suggestion
    """Raised if TaskConfigs of a Pipeline do not exist in the same ScenarioConfig that the Pipeline belongs to."""
```
Or 
```suggestion
    """Raised if there is any TaskConfig of a Pipeline does not exist in the same ScenarioConfig that the Pipeline belongs to."""
```

src/taipy/core/exceptions/exceptions.py
```diff
@@ -147,10 +150,17 @@ def __init__(self, pipeline_id: str, scenario_id: str):
...
 
-    def __init__(self, task_id: str, pipeline_id: str, scenario_id: str):
-        self.message = f"Task {task_id} of Pipeline {pipeline_id} does not exist in Scenario {scenario_id}."
+    def __init__(self, task_config_ids: List[Union[str, None]], pipeline_name: str, scenario_config_id: str):
```
trgiangdo:
```suggestion
    def __init__(self, task_config_ids: List[str], pipeline_name: str, scenario_config_id: str):
```

tests/core/scenario/test_scenario.py
```diff
@@ -226,6 +225,51 @@ def get_label(self):
...
+
+    scenario.tasks = [task_1]
+
+    scenario.add_pipelines({"pipeline_2": {"tasks": [task_1]}})
+
+    with pytest.raises(PipelineTaskDoesNotExistInSameScenario):
+        scenario.add_pipelines({"pipeline_3": {"tasks": [task_2]}})
+
+    with pytest.raises(PipelineTaskDoesNotExistInSameScenario):
+        scenario.add_pipelines({"pipeline_4": {"tasks": [task_1, task_2]}})
```
trgiangdo: I think we should also check for the message of the Exception.

```suggestion
    with pytest.raises(PipelineTaskDoesNotExistInSameScenario) as error:
        scenario.add_pipelines({"pipeline_3": {"tasks": [task_2]}})
        assert "[task_2] ..." in error.message

    with pytest.raises(PipelineTaskDoesNotExistInSameScenario) as error:
        scenario.add_pipelines({"pipeline_4": {"tasks": [task_1, task_2]}})
        assert "[task_2] ..." in error.message
        assert "task_1" not in error.message
```

tests/core/scenario/test_scenario_manager.py
```diff
@@ -220,6 +220,32 @@ def test_set_and_get_scenario(cycle):
...
+
+    scenario_config_1._tasks = [task_config_1]
+    _ScenarioManager._create(scenario_config_1)
+
+    scenario_config_1.add_sequences({"pipeline_2": [task_config_1]})
+    _ScenarioManager._create(scenario_config_1)
+
+    scenario_config_1.add_sequences({"pipeline_3": [task_config_1, task_config_2]})
+    with pytest.raises(PipelineTaskConfigDoesNotExistInSameScenarioConfig):
+        _ScenarioManager._create(scenario_config_1)
```
trgiangdo: Same here, I think we should also check for the message of the exception

trgiangdo (commented):
I have a few comments

src/taipy/core/exceptions/exceptions.py
toan-quach: ah you're right, I think it somehow got messed up with some of the automation tool for type hint

toan-quach: I can't seem to get mypy to agree with me after several if-else guaranteeing the list will not be None, so I settled with `task_ids: List[Optional[str]]` and `task_config_ids: List[Optional[str]]`

jrobinAV (dismissed):

trgiangdo (approved):

PR #735, opened 2023-08-22
Feature/#684 rename pipeline to sequence

toan-quach:
https://app.zenhub.com/workspaces/taipy-613ef328affde6000ff77fb3/issues/gh/avaiga/taipy-core/684


jrobinAV (dismissed):

jrobinAV (dismissed):

src/taipy/core/config/__init__.py
```diff
@@ -68,12 +68,12 @@
...
+    SequenceConfig,
+    "sequences",
+    SequenceConfig.default_config(),
     [
-        ("configure_pipeline", PipelineConfig._configure),
-        ("set_default_pipeline_configuration", PipelineConfig._set_default_configuration),
+        ("configure_sequence", SequenceConfig._configure),
+        ("set_default_sequence_configuration", SequenceConfig._set_default_configuration),
     ],
 )
```
trgiangdo, line 78: This part should not exist anymore, no?

src/taipy/core/config/__init__.py
```diff
@@ -105,5 +105,5 @@
 _Checker.add_checker(_CoreSectionChecker)
 _Checker.add_checker(_DataNodeConfigChecker)
 _Checker.add_checker(_TaskConfigChecker)
-_Checker.add_checker(_PipelineConfigChecker)
+_Checker.add_checker(_SequenceConfigChecker)
```
trgiangdo, line 108: Same for the checker, we should have removed it already right?

src/taipy/core/config/scenario_config.py
```diff
@@ -169,12 +169,12 @@ def _from_dict(
...
         else:
-            pipeline_config_ids = as_dict.pop(cls._PIPELINES_KEY, list())
+            sequence_config_ids = as_dict.pop(cls._SEQUENCES_KEY, list())
             cls.__logger.info(
-                f"The tasks from these PipelineConfig {pipeline_config_ids}"
+                f"The tasks from these SequenceConfig {sequence_config_ids}"
                 f"will be migrated to be directly under ScenarioConfig {id}"
             )
-            tasks = cls.__get_task_configs_from_pipeline_configs(pipeline_config_ids, config)
+            tasks = cls.__get_task_configs_from_sequence_configs(sequence_config_ids, config)
```
trgiangdo, line 177: I have a question. If we rename the PipelineConfig to SequenceConfig, then user's old config should not work anymore. So is this `else` block necessary anymore?

src/taipy/core/config/sequence_config.py
```diff
@@ -86,41 +86,39 @@ def _update(self, as_dict, default_section=None):
...
+        This function creates the *default sequence configuration* object,
+        where all sequence configuration objects will find their default
         values when needed.
 
         Parameters:
             task_configs (Union[TaskConfig^, List[TaskConfig^]]): The list of the task
-                configurations that make the default pipeline configuration. This can be
-                a single task configuration object is this pipeline holds a single task.
+                configurations that make the default sequence configuration. This can be
+                a single task configuration object is this sequence holds a single task.
```
trgiangdo: I have no idea what does this sentence mean 😅

trgiangdo (commented):
I have some concerns regarding the Pipeline/SequenceConfig

src/taipy/core/config/__init__.py
toan-quach: we're postponing the removal of Sequence config, repo and model to make sure everything is still working fine without any dependency on those classes. The plan is to remove them in PR https://github.com/Avaiga/taipy-core/pull/739

src/taipy/core/config/__init__.py
toan-quach: same answer as above

src/taipy/core/config/scenario_config.py
toan-quach: that's true. the code for migrating from old configs to new configs are no longer valid. @jrobinAV I think we will need to discuss about this more

toan-quach: either we need a new logic for migrating old PipelineConfig to new SequenceConfig, or we remove it in core and move this logic to migration script. Either way, we will still need new logic as Scenario is still using the PipelineConfig to pull task configs when none is provided. And regardless, of the solution, I think we can implement it on this PR instead https://github.com/Avaiga/taipy-core/pull/739. To keep this PR strictly for renaming only with no logic changed. Let me know what you guys think @trgiangdo @jrobinAV

trgiangdo (approved):

jrobinAV (approved):

Issue #732, opened 2023-08-21
Add encoding option for CSV, JSON Data node configuration


trgiangdo:
**What would that feature address**

For now, there is no way for the user to manipulate the encoding when reading a file of a CSV and JSON DataNode.

Pandas and (almost) all excel-related libraries do not support explicit encoding types for parquet and excel files, so it is not possible in the near future.
Pickle Data node is not affected by the encoding, because Pickle is always binary file.

***Description of the ideal solution***
The user should be able to configure
```python
json_dn_cfg = Config.configure_json_data_node("dn_id", "path.json", encoding="utf-8")
```
and then, all reading/writing from the `json_dn_cfg` should use UTF-8 encoding.

TODO:
- [x] Implementation the feature.
- [x] Add read/write test cases.
- [x] Update the Configuration documentation of the User Manual.
- [x] Update the release notes.

PR #737, opened 2023-08-23
Feature/#732 - Add encoding option to CSV and JSON DataNode

trgiangdo:
https://github.com/Avaiga/taipy-core/issues/732

In this PR, the encoding option is only added to CSV and JSON DataNode.

For Parquet, we need to use the method provided by pandas since we need to read folder as well, but there is no "encoding" parameter supported.

For Excel, somehow the "encoding" parameter of pandas.read_excel() was dropped since pandas 1.0.1.
And if I try to open an excel file with encoding UTF-xx, I can't, because Excel are binary file and there will be always some sort of special unicode characters of Excel.

### DIFF
--- a/src/taipy/core/config/data_node_config.py
+++ b/src/taipy/core/config/data_node_config.py
@@ -82,6 +82,10 @@ class DataNodeConfig(Section):
         _EXPOSED_TYPE_MODIN,
         _EXPOSED_TYPE_NUMPY,
     ]
+
+    _ENCODING_TYPE_KEY = "encoding"
+    _DEFAULT_ENCODING_TYPE = "utf-8"
+
     # Generic
     _OPTIONAL_READ_FUNCTION_GENERIC_PROPERTY = "read_fct"
     _OPTIONAL_READ_FUNCTION_ARGS_GENERIC_PROPERTY = "read_fct_args"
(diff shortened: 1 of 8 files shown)


src/taipy/core/config/data_node_config.py
```diff
@@ -82,6 +82,10 @@ class DataNodeConfig(Section):
         _EXPOSED_TYPE_MODIN,
         _EXPOSED_TYPE_NUMPY,
     ]
+
+    _ENCODING_TYPE_KEY = "encoding"
```
jrobinAV: Rename to `_OPTIONAL_ENCODING_PROPERTY`

src/taipy/core/config/data_node_config.py
```diff
@@ -82,6 +82,10 @@ class DataNodeConfig(Section):
...
+    _DEFAULT_ENCODING_TYPE = "utf-8"
```
jrobinAV: Rename to `_DEFAULT_ENCODING_VALUE`

src/taipy/core/data/csv.py
```diff
@@ -82,17 +85,20 @@ def __init__(
...
-        properties: Dict = None,
+        properties: Optional[Dict] = None,
     ):
         if properties is None:
             properties = {}
 
         default_value = properties.pop(self.__DEFAULT_DATA_KEY, None)
 
+        if self.__ENCODING_KEY not in properties.keys():
+            properties[self.__ENCODING_KEY] = "utf-8"
```
jrobinAV, line 100: Can you use a variable for "utf-8", like `__ENCODING_DEFAULT_VALUE`

jrobinAV (dismissed):
minor remarks

src/taipy/core/data/csv.py
trgiangdo: I was thinking about it, but we do not use any kind of `__HAS_HEADER_DEFAULT` or something like that at the DataNode level.

I think it is because this is just safety layer, because a DataNode should be created from a Configuration, and the default value is set at the config already.

If we use `__ENCODING_DEFAULT`, we should do the same for other parameter as well. Let me know what you think

jrobinAV: Yes, you are right. No need for it, then.

jrobinAV (approved):

Issue #734, opened 2023-08-21
Investigate the use of `is_consistent` in Pipeline


toan-quach:
**Description**
Discuss and investigate the use of `is_consistent` in Pipeline:

- Does a pipeline have to be connected? Or it's okay as long as it's a DAG or DAGs?
- Does the scenario need to have a consistency check function as well? If yes, does it have to be connected or it's okay just being a DAG or DAGs?

At least the graph should be acyclic for both pipeline and scenarios.

toan-quach:
@jrobinAV 
I have added the questions we had during the discussion for this topic in the description.

toan-quach:
@jrobinAV 
I have added the questions we had during the discussion for this topic in the description.

toan-quach:
@jrobinAV 

To conclude from our discussion from before:
- For Sequence, we will check if a Sequence is a "connected DAG", if not, we will raise an error
- For Scenario, we will check if a Scenario is a DAG or not (it can be connected or not connected)

jrobinAV:
we will also check that the DAG is made of Task and Data nodes only and that a task is not linked to other tasks and data node not linked to other data nodes.

PR #745, opened 2023-08-31
feature/#734 apply is-consistent to sequence and scenario

toan-quach:
#734

### DIFF
--- a/src/taipy/core/scenario/scenario.py
+++ b/src/taipy/core/scenario/scenario.py
@@ -34,6 +34,7 @@ from ..data._data_manager_factory import _DataManagerFactory
 from ..data.data_node import DataNode
 from ..data.data_node_id import DataNodeId
 from ..exceptions.exceptions import (
+    InvalidSequence,
     NonExistingDataNode,
     NonExistingSequence,
     NonExistingTask,
@@ -144,7 +145,7 @@ class Scenario(_Entity, Submittable, _Labeled):
         if protected_attribute_name in self._properties:
             return _tpl._replace_templates(self._properties[protected_attribute_name])
 
-        sequences = self.__get_sequences()
+        sequences = self._get_sequences()
         if protected_attribute_name in sequences:
             return sequences[protected_attribute_name]
         tasks = self.tasks
@@ -158,7 +159,7 @@ class Scenario(_Entity, Submittable, _Labeled):
     @property  # type: ignore
     @_self_reload(_MANAGER_NAME)
     def sequences(self) -> Dict[str, Sequence]:
-        return self.__get_sequences()
+        return self._get_sequences()
 
     @sequences.setter  # type: ignore
     @_self_setter(_MANAGER_NAME)
@@ -166,6 +167,10 @@ class Scenario(_Entity, Submittable, _Labeled):
         self, sequences: Dict[str, Dict[str, Union[List[Task], List[TaskId], _ListAttributes, List[_Subscriber], Dict]]]
     ):
         self._sequences = sequences
+        actual_sequences = self._get_sequences()
+        for sequence_name in sequences.keys():
+            if not actual_sequences[sequence_name]._is_consistent():
+                raise InvalidSequence(actual_sequences[sequence_name].id)
 
     def add_sequences(self, sequences: Dict[str, Dict[str, Union[List[Task], List[TaskId], List[_Subscriber], Dict]]]):
         """Add sequences to the scenario.
@@ -188,6 +193,11 @@ class Scenario(_Entity, Submittable, _Labeled):
 
         _sequences.update(sequences)
         self.sequences = _sequences  # type: ignore
+
+        actual_sequences = self._get_sequences()
+        for sequence_name in sequences.keys():
+            if not actual_sequences[sequence_name]._is_consistent():
+                raise InvalidSequence(actual_sequences[sequence_name].id)
 
     def remove_sequences(self, sequence_names: List[str]):
         """Remove sequences from the scenario.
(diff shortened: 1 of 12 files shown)


src/taipy/core/sequence/_sequence_manager.py
```diff
@@ -173,6 +174,10 @@ def _create(
             if sequence_id not in task._parent_ids:
                 task._parent_ids.update([sequence_id])
                 task_manager._set(task)
+
+        if not sequence._is_consistent():
+            # TODO: add test check raise this exception
+            raise InvalidSequence(sequence_id)
```
jrobinAV: Since the `_create` method is only used when reading the sequence, we should also use `sequence._is_consistent` method when adding or changing a scenario sequence.

src/taipy/core/sequence/sequence.py
```diff
@@ -130,14 +130,28 @@ def _is_consistent(self) -> bool:
         dag = self._build_dag()
         if not nx.is_directed_acyclic_graph(dag):
             return False
-        is_data_node = True
+        if dag.number_of_nodes() != 0 and not nx.is_connected(dag.to_undirected(as_view=True)):
```
jrobinAV: I finally found this `nx.is_weakly_connected(dag)`.
[weakly connectivity](https://networkx.org/documentation/stable/reference/algorithms/generated/networkx.algorithms.components.weakly_connected_components.html#networkx.algorithms.components.weakly_connected_components)

src/taipy/core/scenario/scenario.py
```diff
@@ -499,3 +512,12 @@ def get_simple_label(self) -> str:
...
         return self._get_simple_label()
+
+    def _is_consistent(self) -> bool:
+        dag = self._build_dag()
+
+        if dag.number_of_nodes() == 0:
+            return True
+        if not nx.is_directed_acyclic_graph(dag):
+            return False
+        return True
```
jrobinAV: I came up with this using stack overflow and the following links : 
[subgraph](https://networkx.org/documentation/stable/reference/classes/generated/networkx.Graph.subgraph.html)
[weakly connectivity](https://networkx.org/documentation/stable/reference/algorithms/generated/networkx.algorithms.components.weakly_connected_components.html#networkx.algorithms.components.weakly_connected_components)

**I have not tested it at all**

```suggestion
    def _is_consistent(self) -> bool:
        dag = self._build_dag()

        if dag.number_of_nodes() == 0:
            return True
        if not nx.is_directed_acyclic_graph(dag):
            return False
        for con_comp_nodes in nx.weakly_connected_components(dag):
            sub_dag = dag.subgraph(con_comp_nodes).copy()
            current_node_type: Any = None
            for nodes in nx.topological_generations(sub_dag):
                for node in nodes:
                    if is_data_node and not isinstance(node, DataNode):
                    if not current_node_type:
                        if isinstance(node, Task):
                            current_node_type = Task
                        elif isinstance(node, DataNode):
                            current_node_type = DataNode
                        else:
                            return False
                    if not isinstance(node, current_node_type):
                        return False
                    if not is_data_node and not isinstance(node, Task):
                        return False
                is_data_node = not is_data_node

                if issubclass(current_node_type, Task):
                    current_node_type = DataNode
                elif issubclass(current_node_type, DataNode):
                    current_node_type = Task
                else:
                    return False
        return True
```

tests/core/sequence/test_sequence.py
```diff
@@ -135,8 +135,15 @@ class FakeDataNode:
...
-    assert not sequence_2._is_consistent()
+    sequence_5 = Sequence({}, [task_5_1, task_5_2], "name_5")
+    assert not sequence_5._is_consistent()
+
+    intermediate_6 = DataNode("foo", Scope.SCENARIO, "intermediate_id_6")
+    output_6 = DataNode("foo", Scope.SCENARIO, "output_id_6")
+    task_6_1 = Task("foo", {}, print, [], [intermediate_6], TaskId("task_id_6_1"))
+    task_6_2 = Task("foo", {}, print, [intermediate_6], [output_6], TaskId("task_id_6_2"))
+    sequence_6 = Sequence({}, [task_6_1, task_6_2], "name_6")
+    assert sequence_6._is_consistent()
```
jrobinAV, line 146:
```suggestion
    assert sequence_6._is_consistent()
    
    input_7 = DataNode("foo", Scope.SCENARIO, "output_id_7")
    intermediate_7 = DataNode("foo", Scope.SCENARIO, "intermediate_id_7")
    task_7_1 = Task("foo", {}, print, [input_7 ], [intermediate_7], TaskId("task_id_7_1"))
    task_7_2 = Task("foo", {}, print, [intermediate_7], [], TaskId("task_id_7_2"))
    sequence_7 = Sequence({}, [task_7_1, task_7_2], "name_7")
    assert sequence_7._is_consistent()
```

src/taipy/core/scenario/scenario.py
toan-quach: I think it will be faster to just check if all the edges are connecting a Task to a DN, no? Here's my version of it:
```
    def _is_consistent(self) -> bool:
        dag = self._build_dag()

        if dag.number_of_nodes() == 0:
            return True
        if not nx.is_directed_acyclic_graph(dag):
            return False
        for left_node, right_node in dag.edges:
            if (isinstance(left_node, DataNode) and isinstance(right_node, Task)) or (isinstance(left_node, Task) and isinstance(right_node, DataNode)):
                continue
            return False
        return True
```

jrobinAV: Oh yes !!!
Much better !

jrobinAV (approved):

Issue #685, opened 2023-07-17
Update config.schema.json to match new pipeline stuffs


jrobinAV:
taipy/core/config/config.schema.json

- Ensure and discuss compatibility with studio-config

FredLL-Avaiga:
We should discuss how we handle different versions of the schema within taipy-studio...

PR #749, opened 2023-09-06
feature/#685 udpate changes to config schema json

toan-quach:
#685

### DIFF
--- a/src/taipy/core/config/config.schema.json
+++ b/src/taipy/core/config/config.schema.json
@@ -363,25 +363,6 @@
         }
       }
     },
-    "SEQUENCE": {
-      "description": "Sequences",
-      "type": "object",
-      "properties": {},
-      "required": [],
-      "additionalProperties": {
-        "type": "object",
-        "properties": {
-          "tasks": {
-            "description": "The list of tasks configurations.",
-            "type": "array",
-            "items": {
-              "type": "string"
-            },
-            "uniqueItems": true
-          }
-        }
-      }
-    },
     "SCENARIO": {
       "description": "Scenarios",
       "type": "object",
@@ -390,8 +371,16 @@
       "additionalProperties": {
         "type": "object",
         "properties": {
-          "sequences": {
-            "description": "The list of sequence configs.",
+          "tasks": {
+            "description": "The list of task configs.",
+            "type": "array",
+            "items": {
+              "type": "string"
+            },
+            "uniqueItems": true
+          },
+          "additional_data_nodes": {
+            "description": "The list of additional data node configs.",
             "type": "array",
             "items": {
               "type": "string"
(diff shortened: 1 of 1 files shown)


src/taipy/core/config/config.schema.json
```diff
@@ -417,6 +406,14 @@
               "type": "string",
               "taipy_function": true
             }
+          },
+          "sequences": {
+            "description": "The dictionary of sequence sequence descriptions.",
```
jrobinAV:
```suggestion
            "description": "The dictionary of sequence descriptions.",
```

jrobinAV (dismissed):

src/taipy/core/config/config.schema.json
toan-quach: oops!

jrobinAV (approved):

src/taipy/core/config/config.schema.json
```diff
@@ -417,6 +406,14 @@
               "type": "string",
               "taipy_function": true
             }
+          },
+          "sequences": {
+            "description": "The dictionary of sequence descriptions.",
```
FredLL-Avaiga, line 411: is this a dictionary of a list of unique strings ?

src/taipy/core/config/config.schema.json
```diff
@@ -417,6 +406,14 @@
...
+            "type": "object",
+            "items": {
+              "type": "string"
+            },
+            "uniqueItems": true
```
FredLL-Avaiga, line 416: Should we add a version in the TAIPY block ?

FredLL-Avaiga (approved):
looks good to me
Wondering about the type of sequences ...

src/taipy/core/config/config.schema.json
toan-quach: it's a dictionary :D

toan-quach: this is the signature for the sequence in ScenarioConfig: `sequences: Optional[Dict[str, List[TaskConfig]]] = None`
It takes a str for sequence name as the key and list of TaskConfig as the value

src/taipy/core/config/config.schema.json
toan-quach: I dont have any comments on that :/ @jrobinAV what do you think?

src/taipy/core/config/config.schema.json
FredLL-Avaiga: so how would that translate in a TOML file ?
I suppose it could be a section or an inline object ?

FredLL-Avaiga: I suppose for the schema they both translate into another (sequence|anonymous) Object ...

toan-quach: yep it's something like this 
image
or we can declare it in a separate block for sequence of a scenario
image

src/taipy/core/config/config.schema.json
jrobinAV: I don't know yet, because we will need to maintain this. Can we discuss it in a dedicated discussion in a separate ticket/PR?

Issue #740, opened 2023-08-25
make _track_edit public


jrobinAV:
- [x] remove underscore in the name of the method
- [ ] improve ref manual documentation
- [ ] add release notes
- [ ] add new entry in the user manual

PR #751, opened 2023-09-07
feature/#740 rename _track_edit to track_edit

toan-quach:
#740

### DIFF
--- a/src/taipy/core/data/data_node.py
+++ b/src/taipy/core/data/data_node.py
@@ -63,7 +63,7 @@ class DataNode(_Entity, _Labeled):
         edits (List[Edit^]): The list of Edits (an alias for dict) containing metadata about each
             data edition including but not limited to timestamp, comments, job_id:
             timestamp: The time instant of the writing
-            comments: Representation of a free text to explain or comment on a change of a data node
+            comments: Representation of a free text to explain or comment on a data change
             job_id: Only populated when the data node is written by a task execution and corresponds to the job's id.
             Additional metadata related to the edition made to the data node can also be provided in Edits.
         version (str): The string indicates the application version of the data node to
--- a/src/taipy/core/data/data_node.py
+++ b/src/taipy/core/data/data_node.py
@@ -61,7 +61,7 @@ class DataNode(_Entity, _Labeled):
         parent_ids (Optional[Set[str]]): The set of identifiers of the parent tasks.
         last_edit_date (datetime): The date and time of the last modification.
         edits (List[Edit^]): The list of Edits (an alias for dict) containing metadata about each
-            edition of that node including but not limited to timestamp, comments, job_id:
+            data edition including but not limited to timestamp, comments, job_id:
             timestamp: The time instant of the writing
             comments: Representation of a free text to explain or comment on a change of a data node
             job_id: Only populated when the data node is written by a task execution and corresponds to the job's id.
--- a/src/taipy/core/data/data_node.py
+++ b/src/taipy/core/data/data_node.py
@@ -325,7 +325,12 @@ class DataNode(_Entity, _Labeled):
         _DataManagerFactory._build_manager()._set(self)
 
     def track_edit(self, **options):
-        """Add edit tracking information to this data node."""
+        """Add edit tracking information to this data node.
+
+        Parameters:
+            options: track `timestamp`, `comments`, `job_id`. The others are user-custom, users can use options
+                to attach any information to an external edit of a data node.
+        """
         edit = {}
         for k, v in options.items():
             if v is not None:
--- a/src/taipy/core/data/data_node.py
+++ b/src/taipy/core/data/data_node.py
@@ -359,7 +359,7 @@ class DataNode(_Entity, _Labeled):
         join operator (_AND_ or _OR_).
 
         Parameters:
-            operators (Union[List[Tuple], Tuple]): 3-tuples or list of 3-tuples (key, value, `Operator^`).
+            operators (Union[List[Tuple], Tuple]): A 3-element tuple or a list of 3-element tuples, each is in the form of (key, value, `Operator^`).
             join_operator (JoinOperator^): The operator used to join the multiple filter
                 3-tuples.
         """
--- a/src/taipy/core/data/data_node.py
+++ b/src/taipy/core/data/data_node.py
@@ -320,12 +320,12 @@ class DataNode(_Entity, _Labeled):
         from ._data_manager_factory import _DataManagerFactory
 
         self._write(data)
-        self._track_edit(job_id=job_id, **kwargs)
+        self.track_edit(job_id=job_id, **kwargs)
         self.unlock_edit()
         _DataManagerFactory._build_manager()._set(self)
 
-    def _track_edit(self, **options):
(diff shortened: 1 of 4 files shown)


src/taipy/core/data/data_node.py
```diff
@@ -359,7 +359,7 @@ def filter(self, operators: Union[List, Tuple], join_operator=JoinOperator.AND):
         join operator (_AND_ or _OR_).
 
         Parameters:
-            operators (Union[List[Tuple], Tuple]): TODO
+            operators (Union[List[Tuple], Tuple]): 3-tuples or list of 3-tuples (key, value, `Operator^`).
```
trgiangdo:
```suggestion
            operators (Union[List[Tuple], Tuple]): A 3-element tuple or a list of 3-element tuples, each is in the form of (key, value, `Operator^`).
```
This is unrelated to the purpose of the PR. But I think this is clearer?

I'm not sure if I understand it correctly or not though, but we need to make it clearer for the user.

src/taipy/core/data/data_node.py
```diff
@@ -320,12 +320,12 @@ def write(self, data, job_id: Optional[JobId] = None, **kwargs: Dict[str, Any]):
...
         self._write(data)
-        self._track_edit(job_id=job_id, **kwargs)
+        self.track_edit(job_id=job_id, **kwargs)
         self.unlock_edit()
         _DataManagerFactory._build_manager()._set(self)
 
-    def _track_edit(self, **options):
-        """Add Edit tracking information to this data node."""
+    def track_edit(self, **options):
+        """Add edit tracking information to this data node."""
```
trgiangdo: I think we need to explain the `options` parameter as well.

trgiangdo (dismissed):
I only have a few comments about the docstring

src/taipy/core/data/data_node.py
toan-quach: yeah... trying to think of a way to explain it actually 🤣

jrobinAV: we need to say that Taipy interprets some options:
- timestamp
- comments
- job_id
- maybe some others

The others are user custom. Users can use options to attach any information to an external  edit of a dn.

src/taipy/core/data/data_node.py
```diff
@@ -320,12 +320,17 @@ def write(self, data, job_id: Optional[JobId] = None, **kwargs: Dict[str, Any]):
...
         self.unlock_edit()
         _DataManagerFactory._build_manager()._set(self)
 
-    def _track_edit(self, **options):
-        """Add Edit tracking information to this data node."""
+    def track_edit(self, **options):
+        """Add edit tracking information to this data node.
+
+        Parameters:
+            options: track `timestamp`, `comments`, `job_id`. The others are user-custom, users can use options
```
toan-quach, line 335: Added description for `options` parameter. Let me know what you think? @trgiangdo @jrobinAV

trgiangdo: I am really confused now.
Do we track track `timestamp`, `comments`, `job_id` by default, and the user can add more tracking options?
Or when the user provides new options, `timestamp`, `comments`, `job_id` will not be tracked anymore?

toan-quach: We track timestamp, comments, job_id, and user can add more information that suit their case. On second thought. should we add timestamp, comments, job_id to the signature to make it explicit? Something like: 
`def track_edit(self,  timestamp: datetime.datetime = None, comments: str = None, job_id: JobId = None, **options):`

What do you think? @jrobinAV @trgiangdo

trgiangdo: I think we should clear it in the doc only.
The default value of timestamp, comments, job_id are None may cause confusion.

toan-quach: hmm I don't think it will cause any confusions though :/

jrobinAV: First, we can improve the whole class documentation. not only the method documentation. We can add more context to this method, which is greatly linked to the `edits` attribute. We can explain what an edit is and how it is updated and used by Taipy. We can add an explanation for each of the three Taipy options. 

1. **timestamp** represents the time instant of the writing. It is used to sort the edits and trace a data node's change history. It is populated when the `write` method is called (by task execution or manually), when the data node is edited from the `data_node` viz elmt, or when the `track_edit` is called. It is displayed in the `data_node` visual element. 
2. **comments** represents a free text to explain or comment on a change of a data node. It can be populated by an end-user when the data node is edited from the `data_node` viz elmt, the `write` method is called manually. It is displayed in the `data_node` visual element.
3. **job_id** is only populated when the data node is written by a task execution and corresponds to the job's id. It is displayed in the `data_node` visual element.
4. We can give examples of custom options when the `write` and `track_edit` methods are manually called. The name of the third-party software. The id of the process that modified the data. The contact information of the IT responsible for the change, The user name of the end-user that edited the data, etc.

Once the `edits` attribute is clearly documented, we can say that the method creates and adds a new entry in the `edits` attribute without writing the data. We can refer to the edits attribute documentation and explain the motivation to open this API. It is particularly handy when data (referred by the data node) is modified independently from Taipy tasks and data nodes (a file is updated on the disk, a database is written by third-party software, etc.). 
It also lets Taipy know that some data node has been edited, so skippable downstream tasks should not be skipped.
We can also repeat the examples of custom options as well.

trgiangdo: First of all, that's more work for Toan 😂

Secondly, I'm not sure about the example, since this is in refman and I don't want it to be too long.
But a short and simple one is better than none I guess.

jrobinAV: The ref manual can be more detailed than what we used to do. For instance, the GUI team uses much more the ref manual than we do.
There is no need to add all the details for all Classes and functions. However, this feature is tricky. I believe it does make sense.

src/taipy/core/data/data_node.py
```diff
@@ -60,8 +60,12 @@ class DataNode(_Entity, _Labeled):
             None.
         parent_ids (Optional[Set[str]]): The set of identifiers of the parent tasks.
         last_edit_date (datetime): The date and time of the last modification.
-        edits (List[Edit^]): The list of Edits (an alias for dict) containing medata about each
-            edition of that node.
+        edits (List[Edit^]): The list of Edits (an alias for dict) containing metadata about each
+            edition of that node including but not limited to timestamp, comments, job_id:
```
jrobinAV:
```suggestion
            data edition including but not limited to timestamp, comments, job_id:
```

src/taipy/core/data/data_node.py
```diff
@@ -60,8 +60,12 @@ class DataNode(_Entity, _Labeled):
...
+            timestamp: The time instant of the writing
+            comments: Representation of a free text to explain or comment on a change of a data node
```
jrobinAV:
```suggestion
            comments: Representation of a free text to explain or comment on a data change
```

jrobinAV (dismissed):

jrobinAV (approved):

trgiangdo (approved):

PR #752, opened 2023-09-07
bug/ fix incomplete scenario config checker for sequence

toan-quach:
fixed scenario config checker for sequence missed checking taskconfig type

### DIFF
--- a/src/taipy/core/config/checkers/_scenario_config_checker.py
+++ b/src/taipy/core/config/checkers/_scenario_config_checker.py
@@ -119,8 +119,6 @@ class _ScenarioConfigChecker(_ConfigChecker):
         for task_config in scenario_config.tasks:
             if isinstance(task_config, TaskConfig):
                 scenario_task_ids.add(task_config.id)
-            else:
-                return
         for sequence_tasks in scenario_config.sequences.values():
             self._check_children(
                 ScenarioConfig,
--- a/src/taipy/core/config/checkers/_scenario_config_checker.py
+++ b/src/taipy/core/config/checkers/_scenario_config_checker.py
@@ -115,10 +115,23 @@ class _ScenarioConfigChecker(_ConfigChecker):
     def _check_tasks_in_sequences_exist_in_scenario_tasks(
         self, scenario_config_id: str, scenario_config: ScenarioConfig
     ):
-        scenario_tasks = scenario_config.tasks
-        for sequence in scenario_config.sequences.values():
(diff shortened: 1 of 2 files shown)


src/taipy/core/config/checkers/_scenario_config_checker.py
```diff
@@ -115,10 +115,23 @@ def _check_additional_dns_not_overlapping_tasks_dns(self, scenario_config_id: st
...
-        scenario_tasks = scenario_config.tasks
-        for sequence in scenario_config.sequences.values():
-            for task in sequence:
-                if task not in scenario_tasks:
+        scenario_task_ids = set()
+        for task_config in scenario_config.tasks:
+            if isinstance(task_config, TaskConfig):
+                scenario_task_ids.add(task_config.id)
+            else:
+                return
```
jrobinAV: Why do we return here?
- If it is not a TaskConfig, that's an error to collect, no?
- If the error is checked somewhere else, we should continue and not return. No?

toan-quach: That's actually for the unit tests only. Some of the tests we are catching the SystemExit error and keep on running the code. As a result, the rest of the code will be run and result in failure. In reality, the condition will not fall into the False case as if it does, it'd have been caught by `_check_task_configs` in ScenarioConfig checkers

tests/core/config/checkers/test_scenario_config_checker.py
```diff
@@ -293,6 +293,19 @@ def test_check_tasks_in_sequences_exist_in_scenario_tasks(self, caplog):
         assert len(Config._collector.warnings) == 0
         assert len(Config._collector.infos) == 0
 
+        config._sections[ScenarioConfig.name]["new"].add_sequences({"sequence_2": ["task_config"]})
```
trgiangdo, line 296:
```suggestion
        config._sections[ScenarioConfig.name]["new"].add_sequences({"sequence_2": [task_config_1, "task_config"]})
```
I propose testing a list consisting of both TaskConfig and non-TaskConfig

trgiangdo (dismissed):
Nice catch

tests/core/config/checkers/test_scenario_config_checker.py
toan-quach: haha sure why not

trgiangdo (approved):

jrobinAV (approved):

Issue #754, opened 2023-09-11
Update config schema


jrobinAV:
- add taipy version in Global Config section
- describe sequence values as an array of task config ids

PR #756, opened 2023-09-11
feature/#754 improve config schema

toan-quach:
#754

### DIFF
--- a/src/taipy/core/config/config.schema.json
+++ b/src/taipy/core/config/config.schema.json
@@ -8,7 +8,9 @@
     "CORE": {
       "description": "Data nodes",
       "type": "object",
-      "properties": {},
+      "properties": {
+        "version": "3.0.0"
+      },
       "required": [],
       "additionalProperties": {
         "properties": {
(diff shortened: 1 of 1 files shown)


src/taipy/core/config/config.schema.json
```diff
@@ -8,7 +8,9 @@
     "CORE": {
       "description": "Data nodes",
       "type": "object",
-      "properties": {},
+      "properties": {
+        "version": "3.0.0"
```
jrobinAV, line 12: should we have a description and a type instead of a value ?

toan-quach: @jrobinAV do you mean something like: 
```suggestion
        {"version": 
                {"description": "The version of Taipy Core",
                "type": "string"
                }
        }
```

jrobinAV: Yes, but @FredLL-Avaiga should know better.

jrobinAV:
Is this PR closed on purpose?

src/taipy/core/config/config.schema.json
FredLL-Avaiga: yes that would do.
What about the sequences description ?

toan-quach:
@jrobinAV @FredLL-Avaiga this PR has been closed and will be moved over to PR https://github.com/Avaiga/taipy-core/pull/762 so the topic is easier to maintain and follow

PR #762, opened 2023-09-13
feature/#289 added core version to core section

toan-quach:


src/taipy/core/config/config.schema.json
```diff
@@ -408,8 +412,8 @@
             }
           },
           "sequences": {
-            "description": "The dictionary of sequence descriptions.",
-            "type": "object",
+            "description": "The list of task configs.",
+            "type": "array",
```
FredLL-Avaiga: I think `sequences` is an object with properties that are string arrays ?

src/taipy/core/config/config.schema.json
```diff
@@ -12,6 +12,10 @@
       "required": [],
       "additionalProperties": {
         "properties": {
+          "core_version": {
```
FredLL-Avaiga: I thought it would simply be `version` ?

src/taipy/core/config/core_section.py
```diff
@@ -93,6 +99,11 @@ def __init__(
         self.mode = mode or self._DEFAULT_MODE
         self.version_number = version_number or self._DEFAULT_VERSION_NUMBER
         self.force = force or self._DEFAULT_FORCE
+
+        self._core_version = core_version or self._DEFAULT_LATEST_CORE_VERSION
+        if self._core_version != self._DEFAULT_LATEST_CORE_VERSION:
```
FredLL-Avaiga: how do you handle minor or patched versions ie 3.0.1 ... ?

FredLL-Avaiga: I suppose there's a package for that ....
ex: https://pypi.org/project/sem-ver/

FredLL-Avaiga (changes requested):

src/taipy/core/config/core_section.py
```diff
@@ -258,6 +273,10 @@ def _update(self, as_dict: Dict[str, Any]):
         if self.force != force:
             self.force = force
 
+        core_version = _tpl._replace_templates(as_dict.pop(self._CORE_VERSION_KEY, self._core_version))
+        if self._core_version != core_version:
+            self._core_version = core_version
+
```
jrobinAV: We should not make it updatable. The default value (`taipy.core.__version__`) should be populated when instantiating the Singleton and never change.
```suggestion
        core_version = _tpl._replace_templates(as_dict.pop(self._CORE_VERSION_KEY, None))
        self.check_compatibility(core_version)
```

src/taipy/core/config/core_section.py
```diff
@@ -71,6 +73,9 @@ class CoreSection(UniqueSection):
     _FORCE_KEY = "force"
     _DEFAULT_FORCE = False
 
+    _CORE_VERSION_KEY = "core_version"
+    _DEFAULT_LATEST_CORE_VERSION = "3.0.0"
```
jrobinAV: Can't we use `taipy.core.__version__` value?
```suggestion
    _CORE_VERSION = "3.0.0"
    
```

src/taipy/core/config/core_section.py
```diff
@@ -93,6 +99,11 @@ def __init__(
         self.mode = mode or self._DEFAULT_MODE
         self.version_number = version_number or self._DEFAULT_VERSION_NUMBER
         self.force = force or self._DEFAULT_FORCE
+
+        self._core_version = core_version or self._DEFAULT_LATEST_CORE_VERSION
+        if self._core_version != self._DEFAULT_LATEST_CORE_VERSION:
+            raise ConfigCoreVersionMismatched(self._core_version, self._DEFAULT_LATEST_CORE_VERSION)
```
jrobinAV: We don't want to check compatibility here but in the update method.
```suggestion
        self._core_version = core_version
```

src/taipy/core/config/core_section.py
```diff
@@ -93,6 +99,11 @@ def __init__(
...
+
         super().__init__(**properties)
```
jrobinAV, line 109:
```suggestion

    def _check_compatibility(core_version):
        version_pattern = r'^(\d+)\.(\d+)\.(\d+)$'

        installed_match = re.match(version_pattern, self._core_version)
        required_match = re.match(version_pattern, core_version)

        if required_match and installed_match:
            required_major, required_minor, required_patch = map(int, required_match.groups())
            installed_major, installed_minor, installed_patch = map(int, installed_match.groups())

            if (
                required_major == installed_major
                and required_minor == installed_minor
            ):
                return True
        return False

```

jrobinAV (changes requested):
1. the core version is missing:
- in the `default_config` class method
- in the `__copy__` method
- in the `_clean` method

2. The core version must NOT be changed in the `_update` method.
The default config value should never change since it corresponds to the taipy version currently used. Instead, we should check during the _update (called basically when loading a TOML file) that the file version is compatible with the current version.

3. We should not only compare the two versions values, but we should check the compatibility (basically the two versions should have the same Major.minor values)

4. We need to have dedicated class for testing the various use cases:
- loading a config file with current version value
- loading a config file with future incompatible version value
- loading a config file with future but compatible version value
- loading a config file with past incompatible version value
- loading a config file with past but compatible version value

src/taipy/core/config/config.schema.json
toan-quach: Not sure if I understand correctly, have made changes to it

src/taipy/core/config/config.schema.json
toan-quach: I wanted to at first, but in core_section, we also have version_number that's for the version management system so the name is confusing

src/taipy/core/config/core_section.py
toan-quach: this is still a topic I'm trying to improve with no good idea as of the moment

src/taipy/core/config/core_section.py
toan-quach: I'm trying to find a way to use it, not much luck so far due to circular import :/

src/taipy/core/config/core_section.py
toan-quach: I'd still prefer to keep `core_version or self._DEFAULT_LATEST_CORE_VERSION` as core_version can be None and it will trigger issues in check_compatibility

src/taipy/core/config/core_section.py
toan-quach: Have decided to use the `__version__` value in `src/taipy/core/__init__.py`

src/taipy/core/config/core_section.py
toan-quach: with some refactoring, it works now :D

src/taipy/core/config/core_section.py
toan-quach: it will also help keep a value for core_service when serialize

src/taipy/core/config/config.schema.json
FredLL-Avaiga: I think the right way of describing it is:
```
"sequences": {
            "description": "The dictionary of sequence descriptions.",
            "type": "object",
            "additionalProperties": {"type": "array", "items": {"type": "string", "uniqueItems": true}}
          }
```

so that we support a TOML like

```
[SCENARIO.mon_scenario.sequences]
   seq1= ["hh", "hh"]
   seq2= ...

```
or

```
[SCENARIO.mon_scenario]
...
sequences={
   seq1= ["hh", "hh"]
   seq2= ...
   }
```

toan-quach: I'm not an expert in this so I'll just follow what the expert tells me to do 😃

FredLL-Avaiga: I'm no expert either :-)
https://json-schema.org/understanding-json-schema/reference/object.html

src/taipy/core/config/config.schema.json
FredLL-Avaiga: In fact the whole CORE description is wrong:
additionalProperties.properties.* should be in properties

CORE is not like DATA_NODE: it doesn't have an intermediate level

FredLL-Avaiga (dismissed):
except for the whole CORE definition of properties (and description) the Schema looks good to me

tests/core/config/test_core_version.py
```diff
@@ -0,0 +1,67 @@
...
+            mode = "development"
+            version_number = ""
+            force = "False:bool"
+            core_version = "{core_version}"
+
+            [VERSION_MIGRATION.migration_fcts]
+            """
+        )
+        if is_compatible:
+            Config.override(file_config.filename)
```
jrobinAV, line 105:
```suggestion
            Config.load(file_config.filename)
```

tests/core/config/test_core_version.py
```diff
@@ -0,0 +1,67 @@
...
+        else:
+            with pytest.raises(ConfigCoreVersionMismatched):
+                Config.override(file_config.filename)
```
jrobinAV, line 109:
```suggestion
                Config.load(file_config.filename)
                
    @pytest.mark.parametrize(
        "core_version,is_compatible",
        [
            ("3.0.1", True),  # current version
            ("3.0.1.dev0", True),  # current dev version
            ("3.0.2", True),  # future but compatible
            ("3.0.2.dev0", True),  # dev future but compatible
            ("3.0.0", True),  # past but compatible
            ("3.0.0.dev0", True),  # dev past but compatible
            ("3.1.0", False),  # future but incompatible
            ("3.1.0.dev0", False),  # dev future but incompatible
            ("2.1.0", False),  # past but incompatible
            ("2.1.0.dev0", False),  # dev past but incompatible
        ],
    )
    def test_override_configuration_file(self, core_version, is_compatible):

        # breakpoint()
        file_config = NamedTemporaryFile(
            f"""
            [TAIPY]
            [JOB]
            mode = "standalone"
            max_nb_of_workers = "2:int"
            [CORE]
            root_folder = "./taipy/"
            storage_folder = ".data/"
            repository_type = "filesystem"
            read_entity_retry = "0:int"
            mode = "development"
            version_number = ""
            force = "False:bool"
            core_version = "{core_version}"
            [VERSION_MIGRATION.migration_fcts]
            """
        )
        if is_compatible:
            Config.override(file_config.filename)
        else:
            with pytest.raises(ConfigCoreVersionMismatched):
                Config.override(file_config.filename)
```

tests/core/config/test_core_version.py
```diff
@@ -0,0 +1,67 @@
...
+            ("3.0.2.dev0", True),  # dev future but compatible
+            ("3.0.0", True),  # past but compatible
+            ("3.0.0.dev0", True),  # dev past but compatible
+            ("3.1.0", False),  # future but incompatible
+            ("3.1.0.dev0", False),  # dev future but incompatible
+            ("2.1.0", False),  # past but incompatible
+            ("2.1.0.dev0", False),  # dev past but incompatible
+        ],
+    )
+    # def test_load_configuration_file(content, columns):
```
jrobinAV:
```suggestion
```

tests/core/config/test_core_version.py
```diff
@@ -0,0 +1,67 @@
...
+
+from src.taipy.core.config.core_section import CoreSection
+from src.taipy.core.exceptions import ConfigCoreVersionMismatched
+from taipy.config.config import Config
+from tests.core.utils.named_temporary_file import NamedTemporaryFile
+
+
+class TestCoreVersionInCoreSectionConfig:
+    CoreSection._CURRENT_CORE_VERSION = "3.0.1"
+
```
jrobinAV, line 22:
```suggestion

    def test_load_configuration_file_without_core_section(self, core_version, is_compatible):
        file_config = NamedTemporaryFile(
            f"""
            [TAIPY]
            [JOB]
            mode = "standalone"
            max_nb_of_workers = "2:int"
            [CORE]
            root_folder = "./taipy/"
            storage_folder = ".data/"
            repository_type = "filesystem"
            read_entity_retry = "0:int"
            mode = "development"
            version_number = ""
            force = "False:bool"
            [VERSION_MIGRATION.migration_fcts]
            """
        )
        with pytest.raises(ConfigCoreVersionMismatched):
            Config.load(file_config.filename)
                
```

jrobinAV (commented):
Actually I believe we are still missing a use case.

If a TOML config does not contain the core_version (if it is None) then it is incompatible.

I added as a suggestion a proposal for a new test to check it.

src/taipy/core/config/core_section.py
```diff
@@ -93,6 +101,10 @@ def __init__(
         self.mode = mode or self._DEFAULT_MODE
         self.version_number = version_number or self._DEFAULT_VERSION_NUMBER
         self.force = force or self._DEFAULT_FORCE
+
+        self._core_version = core_version or self._CURRENT_CORE_VERSION
+        self._check_compatibility(self._core_version)
```
jrobinAV: If there is no version in a TOML file, we want to keep it `None` and the check_compatibility method should raise an error. 
No? 
We don't want to use the current version and then check the compatibility.

toan-quach: then wouldn't that make all of current existing toml raises error? since we haven't introduce core_version up until now

src/taipy/core/config/config.schema.json
```diff
@@ -8,63 +8,64 @@
     "CORE": {
       "description": "Data nodes",
```
FredLL-Avaiga, line 9: change description ?

FredLL-Avaiga (approved):
schema looks good to me

jrobinAV (approved):

PR #765, opened 2023-09-25
editor_id and editor_expiration_date must be stored in the datamodel.

jrobinAV:
The editor_id and the expiration date must be stored in the repo.

Before writing a data node, a user can lock it providing an editor id. The UI can read the editor_id before to let a user write the data to guarantee he s the one that locked the data node.
To be able to retrieve the editor_id, it must be stored in the repo.


FredLL-Avaiga (approved):
a lot of impacted files
Tx

toan-quach (approved):
lgtm

Issue #668, opened 2023-07-04
Add new API to know if a Data Node is up to date


FlorianJacta:
### What problem does this feature solve?
The feature aims to provide a solution for determining whether a Data Node is up-to-date. This functionality would simplify the management and utilization of Data Nodes within a Taipy application.

When called, `<Data Node>.is_up_to_date` would return True if all preceding Data Nodes in the graph have not been modified since the last execution.
If any of the preceding Data Nodes have been written or modified, `<Data Node>.is_up_to_date` would return False.

The revised implementation would involve renaming the existing function, `<Data Node>.is_up_to_date`, to `<Data Node>.is_valid`.

jrobinAV:
For end users, it will be difficult to remember the difference between `is_up_to_date` `from is_valid`.
What about replacing `is_up_to_date` by `has_new_upstream_data` ?

toan-quach:
Requirement:

- [ ] get all previous preceding data nodes of the targeted dn (based on the scenario it belongs to)
- [ ] check if any of the preceding data node has its _last_edit_date larger than its children nodes
- [ ] check step 2 for all parent scenarios
- [ ] add unit tests

toan-quach:
@FlorianJacta @jrobinAV 
So to conclude:
We will have 2 functions:
- `is_valid`: as Florian raised, it will be used to check if the dn is still valid (expired or not)
- `is_up_to_date`: to check if preceding/ancestor DNS have been updated (`last_edit_date` was changed and it was after the selected dn). This check will happen on all preceding datanodes that share the same scenario with the selected dn

jrobinAV:
What I understand is that the is_up_to_date method should check both:
- All ancestors have not been written after the current DN.
- The current DN is valid regarding the validity period.

toan-quach:
@jrobinAV yep I think you're right!

PR #767, opened 2023-10-02
Feature/#668 dn is up to date

toan-quach:
#668

### DIFF
--- a/src/taipy/core/data/data_node.py
+++ b/src/taipy/core/data/data_node.py
@@ -570,9 +570,6 @@ class DataNode(_Entity, _Labeled):
 
     @property
     def is_up_to_date(self) -> bool:
-        # NOTE:
-        # What about prev_1.last_edit_date > prev_2.last_edit_date but < current/prev_3.last_edit_date
-
         from ..scenario.scenario import Scenario
         from ..taipy import get_parents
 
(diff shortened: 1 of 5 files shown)


src/taipy/core/data/data_node.py
```diff
@@ -567,6 +568,22 @@ def is_up_to_date(self) -> bool:
             return False
         return True
 
+    @property
+    def is_up_to_date(self) -> bool:
```
trgiangdo, line 572: Please add a docstring for the refman to explain this property

src/taipy/core/data/data_node.py
```diff
@@ -549,7 +550,7 @@ def is_ready_for_reading(self) -> bool:
 
     @property  # type: ignore
     @_self_reload(_MANAGER_NAME)
-    def is_up_to_date(self) -> bool:
+    def is_valid(self) -> bool:
         """Indicate if this data node is up-to-date.
```
trgiangdo: The docstring here can be misleading since we have a new `is_up_to_date` property

trgiangdo (commented):
I have a few comments about the refman

trgiangdo (approved):
Both the document and the code looks good now. Nice work

jrobinAV (approved):
Neat!

Issue #336, opened 2022-10-07
Extend DataNode filter method to other formats


jrobinAV:
### Description

The filter method on the data node class does filter the data only when the expose type is dataframe. This should be extended for every possible data node

trgiangdo:
The technical requirement documentation (WIP): https://docs.google.com/document/d/1tQyMz0XHzjcwG88o3eNA1Pop1EsE-jL1c42jsapUSyU

PR #768, opened 2023-10-02
Feature/#336 - Filter numpy array exposed type

trgiangdo:
https://github.com/Avaiga/taipy-core/issues/336

A few things we need to consider:
- We should probably also consider to expose the Operator and JoinOperator at the taipy init level.
- If the exposed type is numpy, then the key is the index of the column. If the user wants to filter by row value, we need to find a way to allow that.


toan-quach (dismissed):
lgtm

src/taipy/core/data/data_node.py
```diff
@@ -436,11 +437,16 @@ def filter(self, operators: Union[List, Tuple], join_operator=JoinOperator.AND):
...
+                return DataNode.__filter_numpy_array(data, list_operators)
             if isinstance(data, List):
                 return DataNode.__filter_list_per_key_value(data, operators[0], operators[1], operators[2])
         else:
             if isinstance(data, (pd.DataFrame, modin_pd.DataFrame)):
                 return DataNode.__filter_dataframe(data, operators, join_operator=join_operator)
+            if isinstance(data, np.ndarray):
+                return DataNode.__filter_numpy_array(data, operators, join_operator=join_operator)
             if isinstance(data, List):
                 return DataNode.__filter_list(data, operators, join_operator=join_operator)
```
jrobinAV, line 451: The `datanode.filter` method could delegate the filtering logic to a dedicated class. What do you think? is n't it the purpose of the `_FilterDataNode` class?

jrobinAV (commented):
1. excel.py should also be updated with the return type of method `_read_as_numpy`.

2. I would like to challenge the design:
    - I am not sure to understand the purpose of the `_FilterDataNode` class
    - I would like to move the whole filtering logic out of the `DataNode` class. The `datanode.filter` method could delegate the filtering logic to a dedicated class like `Filter`. What do you think?

trgiangdo:
1. That method of excel can also return a dict of array, I should update it as well.

2.1. Me neither 🥲 I think we can remove it since it is not documented at all.

2.2. I agree with moving the logic to a dedicate file. And if I understand correctly, the filter() method should still return the filtered data directly, not returning the Filter class right?

trgiangdo:
In this PR:

New feature:
- Ability to filter numpy exposed_type

Refactoring:
- Refactor the _FilterDataNode class. Move the logic of the DataNode.filter() method to the _FilterDataNode class.
- Simplify the __getitem__ method of DataNode, make it consistent with DataNode.filter() method.
- Add unit tests for filtering all data node types.

Bug fixes:
- Fix unable to to filter excel data node (because excel data node return a dictionary of exposed_type, not the direct exposed_type).
- Fix unable to filter JSONDataNode when the data is a list of dictionaries.
- Fix unable to filter a list of objects (custom exposed type) when some object has an attribute that others don't.
- Fix unable to write to SQLDataNode due to not converting the string query into SQL text query (raised by issue https://github.com/Avaiga/taipy-core/issues/816).

src/taipy/core/data/_filter.py
```diff
@@ -10,173 +10,250 @@
...
+            has_df_children = all([isinstance(e, (pd.DataFrame, modin_pd.DataFrame)) for e in data.values()])
+            has_list_children = all([isinstance(e, List) for e in data.values()])
+            has_np_array_children = all([isinstance(e, np.ndarray) for e in data.values()])
+            return has_df_children or has_list_children or has_np_array_children
         return False
 
-    def data_is_dataframe(self) -> bool:
-        return self.data_type == self.__DATAFRAME_DATA_TYPE
+    @staticmethod
+    def __is_list_of_dict(self) -> bool:
```
jrobinAV: what is self for a static method?

jrobinAV (dismissed):

src/taipy/core/data/_filter.py
trgiangdo: My bad, the `self` should be removed

jrobinAV (approved):

PR #769, opened 2023-10-03
Feature/test benchmark

joaoandre-avaiga:


tests/__init__.py
```diff
@@ -9,4 +9,4 @@
 # an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the
 # specific language governing permissions and limitations under the License.
 
-"""Unit test package for taipy."""
+"""Unit test package for taipy.."""
```
jrobinAV:
```suggestion
"""Unit test package for taipy."""
```

.github/workflows/trigger-benchmark.yml
```diff
@@ -1,7 +1,7 @@
 name: Benchmark on taipy-integration-testing
 on:
   push:
```
jrobinAV, line 3:
```suggestion
name: Benchmark on taipy-benchmark
on:
  push:
```

.github/workflows/trigger-benchmark.yml
jrobinAV, line 1: Need to be updated regarding the new repository taipy-benchmark

jrobinAV:
Do we want to merge it before the repository merge? Otherwise, I believe we can just close it and re do it in taipy repository directly once the repos are merged.
@joaoandre-avaiga What do you think?

Issue #719, opened 2023-08-07
Remove migration paths from taipy 1.X to 2.0


trgiangdo:
### Description

Currently, there are a few migration paths for application from taipy 1.X to work on taipy 2.X.
- `_version_migration()`
- `_to_edits_migration()`
- `cacheable` to `skippable` of a Task
- `_migrate_subscriber()`

We need to discuss if we can remove those in taipy 3.0.

PR #771, opened 2023-10-05
Feature/#719 - Remove migration methods at model level

trgiangdo:
https://github.com/Avaiga/taipy-core/issues/719

This PR can only be merged when the migration CLI is ready to cover all the migrations.


jrobinAV (commented):
Sounds good.

Issue #776, opened 2023-10-09
Add unit tests for migrate CLI


jrobinAV:
https://github.com/Avaiga/taipy-core/pull/773

PR #786, opened 2023-10-12
Test/#776 - Add unit tests for migrate CLI

trgiangdo:
https://github.com/Avaiga/taipy-core/issues/776

In this PR, alongside the tests, some issues have been fixed:
- The pipeline entities are not removed after migration finished.
- The path for FS or SQL is not checked if existed or not before migrating.
- Task entities are not migrated at all.
- Remove unnecessary default values on some methods in the migration script.
- Improve the CLI info and error messages.

TODO:
- [x] Improve the coverage for migrating entities on FS repo.
- [x] Add tests for migrating entities on SQL repo.
- [x] Add tests for migrating entities on Mongo repo.
- [x] Fix any remaining issue.


src/taipy/core/_entity/_migrate_cli.py
```diff
@@ -8,26 +8,29 @@
...
             "migrate",
-            help="Migrate entities from old taipy versions to current taipy version. The entity migration "
-            "should be performed only after updating taipy code to the current version.",
+            help="Migrate entities created from old taipy versions to be compatible with the current taipy version. "
+            " The entity migration should be performed only after updating taipy code to the current version.",
         )
         migrate_parser.add_argument(
             "--repository-type",
             nargs="+",
-            choices=["filesystem", "sql", "mongo"],
```
jrobinAV, line 30: how come we don't specify choices anymore?

src/taipy/core/_entity/_migrate.py
```diff
@@ -274,13 +275,21 @@ def _load_all_entities_from_fs(root: str = ".data") -> Dict:
     return entities
 
 
-def __write_entities_to_fs(_entities: Dict, root: str = ".data"):
+def __write_entities_to_fs(_entities: Dict, root: str):
```
jrobinAV, line 289: We should have as a default value the default value of the configuration.
What do you think?

src/taipy/core/_entity/_migrate.py
```diff
@@ -295,10 +304,10 @@ def __connect_mongodb(db_host: str, db_port: int, db_username: str, db_password:
...
 
 def _load_all_entities_from_mongo(
-    hostname: str = "localhost",
-    port: int = 27017,
-    user: str = "taipy",
-    password: str = "password",
+    hostname: str,
+    port: int,
+    user: str,
+    password: str,
```
jrobinAV, line 321: We should have as a default value the default value of the configuration.
What do you think?

src/taipy/core/_entity/_migrate.py
```diff
@@ -322,16 +331,15 @@ def _load_all_entities_from_mongo(
...
 def __write_entities_to_mongo(
     _entities: Dict,
-    hostname: str = "localhost",
-    port: int = 27017,
-    user: str = "taipy",
-    password: str = "password",
+    hostname: str,
+    port: int,
+    user: str,
+    password: str,
```
jrobinAV, line 348: We should have as a default value the default value of the configuration.
What do you think?

src/taipy/core/_entity/_migrate.py
```diff
@@ -344,7 +352,7 @@ def __write_entities_to_mongo(
         )
 
 
-def _load_all_entities_from_sql(db_file: str = "test.db") -> Tuple[Dict, Dict]:
+def _load_all_entities_from_sql(db_file: str) -> Tuple[Dict, Dict]:
```
jrobinAV, line 366: We should have as a default value the default value of the configuration.
What do you think?

src/taipy/core/_entity/_migrate.py
```diff
@@ -443,7 +451,7 @@ def __insert_version(version: dict, conn):
     conn.commit()
 
 
-def __write_entities_to_sql(_entities: Dict, _versions: Dict, db_file: str = "test.db"):
+def __write_entities_to_sql(_entities: Dict, _versions: Dict, db_file: str):
```
jrobinAV, line 465: We should have as a default value the default value of the configuration.
What do you think?

src/taipy/core/_entity/_migrate.py
```diff
@@ -492,19 +500,19 @@ def __migrate(entities: Dict, versions: Optional[Dict] = None) -> Tuple[Dict, Op
     return entities, versions
 
 
-def _migrate_fs_entities(path: str) -> None:
```
jrobinAV, line 495: Why removing the returned type?

src/taipy/core/_entity/_migrate_cli.py
trgiangdo: Because this is a `nargs="+"` option, meaning that the user can also provide path, host, port, etc.

If we keep the `choices`, the path, host, port also need to be chosen from that `choices` list.

src/taipy/core/_entity/_migrate.py
trgiangdo: The default value should be processed at the CLI level.

I don't think we need/should have the default here.

src/taipy/core/_entity/_migrate.py
trgiangdo: The default value should be provided on the `_migrate_mongo_entities()` method. I removed these from all other methods to prevent any conflict in values in the future.

src/taipy/core/_entity/_migrate.py
trgiangdo: The `db_file` should always be provided by the user and should be processed at the CLI level.

src/taipy/core/_entity/_migrate.py
trgiangdo: It's just not necessary so I removed it to make it cleaner.

src/taipy/core/_entity/_migrate.py
```diff
@@ -23,9 +24,6 @@
 
 
 def __update_parent_ids(entity: Dict, data: Dict) -> Dict:
-    # parent_id was replaced by parent_ids
-    _ = entity.pop("parent_id", None)
```
joaoandre-avaiga, line 27: Not sure I understand why we are removing this line.

src/taipy/core/_entity/_migrate.py
```diff
@@ -113,18 +107,27 @@ def _migrate_scenario(scenario: Dict, data: Dict) -> Dict:
...
             return False
     return True
 
 
 def _migrate_task(task: Dict, data: Dict) -> Dict:
-    # owner_id was not present in 2.0
-    if task.get("parent_id"):
-        task["owner_id"] = task.get("owner_id")
+    # parent_id has been renamed to owner_id
+    try:
```
joaoandre-avaiga, line 126: I would maintain the original structure in this case, just fixing the worng key on line 126. Catching a exception just to do nothing adds unnecessary complexity. But is just preference, feel free to keep it as you like.

src/taipy/core/_entity/_migrate.py
joaoandre-avaiga: Actually it should always have a return type hint. Either `-> None` or `-> NoReturn `. Mypy asks to always keep the return type explicit. Since it didn't break on pre-commit and the GH action, maybe we are ignoring this specific mypy error, in that case we can decide what should be the standard (I prefer to always have a type hint if possible)

src/taipy/core/_entity/_migrate.py
trgiangdo: Because we will pop it when we handle the `parent_id` -> `owner_id` migration.

`parent_id` and `parent_ids` are actually not related so I don't prefer doing it here.

src/taipy/core/_entity/_migrate.py
trgiangdo: I changed it because in case `"parent_id': null`, the `if task.get("parent_id")` also is None, so the `"owner_id"` will not exist in the entity, which is not what we want.

joaoandre-avaiga (dismissed):

src/taipy/core/_entity/_migrate.py
jrobinAV: I do prefer too. It is easier to understand what does a function/method when just reading the signature. It's a way to document it.

jrobinAV (dismissed):

jrobinAV (approved):

Issue #787, opened 2023-10-13
BUG-Can not change the db_location of SQL repository


trgiangdo:
**Description**
The SQL engine is always called when init taipy, so that whatever the db_location is set by the `Config.configure_core()` method, the db_location will always be "None".

The problem is caused by:

In `_sql_repository.py`, there is an import
```python
from .db._init_db import init_db
```
which called to the `db._init_db.py` file, which has an import
```python
from ._sql_session import engine
```
which the `engine` is a global variable and will always be generated at the import.

**How to reproduce**

```python
import taipy as tp
from taipy import Config

Config.configure_core(repository_type="sql", repository_properties={"db_location": "test.sqlite"})
tp.Core().run()
```

After running the code (with some configuration if possible), there will be a "None" file which is a SQLite database.

**Expected behavior**
"db_location" should be properly assigned as the path of the SQLite database.

PR #792, opened 2023-10-14
Fix/#787 - Cannot change db_location of sql repository type

trgiangdo:
https://github.com/Avaiga/taipy-core/issues/787


tests/conftest.py
```diff
@@ -439,28 +436,29 @@ def sql_engine():
     return create_engine("sqlite:///:memory:")
 
 
-@pytest.fixture()
+@pytest.fixture
 def init_sql_repo(tmp_sqlite):
```
trgiangdo, line 440: Now the `init_sql_repo` fixture will also clean the SQLite database based on the `tmp_sqlite` path

tests/core/job/test_job_manager_with_sql_repo.py
```diff
@@ -51,6 +52,12 @@ def init_managers():
     _JobManagerFactory._build_manager()._delete_all()
 
 
+def clear_sql_session():
+    _build_engine.cache_clear()
+    _SQLSession._SessionLocal = None
+    _SQLSession._engine = None
```
trgiangdo, line 58: This method is used at standalone tests to clear the cache created by the previous test, so that the subprocess can connect properly to the database.

src/taipy/core/_repository/db/_sql_session.py
```diff
@@ -22,22 +23,48 @@
...
-        # More sql databases can be easily added in the future
-        engine = create_engine(
-            f"sqlite:///{properties.get('db_location')}?check_same_thread=False",
-            poolclass=StaticPool,
-            json_serializer=dumps,
-            json_deserializer=loads,
-        )
-        return engine
-
+        db_location = properties["db_location"]
```
jrobinAV, line 59: Are we sure db_location does exist in properties?

tests/core/job/test_job_repositories.py
jrobinAV, line 1: I don't get why you replaced DataFSRepo by DataSQLRepo and TaskFSRepo by TaskSQLRepo? Don't you want to test both repo type instead?

src/taipy/core/_repository/db/_sql_session.py
joaoandre-avaiga: If is not present, a MissingRequiredProperty will be raise.

tests/core/job/test_job_repositories.py
joaoandre-avaiga: I guess in this case it doesn't really matter? Data and Task objects are just to build the job that will be saved. I think we don't check for consistency, for instance:

```
@pytest.mark.parametrize("repo", [_JobFSRepository, _JobSQLRepository])
def test_save_and_load(self, data_node, job, repo, init_sql_repo):
    _DataSQLRepository()._save(data_node)
    task = Task("task_config_id", {}, print, [data_node], [data_node])
    _TaskSQLRepository()._save(task)
    job._task = task

    repository = repo()
    repository._save(job)

    obj = repository._load(job.id)
    assert isinstance(obj, Job)
 ```
 
 So first test run:
 - repo is _JobFSRepository
 - Save Datanode and Task to _DataSQLRepository and _TaskSQLRepository
 - save job to _JobFSRepository
 
 In this scenario, both task and datanode exists only on SQL, and we are testing _JobFSRepo. Maybe we don't need 
 `_DataSQLRepository()._save(data_node)` and ` _TaskSQLRepository()._save(task)` for the test to be run?
 
 If we do, is kind of a weird interaction.

joaoandre-avaiga:
Small comments about the tests, otherwise looks good 👍

tests/core/job/test_job_repositories.py
trgiangdo: Thank you Joao.

Yes, the main point is to test different types of Job repositories.

We are saving the entities to SQL repo in this case because we are using the `init_sql_repo` fixture. So when loading a Job (whatever the Job Repo is), we also load the corresponding tasks and datanodes from SQL.

The flow is something like this:
So first test run:

- Repo is _JobFSRepository
- Save Datanode and Task to _DataSQLRepository and _TaskSQLRepository
- Save job to _JobFSRepository
- Load job from FS -> Load tasks and datanodes from SQL.

Again, where we store the tasks and datanodes doesn't matter, we just want to test the behavior of the Job repository.

toan-quach (approved):

jrobinAV (approved):

PR #796, opened 2023-10-17
feature/enterprise/#242 align pymongo version for core and enterprise

toan-quach:
https://app.zenhub.com/workspaces/taipy-613ef328affde6000ff77fb3/issues/gh/avaiga/taipy-enterprise/242

Since additional "srv" option for pymongo can be installed really quickly, I decided to keep it and not make it optional and complicating the solution

### DIFF
--- a/src/taipy/core/common/_mongo_connector.py
+++ b/src/taipy/core/common/_mongo_connector.py
@@ -19,8 +19,8 @@ def _connect_mongodb(
     db_host: str, db_port: int, db_username: str, db_password: str, db_extra_args: frozenset, db_driver: str
 ) -> pymongo.MongoClient:
     """Create a connection to a Mongo database.
-    The `"mongodb_extra_args"` passed by the user is originally a dictionary, but since `@lru_cache` wrapper only accepts
-    hashable parameters, the `"mongodb_extra_args"` should be converted into a frozenset beforehand.
+    The `"mongodb_extra_args"` passed by the user is originally a dictionary, but since `@lru_cache` wrapper only
+    accepts hashable parameters, the `"mongodb_extra_args"` should be converted into a frozenset beforehand.
 
     Parameters:
         db_host (str): the database host.
--- a/src/taipy/core/common/_mongo_connector.py
+++ b/src/taipy/core/common/_mongo_connector.py
@@ -9,29 +9,26 @@
 # an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the
 # specific language governing permissions and limitations under the License.
 
-from typing import Dict, Optional
+from functools import lru_cache
 
 import pymongo
 
 
+@lru_cache
 def _connect_mongodb(
-    db_host: str,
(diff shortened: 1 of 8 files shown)


trgiangdo (commented):
We need to update the connection method as well, to accept the server parameter.

src/taipy/core/common/_mongo_connector.py
```diff
@@ -0,0 +1,54 @@
...
+# Unless required by applicable law or agreed to in writing, software distributed under the License is distributed on
+# an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the
+# specific language governing permissions and limitations under the License.
+
+from typing import Dict, Optional
+
+import pymongo
+
+
+def _connect_mongodb(
```
toan-quach, line 18: not sure if this makes sense to put under a lru cache function as it will require extra step from user, tried to wrap it within another function as well but tests are failing. I'll have a look at it but not sure if it will be on time for the release. The effect is (probably) minimal as of now if not many users are using mongoDN so might not be a big issue after all

trgiangdo, line 18: Oh we did not have the `@lru_cache` here? Can you copy the method from taipy-enterprise?

trgiangdo: I believe we can copy the `_connect_mongo()` method from taipy-enterprise and it should work with @lru_cache.
However, that would require to update a line of code to convert the dictionary to frozen set when connect to the db in the MongoCollectionDN

toan-quach: Well I tried but it failed for unknown reason :/ and so I added lru on enterprise while keeping the code the same on core. Still investigating the issue

toan-quach: As mentioned above, I tried to add lru and have another function as wrapper, but some tests passed while some failed and havent found the reason for that yet. Still investigating the issue but it also might not be as urgent since we didnt have it before and Im more concerned if it's blocking the release :D

jrobinAV: I share the same concern.

The purpose of the ticket is first to align the Pymongo versions. Using lru_cache is a bonus. If we can make it on time, that is great, but I would not postpone the release for that.

Can't we split the two topics into two PRs so that first we align the version and then we work on the lru_cache?

My two cents,

jrobinAV (dismissed):

src/taipy/core/common/_mongo_connector.py
toan-quach: I have created a ticket for the lru cache topic. Otherwise, I think the ticket is ready to be merged.

https://app.zenhub.com/workspaces/taipy-613ef328affde6000ff77fb3/issues/gh/avaiga/taipy-core/797

toan-quach: Since Giang has already fixed the issue, I have closed the ticket

jrobinAV (approved):

PR #807, opened 2023-10-26
Accept to set a data node name using the property dict

jrobinAV:
Accept to set a data node name using the property dict instead of the actual attribute.


FabienLelaquais (commented):
😶

FabienLelaquais (approved):
👌

PR #810, opened 2023-10-30
Fix - Failed tests due to changed version

trgiangdo:
In this PR, tests on `tests/core/config/test_core_version.py` are updated as follow:
- The `CoreSection._CURRENT_CORE_VERSION` should not be run at the global level because it will affect all other tests.
- The `mock_core_version()` fixture is added for all tests in this file to mock the core versions on the `Config` singleton and update the `CoreSection._CURRENT_CORE_VERSION`.
- The `reset_core_version()` fixture is added to be run at the end of the session to reset the `CoreSection._CURRENT_CORE_VERSION` to its original value.


jrobinAV (approved):

toan-quach (approved):
lgtm!

PR #811, opened 2023-10-30
Refactor - SQLRepository uses sqlite3 instead of sqlalchemy

toan-quach:


src/taipy/core/_repository/_sql_repository.py
```diff
@@ -35,8 +36,7 @@ def __init__(self, model_type: Type[ModelType], converter: Type[Converter], sess
             converter: A class that handles conversion to and from a database backend
             db: An SQLAlchemy session object
```
trgiangdo:
```suggestion
            db: An sqlite3 connection object
```

src/taipy/core/_repository/_sql_repository.py
```diff
@@ -45,56 +45,79 @@ def __init__(self, model_type: Type[ModelType], converter: Type[Converter], sess
...
-            raise ModelNotFound(str(self.model_type.__name__), entity_id)
+        delete_query = self.model_type.__table__.delete().filter_by(id=entity_id)  # type: ignore
+        cursor = self.db.execute(str(delete_query.compile(dialect=sqlite.dialect())), [entity_id])
         self.db.commit()
 
+        if cursor.rowcount == 0:
+            raise ModelNotFound(str(self.model_type.__name__), entity_id)
+
+        if cursor.rowcount == 0:
+            raise ModelNotFound(str(self.model_type.__name__), entity_id)
```
trgiangdo: The `if` block is duplicated I believe. Also, should we run the `if` before commit the connection?

```suggestion
        if cursor.rowcount == 0:
            raise ModelNotFound(str(self.model_type.__name__), entity_id)
            
        self.db.commit()
```

trgiangdo (commented):
I have some small comments.

Other than that, there are a lot of `# type: ignore` due to the `ModelType` in _sql_repository. We will need to come back to the typing in the future.

trgiangdo (dismissed):
A lot of `# type: ignore` have been removed. Neat :D

florian-vuillemot:
Can "sqlachemy" be removed from setup.py and Pipfile?

toan-quach:
@florian-vuillemot sadly no since we still need it for the SQL Datanode

src/taipy/core/_repository/_encoder.py
```diff
@@ -27,7 +25,7 @@ def _timedelta_to_str(self, obj: timedelta) -> str:
             f"{int(total_seconds % 60)}s"
         )
 
-    def default(self, o: Any) -> Json:
```
jrobinAV, line 30: Why do we need to remove the typing?

src/taipy/core/job/_job_model.py
```diff
@@ -66,8 +58,8 @@ def from_dict(data: Dict[str, Any]):
             submit_id=data["submit_id"],
             submit_entity_id=data["submit_entity_id"],
             creation_date=data["creation_date"],
-            subscribers=subscribers,
-            stacktrace=stacktrace,
+            subscribers=_BaseModel._deserialize_attribute(data["subscribers"]),
+            stacktrace=data["stacktrace"],
```
jrobinAV: don't we want to use the deserialize method?

src/taipy/core/_repository/_base_taipy_model.py
```diff
@@ -26,3 +30,13 @@ def to_dict(self) -> Dict[str, Any]:
...
+
+    @staticmethod
+    def _serialize_attribute(value):
+        return json.dumps(value, ensure_ascii=False, cls=_Encoder)
+
+    @staticmethod
+    def _deserialize_attribute(value):
+        if isinstance(value, str):
+            return json.loads(value.replace("'", '"'), cls=_Decoder)
+        return value
```
jrobinAV, line 46: Much better with serialize and deserialize methods.

Thanks!!

src/taipy/core/_version/_version_model.py
```diff
@@ -37,8 +37,22 @@ class _VersionModel(_BaseModel):
...
+        return model
+
+    def to_list(self):
+        return [
+            self.id,
+            self.config,
+            self.creation_date,
+            self.is_production,
+            self.is_development,
+            self.is_latest,
```
jrobinAV, line 57: If I am not mistaken, the order of this list must be consistent with the order of the columns.  Am I correct?
 Do we have a way to test our code to ensure that?

jrobinAV:
> I have some small comments.
> 
> Other than that, there are a lot of `# type: ignore` due to the `ModelType` in _sql_repository. We will need to come back to the typing in the future.

Why not now? Do you think it will block the PR for too long? or do you believe it will make the PR too messy?

src/taipy/core/job/_job_model.py
jrobinAV: If yes, that means we are missing a unit test abt the stacktrace attribute.

src/taipy/core/_version/_version_model.py
toan-quach: ah yes, the order of list must be consistent with the column orders we defined in __table__

src/taipy/core/_repository/_encoder.py
toan-quach: it's due to a circular import issue

src/taipy/core/job/_job_model.py
toan-quach: I think you're right, it raised an error when I attempt to use deserialize here, will have a look

src/taipy/core/_repository/_encoder.py
jrobinAV: ok thx

src/taipy/core/job/_job_model.py
toan-quach: yep it worked now, I think there was some issue with multithreading but got fixed. Though will need unit tests

trgiangdo (approved):

jrobinAV (approved):

joaoandre-avaiga (approved):

Issue #806, opened 2023-10-26
Move name from data_node attribute and replace it by property


jrobinAV:
Store the name property of Datanodes into the property and not as an attribute.

PR #812, opened 2023-10-30
feature/#806 move name attribute of dn to properties

toan-quach:
#806


tests/core/notification/test_notifier.py
```diff
@@ -506,7 +506,7 @@ def test_publish_event():
         t.properties.update({"debug": True})
         assert registration_queue.qsize() == 0
 
-        d.name = "another new datanode name"
+        d.editor_id = "another new editor id"
```
jrobinAV, line 509: Why not test the name here? name is still exposed as a property, so it should still work.

tests/core/data/test_data_repositories.py
```diff
@@ -52,9 +52,9 @@ def test_load_all_with_filters(self, data_node, repo, init_sql_repo):
 
         for i in range(10):
             data_node.id = DataNodeId(f"data_node-{i}")
-            data_node.name = f"data_node-{i}"
```
jrobinAV, line 55: Why do we stop using `name` here? It should still work, shouldn't it?

toan-quach: not exactly, because it's not an attribute of dn anymore, and so when we try to find it with filtering in `_load_all` and `_search` for sql_repo, it returns an error as it's not a column in dn table.

tests/core/notification/test_notifier.py
toan-quach: the changes attribute will be under `properties` and not `name`, but we have already tested changes to `properties` so it's not needed

tests/core/data/test_data_repositories.py
jrobinAV: Oh... ok got it.

trgiangdo (dismissed):
LGTM

trgiangdo (approved):

jrobinAV (approved):

PR #817, opened 2023-11-06
Tests - Update filter data node tests on exposed types

trgiangdo:
In this PR, tests on data nodes with exposed_type are updated with more specific assertions.


jrobinAV (approved):
Neat!

Issue #815, opened 2023-10-31
Filter database datanode should not read all data


trgiangdo:
**What would that feature address**

The current steps to filter a data node:
1. Read **all** data from the data node
2. Apply the conditions and return the filtered data.

For a data node that uses an indexed database (SQL, Mongo), “read all data” will not be able to take advantage of the queries when reading data.

***Description of the ideal solution***
For indexed database data nodes, the conditions should be integrated into the read query, taking advantage of the query speed of the database.

PR #820, opened 2023-11-07
Feature/#815- Filter database datanodes should not read all data

trgiangdo:
https://github.com/Avaiga/taipy-core/issues/815


src/taipy/core/data/data_node.py
```diff
@@ -441,6 +440,10 @@ def __getitem__(self, item):
     def _read(self):
         raise NotImplementedError
 
+    @abstractmethod
+    def _read_by(self, operators: Union[List, Tuple], join_operator=JoinOperator.AND):
```
jrobinAV: Why don't you override the `filter` method directly instead of going through a `_read_by` method that is eventually not implemented?
It is possible to override the `filter` method in MongoCollectionDataNode and not in CsvDataNode, for instance.

trgiangdo: The reason is that I completely forgot 🤣

jrobinAV (approved):

PR #822, opened 2023-11-08
feature/#764 added submission entity, manager and repo

toan-quach:


src/taipy/core/_orchestrator/_orchestrator.py
```diff
@@ -66,31 +67,37 @@ def submit(
...
                 for task in ts:
                     res.append(
                         cls._submit_task(
-                            task, submit_id, submittable.id, callbacks=callbacks, force=force  # type: ignore
+                            task, submit_id, submission.entity_id, callbacks=callbacks, force=force  # type: ignore
                         )
                     )
 
+        submission.jobs = res
+        submission.update_submission_status()
```
jrobinAV: I believe we can discuss when the submission status is updated. 
I was initially thinking of updating the status when each job status is updated.

src/taipy/core/submission/submission.py
```diff
@@ -0,0 +1,146 @@
...
+        output (Union[DataNode^, List[DataNode^]]): The list of outputs.
+        id (str): The unique identifier of the task.
+        owner_id (str):  The identifier of the owner (sequence_id, scenario_id, cycle_id) or None.
+        parent_ids (Optional[Set[str]]): The set of identifiers of the parent sequences.
+        version (str): The string indicates the application version of the task to instantiate. If not provided, the
+            latest version is used.
+        skippable (bool): If True, indicates that the task can be skipped if no change has been made on inputs. The
+            default value is _False_.
+
+    """
```
jrobinAV: Busted!

src/taipy/core/submission/submission.py
```diff
@@ -0,0 +1,146 @@
...
+
+    _ID_PREFIX = "SUBMISSION"
+    _MANAGER_NAME = "submission"
+    __SEPARATOR = "_"
+
+    def __init__(
+        self,
+        entity_id: str,
+        jobs: Optional[Union[List[JobId], List[Job]]] = None,
+        submission_id: Optional[str] = None,
```
jrobinAV: We need to add an alias for the submission id, like for other entities.
```suggestion
        submission_id: Optional[SubmissionId] = None,
```

src/taipy/core/submission/submission.py
```diff
@@ -0,0 +1,146 @@
...
+    ):
```
jrobinAV, line 51: We are missing the submission_status and the creation date in the signature.

src/taipy/core/submission/submission.py
```diff
@@ -0,0 +1,146 @@
...
+    @_self_reload(_MANAGER_NAME)
+    def submission_status(self):
+        return self._submission_status
+
+    @submission_status.setter  # type: ignore
+    @_self_setter(_MANAGER_NAME)
+    def submission_status(self, submission_status):
+        self._submission_status = submission_status
+
+    def update_submission_status(self):
```
jrobinAV: If possible, I would call this method from the JobDispatcher in method _update_job_status.
If so, I believe all other usages are unnecessary anymore.

src/taipy/core/submission/_submission_converter.py
```diff
@@ -0,0 +1,41 @@
...
+
+    @classmethod
+    def _model_to_entity(cls, model: _SubmissionModel) -> Submission:
+        submission = Submission(
+            entity_id=model.entity_id,
+            jobs=model.job_ids,
+            submission_id=SubmissionId(model.id),
+        )
+        submission._creation_date = datetime.fromisoformat(model.creation_date)
+        submission._submission_status = model.submission_status
```
jrobinAV: Why don't you pass these in the constructor?

src/taipy/core/submission/_submission_manager.py
jrobinAV, line 1: To be completed with missing methods: etters, deleters, filters, etc...

src/taipy/core/submission/submission_status.py
```diff
@@ -0,0 +1,51 @@
...
+    - `CANCELED`: A `CANCELED` job has been submitted but its execution has been canceled.
+
+    - `FAILED`: A `FAILED` job raised an exception during its execution.
+
+    - `COMPLETED`: A `COMPLETED` job has successfully been executed.
+
+    - `SKIPPED`: A `SKIPPED` job has not been executed because its outputs were already computed.
+
+    - `ABANDONED`: An `ABANDONED` job has not been executed because it depends on a job that could not complete (
+        cancelled, failed, or abandoned).
```
jrobinAV: todo: review

jrobinAV (commented):
It looks good to me. 
It needs to be completed, in particular the unit tests.

src/taipy/core/_orchestrator/_orchestrator.py
toan-quach: I'm thinking we can move it to the `def status` setter in Job (which is what you proposed I believe), what do you think?

src/taipy/core/submission/submission.py
toan-quach: My thinking for it is that, we will create a Submission object before creating a job, then we will create jobs and propagate jobs to the submission object after that. So based on that flow, I think not having them in the Submission signature makes sense. But it can also be easier in different cases to have them in the signature as well, though don't think it's accurate for the use case

src/taipy/core/submission/submission.py
toan-quach: I think we can put it in the status setter function of Job entity

src/taipy/core/submission/_submission_converter.py
toan-quach: as mentioned before, I think the flow should be create a Submission entity to have a submission id before creating jobs, after that, we will create jobs with the new submission id, initially I want to make jobs to be a readable value only as well and not changeable, maybe we can discuss this more to align our thoughts :D

src/taipy/core/submission/submission.py
jrobinAV: Without the creation_date and the status in the signature, when reading a submission from our repo, we will first instantiate a submission with `status` equals to `UNDEFINED` and `creation_date` equals to `datetime.now()`. Immediately after, we will set the proper values. This is strange to me.
Plus it is not consistent with other entities.

src/taipy/core/submission/submission.py
jrobinAV: you are right! in status setter it is better.

tests/core/submission/test_submission.py
```diff
@@ -0,0 +1,110 @@
...
+    assert submission_2.id == "submission_id"
+    assert submission_2.entity_id == scenario.id
+    assert submission_2._jobs == [job]
+    assert submission_2.creation_date == current_datetime
+    assert submission_2._submission_status == SubmissionStatus.COMPLETED
+    assert submission_2._version == "version_id"
+
+
+def test_update_submission_status():
+    pass
```
jrobinAV: todo

src/taipy/core/_orchestrator/_orchestrator.py
```diff
@@ -110,41 +116,41 @@ def submit_task(
...
             cls._check_and_execute_jobs_if_development_mode()
         else:
             if wait:
                 cls.__wait_until_job_finished(job, timeout=timeout)
+
         return job
 
     @classmethod
-    def _submit_task(
+    def _submit_task(  # TODO: rename to create job
```
jrobinAV: change name to `lock_output_and_create_job`

tests/core/notification/test_events_published.py
jrobinAV, line 1: This test is not maintainable enough. We need to split it. Let me push a proposal directly on the branch if that s fine for you.

tests/core/notification/test_notifier.py
```diff
@@ -594,10 +594,26 @@ def test_publish_event():
     while registration_queue.qsize() != 0:
         published_events.append(registration_queue.get())
```
jrobinAV, line 656: This test is not maintainable anymore. Can you split it?

tests/core/notification/test_events_published.py
```diff
@@ -84,29 +84,32 @@ def test_event_published():
     # Submit a scenario trigger 1 scenario submission event
     # 7 data node update events (for last_edit_date, editor_id(x2), editor_expiration_date(x2) and edit_in_progress(x2))
     # 1 job creation event and 3 job update events (for status: PENDING, RUNNING and COMPLETED)
+
     sc.submit()
-    assert_true_after_time(lambda: all_evts.event_collected == 22, time=10)
+    assert_true_after_time(lambda: all_evts.event_collected == 30, time=10)
```
jrobinAV: There are too many events. We publish 6 (`SUBMISSION`, `UPDATE`, `submission_status`) events while only 3 should be published.
I don't know why.

src/taipy/core/job/job.py
```diff
@@ -117,6 +117,10 @@ def status(self):
     @_self_setter(_MANAGER_NAME)
     def status(self, val):
         self._status = val
+        if self.submit_id:
+            from ..submission._submission_manager_factory import _SubmissionManagerFactory
+
+            _SubmissionManagerFactory._build_manager()._get(self.submit_id).update_submission_status()
```
jrobinAV: BUG!
This does not work.
The self_setter decorator works in 3 steps: 
1. It executes the `job.status(val)` property code
2. It sets the job into the repo
3. It publishes the update event

That means in our case: 
1. So first we set the `job._status` in memory only.
    Then we call `update_submission_status` which gets the job from the repo (with the 
    previous job status value) to compute the new `submission_status` which is incorrect.
    Then we set the  INCORRECT `submission.submission_status` into the repository.
2. We set the job writing the new job status into the repository
3. We publish an update event

tests/core/notification/test_events_published.py
jrobinAV: I believe it comes from the auto setter decorator that execute the function twice.

tests/core/notification/test_events_published.py
```diff
@@ -84,29 +84,32 @@ def test_event_published():
...
-    assert_true_after_time(lambda: all_evts.operation_collected[EventOperation.UPDATE] == 14, time=10)
+    assert_true_after_time(lambda: all_evts.entity_type_collected[EventEntityType.SUBMISSION] == 8, time=10)
+    assert_true_after_time(lambda: all_evts.operation_collected[EventOperation.CREATION] == 8, time=10)
+    assert_true_after_time(lambda: all_evts.operation_collected[EventOperation.UPDATE] == 21, time=10)
     assert_true_after_time(lambda: all_evts.operation_collected[EventOperation.SUBMISSION] == 1, time=10)
 
     # Delete a scenario trigger 7 update events
     tp.delete(scenario.id)
-    assert_true_after_time(lambda: all_evts.event_collected == 29, time=10)
+    assert_true_after_time(lambda: all_evts.event_collected == 37, time=10)
```
jrobinAV: The scenario deletion should also trigger the submission deletion. We forgot to implement it.

tests/core/notification/test_events_published.py
jrobinAV, line 1: This test is False. It is hiding some bugs.

jrobinAV (changes requested):

tests/core/notification/test_events_published.py
toan-quach: yep you're right. I've been trying to see if we can fix it but no luck so far :/

tests/core/notification/test_events_published.py
toan-quach: done!

src/taipy/core/submission/submission_status.py
```diff
@@ -0,0 +1,45 @@
...
+
+    - `RUNNING`: A `RUNNING` submission has its jobs currently being executed.
+
+    - `CANCELED`: A `CANCELED` submission has been submitted but its execution has been canceled.
+
+    - `FAILED`: A `FAILED` submission has a job failed during its execution.
+
+    - `COMPLETED`: A `COMPLETED` submission has successfully been executed.
+    """
+
```
jrobinAV, line 41: Add SUBMITTED Status

src/taipy/core/submission/submission.py
```diff
@@ -140,28 +141,40 @@ def _update_submission_status(self, plop: Job):
             if not job:
                 continue
             if job.is_failed():
-                submission_status = SubmissionStatus.FAILED
-                break
+                self.submission_status = SubmissionStatus.FAILED  # type: ignore
```
jrobinAV, line 144: We should type the argument in the signature of the submission_status property setter. So we don't need the type ignore.

toan-quach: I believe I did but it still raises the attribute is read-only :/

src/taipy/core/submission/submission.py
```diff
@@ -0,0 +1,180 @@
...
+    @submission_status.setter  # type: ignore
+    @_self_setter(_MANAGER_NAME)
+    def submission_status(self, submission_status):
+        self._submission_status = submission_status
+
+    def __lt__(self, other):
+        return self.creation_date.timestamp() < other.creation_date.timestamp()
+
+    def __le__(self, other):
+        return self.creation_date.timestamp() == other.creation_date.timestamp() or self < other
```
gmarabout: 🥜 Why not simply `return self.creation_date.timestamp() <= other.creation_date.timestamp()`?

src/taipy/core/submission/submission.py
```diff
@@ -0,0 +1,180 @@
...
+
+    def __gt__(self, other):
+        return self.creation_date.timestamp() > other.creation_date.timestamp()
+
+    def __ge__(self, other):
+        return self.creation_date.timestamp() == other.creation_date.timestamp() or self > other
```
gmarabout: 🥜 Why not simply return self.creation_date.timestamp() >= other.creation_date.timestamp()?

src/taipy/core/submission/submission.py
```diff
@@ -0,0 +1,180 @@
...
+
+    _ID_PREFIX = "SUBMISSION"
+    _MANAGER_NAME = "submission"
+    __SEPARATOR = "_"
+
+    def __init__(
+        self,
+        entity_id: str,
+        id: Optional[str] = None,
+        jobs: Optional[Union[List[Job], List[JobId]]] = None,
```
gmarabout, line 47: ❓Just by curiosity, do we really need to have `Union[List[Job], List[JobId]]`? Why not a simpler `List[JobId]`? 
Having a more precise parameter type lower the risk of bugs in the long term.
M2c

src/taipy/core/submission/submission.py
```diff
@@ -0,0 +1,180 @@
...
+    def __le__(self, other):
+        return self.creation_date.timestamp() == other.creation_date.timestamp() or self < other
+
+    def __gt__(self, other):
+        return self.creation_date.timestamp() > other.creation_date.timestamp()
+
+    def __ge__(self, other):
+        return self.creation_date.timestamp() == other.creation_date.timestamp() or self > other
+
+    def _update_submission_status(self, _: Job):
```
gmarabout, line 132: ⚠️ This method is really hard to read and hard to understand. 
Why not something like this?
```
  if job.is_canceled():
      self.submission_status = SubmissionStatus.CANCELED
  if job.is_blocked():
      self.submission_status = SubmissionStatus.BLOCKED
  if job.is_pending() or job.is_submitted():
      self.submission_status = SubmissionStatus.PENDING
  if job.is_running():
      self.submission_status = SubmissionStatus.RUNNING
  if job.is_completed() or job.is_skipped():
      self.submission_status = SubmissionStatus.COMPLETED
  if job.is_abandoned():
      self.submission_status = SubmissionStatus.UNDEFINED
```

src/taipy/core/submission/submission.py
```diff
@@ -0,0 +1,180 @@
...
+        completed = False
+
+        for job in self.jobs:
+            if not job:
+                continue
+            if job.is_failed():
+                self.submission_status = SubmissionStatus.FAILED  # type: ignore
+                return
+            if job.is_canceled():
+                canceled = True
```
gmarabout, line 147: ⚠️ All the other cases have a `continue`. This one hasn't. Is it intended?

gmarabout (commented):
Hey 👋🏻 
I just added some small comments on the code. 
Maybe a quick description of the PR would have helped me provide you with a better review :-)

jrobinAV (dismissed):

src/taipy/core/submission/submission.py
toan-quach: I want to offer both possibility to pass in the Job entity or just the JobId instead of asking the user for 1 of the 2 if that makes sense to you :D since we use entity and id interchangeably, enforcing just one might be going away from the overall behavior. Though it probably still makes sense in this context as Submission is like for internal use only (at least for now)

src/taipy/core/submission/submission.py
toan-quach: haha thanks for pointing it out, I think we don't need any of the "continue"s after all!

src/taipy/core/submission/submission.py
toan-quach: I believe that's so we can loop through all of the jobs and prioritize some of the statuses. Like for example, if we have a list of canceled jobs, but 1 of them is a failed job, we would want to prioritize the failing status instead of the cancel one. Hope it make sense! 

@jrobinAV can you help fill in if I missed any detail?

src/taipy/core/submission/submission.py
toan-quach: yeah... my bad that was weird...

src/taipy/core/submission/submission.py
toan-quach: thanks for pointing this out! I noticed the same issue on Job, will be fixing that as well!

src/taipy/core/submission/submission.py
gmarabout: I leave you the final decision of course, and it's not a big deal anyway :-)

But as a user, I think I would prefer clear and explicit API. 

With the one you propose, the caller may wonder if passing IDs or Jobs would make the function work differently. 
For example, would the function avoid any additional data loading if we pass the job?
Again, not really a pb, and we can deal with this API :-)

jrobinAV: If we want to change this convention, or at least discuss it, it should be a global decision. It is out of this PR scope.
My2cents

toan-quach: I agree :D as the behavior is quite global, it's not the best place to discuss it!

src/taipy/core/submission/submission.py
jrobinAV: I don't know what to say. 😄 
The behavior of your code proposal @gmarabout, is not the same as what we implemented. 
It does not respect what we want to implement.
We have to parse all the jobs before computing the submission_status.

gmarabout: > It does not respect what we want to implement.

That's is my point: the intent of the method is not obvious when you read it.

toan-quach: @gmarabout 
 would renaming it to `_update_submission_status_based_on_jobs` help?

gmarabout: Definitely not :-) 
The name should reflect the intent of the method, what it is supposed to do.
Not *how* it does it. 
What if we decide to change the logic in the future?
The code should definitely speak by itself about the *how*...
If you think the code is clear enough, then let's go. I don't want to block this PR.

toan-quach: alright thanks Gregoire! I'll merge it for now and think how i can improve it later

jrobinAV (approved):

PR #824, opened 2023-11-10
Feature/#755 - Open append() method on datanodes

trgiangdo:
https://github.com/Avaiga/taipy/issues/408

This PR opens new `.append()` API for CSV, Excel, Json, MongoCollection, SQLTable data node types.

Notes on other data node types:
- GenericDataNode, we can open a new attribute to let the user define their own append method. Let me know if we want to do this.
- InMemoryDataNode, we don't know what kind of data stored in memory.
- PickleDataNode, we don't know what kind of data stored in the pickle file.
- SQLDataNode, the user can just modify the read query to append data.
- ParquetDataNode, the current method is to read the data, append the data, then rewrite, which is slow and not working properly with modin. Using `fastparquet`, there is a simpler way to do this, but I'm working on finding a better way using `pyarrow`, which is our default engine.

### DIFF
--- a/src/taipy/core/config/data_node_config.py
+++ b/src/taipy/core/config/data_node_config.py
@@ -120,6 +120,7 @@ class DataNodeConfig(Section):
     # SQL
     _REQUIRED_READ_QUERY_SQL_PROPERTY = "read_query"
     _REQUIRED_WRITE_QUERY_BUILDER_SQL_PROPERTY = "write_query_builder"
+    _OPTIONAL_APPEND_QUERY_BUILDER_SQL_PROPERTY = "append_query_builder"
     # MONGO
     _REQUIRED_DB_NAME_MONGO_PROPERTY = "db_name"
     _REQUIRED_COLLECTION_NAME_MONGO_PROPERTY = "collection_name"
@@ -207,6 +208,7 @@ class DataNodeConfig(Section):
             _OPTIONAL_HOST_SQL_PROPERTY: "localhost",
             _OPTIONAL_PORT_SQL_PROPERTY: 1433,
             _OPTIONAL_DRIVER_SQL_PROPERTY: "",
+            _OPTIONAL_APPEND_QUERY_BUILDER_SQL_PROPERTY: None,
             _OPTIONAL_FOLDER_PATH_SQLITE_PROPERTY: None,
             _OPTIONAL_FILE_EXTENSION_SQLITE_PROPERTY: ".db",
             _OPTIONAL_DB_EXTRA_ARGS_SQL_PROPERTY: None,
@@ -867,6 +869,7 @@ class DataNodeConfig(Section):
         db_engine: str,
         read_query: str,
         write_query_builder: Callable,
+        append_query_builder: Optional[Callable] = None,
         db_username: Optional[str] = None,
         db_password: Optional[str] = None,
         db_host: Optional[str] = None,
@@ -889,7 +892,9 @@ class DataNodeConfig(Section):
                 or *"postgresql"*.
             read_query (str): The SQL query string used to read the data from the database.
             write_query_builder (Callable): A callback function that takes the data as an input parameter
-                and returns a list of SQL queries.
+                and returns a list of SQL queries to be executed when writing data to the data node.
+            append_query_builder (Optional[Callable]): A callback function that takes the data as an input parameter
+                and returns a list of SQL queries to be executed when appending data to the data node.
             db_username (Optional[str]): The database username. Required by the *"mssql"*, *"mysql"*, and
                 *"postgresql"* engines.
             db_password (Optional[str]): The database password. Required by the *"mssql"*, *"mysql"*, and
@@ -927,6 +932,8 @@ class DataNodeConfig(Section):
             }
         )
 
+        if append_query_builder is not None:
+            properties[cls._OPTIONAL_APPEND_QUERY_BUILDER_SQL_PROPERTY] = append_query_builder
         if db_username is not None:
             properties[cls._OPTIONAL_DB_USERNAME_SQL_PROPERTY] = db_username
         if db_password is not None:
--- a/src/taipy/core/data/sql.py
+++ b/src/taipy/core/data/sql.py
@@ -17,7 +17,7 @@ from sqlalchemy import text
 from taipy.config.common.scope import Scope
 
 from .._version._version_manager_factory import _VersionManagerFactory
-from ..exceptions.exceptions import MissingRequiredProperty
+from ..exceptions.exceptions import MissingAppendQueryBuilder, MissingRequiredProperty
 from ._abstract_sql import _AbstractSQLDataNode
 from .data_node_id import DataNodeId, Edit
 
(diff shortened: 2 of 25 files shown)


jrobinAV:
On SQLDataNode, I believe we need 2 different methods for writing and appending data. Writting will still need to write everything. The orchestrator will require that. Then the append method should be used for the manual edition.
Same for read and filter...

trgiangdo:
According to https://docs.taipy.io/en/latest/manuals/core/config/data-node-config/#example-with-a-microsoft-sql-database-table_1, not really.

In that example, the `write_query_builder()` will return 2 queries, one delete all data, and the next one will write the data.

There is no way to know for sure which table to delete data from when writing SQLDataNode, so we can not force it I think

jrobinAV:
> According to https://docs.taipy.io/en/latest/manuals/core/config/data-node-config/#example-with-a-microsoft-sql-database-table_1, not really.
> 
> In that example, the `write_query_builder()` will return 2 queries, one delete all data, and the next one will write the data.
> 
> There is no way to know for sure which table to delete data from when writing SQLDataNode, so we can not force it I think

My point is that the SqlDataNode must expose two methods like any other Data node: `write(all_data)` and `append(data)`.
- The `write` basically executes the 2 SQL queries returned by the `write_query_builder()`.
- The `append` should execute one SQL query as well. This query can be different from the second query of the `write_query_builder()`. That means we need an `append_query_builder()`. Am I missing something?

trgiangdo:
Oh, then it can be something like `sql_dn.append(data, append_query_builder)` where append_query_builder is a different method.
The signature will be different from other data nodes though

jrobinAV:
No, the append_query_builder can be set in the data node config, like the write_query_builder, but it will be executed at run time.
Something like: 

The configure method could look like:
```
sales_history_cfg = Config.configure_sql_data_node(
    id="sales_history",
    db_username="admin",
    db_password="password",
    db_name="taipy",
    db_engine="mssql",
    read_query="SELECT * from sales",
    write_query_builder=write_query_builder,
    append_query_builder=append_query_builder,
    db_driver="ODBC Driver 17 for SQL Server",
    db_extra_args={"TrustServerCertificate": "yes"},
)
```

The data node implementation would look like this:
```
    def _do_append(self, data) -> None:
        queries = self.properties.get(self._APPEND_QUERY_BUILDER_KEY)(data)
        for query in queries:
            connection.execute(query)
```

What do you think ?

trgiangdo:
I see, that makes much more sense. Thank you.

jrobinAV (commented):
The test coverage is great! Thx!

jrobinAV (dismissed):

trgiangdo:
In a few latest commits:
- Add `append()` method to ParquetDataNode. However, this only works if the `fastparquet` package is installed.
- Add `append()` method to SQLDataNode, which uses the queries return by the `append_query_builder` provided at the config level.

tests/core/data/test_parquet_data_node.py
```diff
@@ -300,6 +300,50 @@ def test_get_system_folder_modified_date_instead_of_last_edit_date(self, tmpdir_
...
+        "content",
+        [
+            ([{"a": 11, "b": 22, "c": 33}, {"a": 44, "b": 55, "c": 66}]),
+            (pd.DataFrame([{"a": 11, "b": 22, "c": 33}, {"a": 44, "b": 55, "c": 66}])),
+        ],
+    )
+    def test_append_pandas(self, parquet_file_path, default_data_frame, content):
+        # !!! append data only works with `fastparquet`
+        if not util.find_spec("fastparquet"):
+            pass
```
jrobinAV: Maybe we can raise a proper error when using append without fastparquet, to make it clean.

src/taipy/core/data/sql.py
```diff
@@ -126,8 +130,15 @@ def storage_type(cls) -> str:
     def _get_base_read_query(self) -> str:
         return self.properties.get(self.__READ_QUERY_KEY)
 
+    def _do_append(self, data, engine, connection) -> None:
+        queries = self.properties.get(self._APPEND_QUERY_BUILDER_KEY)(data)
```
jrobinAV, line 137: We don't have guaranty that _APPEND_QUERY_BUILDER_KEY is provided in properties.

src/taipy/core/config/data_node_config.py
```diff
@@ -120,6 +120,7 @@ class DataNodeConfig(Section):
     # SQL
     _REQUIRED_READ_QUERY_SQL_PROPERTY = "read_query"
     _REQUIRED_WRITE_QUERY_BUILDER_SQL_PROPERTY = "write_query_builder"
+    _OPTIONAL_APPEND_QUERY_BUILDER_SQL_PROPERTY = "append_query_builder"
```
jrobinAV, line 123: If it is optional, what happens when dn.append(data) is called? DO we have a default behavior? do we raise an error? Do we want to make it required?

trgiangdo: We definitely do not want to make this required. I should and will raise an error on `append()` method if the attribute is not provided.

src/taipy/core/data/sql.py
trgiangdo: I should and will raise an error if the attribute is not provided.

tests/core/data/test_parquet_data_node.py
trgiangdo: Oh this is for the test. If someone install taipy-core and do not install the optional fastparquet library, the useless error will not be raised

jrobinAV: And what if a user install taipy-core with fastparquet? What is the behavior when calling append?

trgiangdo: Then whatever the engine defined at the config level is, the append method will use fastparquet.

This is to avoid the situation when the user wants to use pyarrow for a specific reason, but still want to append data, then we do allow that.

jrobinAV (dismissed):

toan-quach (approved):

tests/core/data/test_parquet_data_node.py
```diff
@@ -300,6 +300,44 @@ def test_get_system_folder_modified_date_instead_of_last_edit_date(self, tmpdir_
...
+    )
+    def test_append_pandas(self, parquet_file_path, default_data_frame, content):
+        dn = ParquetDataNode("foo", Scope.SCENARIO, properties={"path": parquet_file_path})
+        assert_frame_equal(dn.read(), default_data_frame)
+
+        dn.append(content)
+        assert_frame_equal(
+            dn.read(),
+            pd.concat([default_data_frame, pd.DataFrame(content, columns=["a", "b", "c"])]).reset_index(drop=True),
+        )
```
jrobinAV, line 319: I was more thinking abt something like that, to be sure we are properly catching and logging the error. What do you think?

```suggestion
        if not util.find_spec("fastparquet"):
            with pytest.raises(FastParquetRequiredError):
                dn.append(content)
        else:
            dn.append(content)
            assert_frame_equal(
                dn.read(),
                pd.concat([default_data_frame, pd.DataFrame(content, columns=["a", "b", "c"])]).reset_index(drop=True),
            )
```

trgiangdo: If the user tries to append to ParquetDN without installing fastparquet, Python will raise the error saying that parquet is not installed.

We will also have a warning in the doc saying that fastparquet is required for appending.

jrobinAV (approved):

PR #829, opened 2023-11-20
Improve Core Events to support additional metadata 

gmarabout:
# Goals

We would like to add additional data to core events so we can use it for telemetry. 
The data we want to add are: the `config_id` of the entity (will eventually be sent as an attribute), the possibility to add extra data, especially for job events (like duration/latency).

# What the PR solves

We will need to create events with different (meta)data depending on the entity type.
The problem is that the method `_publish_event` takes a bunch of data that the callers must provide in a non-generic way. It means that if we add one field to an event, then we have to change it everywhere `_publish_event` is used. 
Moreover, the method being static, it is difficult for the dynamic `_Reloader` and `_Properties` classes to have a generic code. 

The biggest change in this PR is the attempt to make a central piece where the logic of creating an event is delegated to a `_make_event` function which is "overridden" per entity type using the built-in Python `singledispatch` mechanism.
This function could eventually totally replace the static `_publish_event`.

For now, the `_publish_event` is still there for few cases. 

# Changes

* Simplification: Class `Event` is now a frozen dataclass
* Added `Event.attribute_value` to contain the value of a changed attribute
* Added `Event.metadata`, a dict to contain any extra information we would like to transmit via the event (ex: `task_config_id` for job events)./
* Introduce a "single dispatch" function `_make_event` to create an event for a particular entity type. The [base function](https://github.com/Avaiga/taipy-core/pull/829/files#diff-347b29c61420e075036f7934d09389441d3fe858fd3a07644b09151622bec36dR107) corresponds to the "default" case, and will raise an error if called. The overloaded functions are implemented within each entity Python file (in `job.py` for jobs, in `scenario.py` for scenarios, etc) and will implement the logic of creating an event from this specific entity. This method is intended to centralize event creation.


src/taipy/core/notification/event.py
```diff
@@ -71,30 +74,60 @@ class Event:
...
+                <li>DELETION</li>
+                <li>SUBMISSION</li>
+            </ul>
+        attribute_name (Optional[str]): The name of the updated attribute for a `EventOperation.UPDATE`.
+            This argument is always given in case of an UPDATE.
+        attribute_value (Optional[Any]): The value of the udated attribute for a `EventOperation.UPDATE`.
+            This argument is always given in case of an UPDATE.
+        **kwargs (dict[str, any]): Any extra information that would be passed to the metadata event.
+            Note: you should pass only simple types: str, float, double as values."""
+    raise Exception(f"Unexpected entity type: {type(entity)}")
```
gmarabout, line 134: *Notes to reviewers*:
This function must be overloaded for each entity supporting event creation. This one would be called as a fallback if the interpreter does not find any for a entity type.

src/taipy/core/_entity/_entity.py
```diff
@@ -34,6 +34,6 @@ def __exit__(self, exc_type, exc_value, exc_traceback):
             self._properties.data.update(self._properties._pending_changes)
         _get_manager(self._MANAGER_NAME)._set(self)
 
-        while self._in_context_attributes_changed_collector:
-            _publish_event(*self._in_context_attributes_changed_collector.pop(0))
+        for event in self._in_context_attributes_changed_collector:
+            Notifier.publish(event)
```
gmarabout, line 38: *Note to reviewers*: Adding more arguments to the `Event` class make the use of a arg list a bit cumbersome and very error prone (think the order being significant!).
That is why I propose to replace this list of args by a list of `Event` objects. Then, the logic does not change, except that it makes more sense to call `Notifier.publish` directly now...

src/taipy/core/_manager/_manager.py
```diff
@@ -44,7 +48,12 @@ def _delete_many(cls, ids: Iterable):
         cls._repository._delete_many(ids)
         if hasattr(cls, "_EVENT_ENTITY_TYPE"):
             for entity_id in ids:
-                _publish_event(cls._EVENT_ENTITY_TYPE, entity_id, EventOperation.DELETION, None)  # type: ignore
+                _publish_event(
+                    cls._EVENT_ENTITY_TYPE,  # type: ignore
+                    EventOperation.DELETION,
+                    entity_id=entity_id,
+                    delete_all=True,
```
gmarabout: *Note to reviewers*: I think we should gradually replace these `_publish_events` with the more readable and (more important) more generic `Notifier.publish(make_event(entity))`. 
what do you think?

src/taipy/core/notification/event.py
```diff
@@ -58,6 +60,7 @@ class EventEntityType(_ReprEnum):
 }
 
 
+@dataclass(frozen=True)
```
gmarabout, line 65: *Note to reviewers*: the use of dataclass makes the definition of this (read-only) class sleaker. It also provides for free default `repr` and `hash`.

src/taipy/core/notification/notifier.py
```diff
@@ -19,11 +19,38 @@
 
 def _publish_event(
     entity_type: EventEntityType,
```
gmarabout, line 21: *note to reviewers*: the only two mandatory arguments are the `entity_type` and the `operation`. I propose to move them at first position, and be the only positional arguments for better readability. All the others can be specified using the named argument notation (including potential extra medatata).

src/taipy/core/_manager/_manager.py
jrobinAV: I don't have strong opinion on it but I don't believe it is more readable. It is not less readable either. Just a bit longer to me.

The method `_publish_event` makes the event and publishes it in a single call, while your proposal is to split it into 2 calls.

We could make the `_publish_event` a single dispatch method directly to apply your changes (passing the entity directly to add more fields and metadata) and keep a the two steps in a single call.

It's just a minor suggestion, but not at all a deal breaker. It's up to you to keep it as it is or not.

src/taipy/core/notification/event.py
```diff
@@ -71,30 +74,60 @@ class Event:
...
+        # Check attribute name:
+        if self.operation in _NO_ATTRIBUTE_NAME_OPERATIONS and self.attribute_name is not None:
+            raise InvalidEventAttributeName
+
+
+@singledispatch
+def make_event(
+    entity: Any,
+    operation: EventOperation,
+    /,
```
jrobinAV, line 111: I don't know this `/,` param. Is it a way to split mandatory positional parameters and optional parameters for readability? Is it standard?

src/taipy/core/job/_job_manager.py
```diff
@@ -53,7 +52,7 @@ def _create(
             version=version,
         )
         cls._set(job)
-        _publish_event(cls._EVENT_ENTITY_TYPE, job.id, EventOperation.CREATION, None)
+        _publish_event(EventEntityType.JOB, EventOperation.CREATION, entity_id=job.id, config_id=job._task.config_id)
```
jrobinAV: I would not pass the task configuration identifier as the `config_id` parameter for a job event. This looks strange to me. I would put it in the metadata. 
`config_id` is used differently for other entities. It corresponds to the actual configuration object used to generate the entity. It is different for the jobs.

Is there any issue with naming the event attribute `task_config_id`?

src/taipy/core/job/job.py
```diff
@@ -351,3 +355,24 @@ def is_deletable(self) -> bool:
...
+    /,
+    attribute_name: Optional[str] = None,
+    attribute_value: Optional[Any] = None,
+    **kwargs,
+) -> Event:
+    metadata = {"creation_date": job.creation_date}
+    return Event(
+        entity_type=EventEntityType.JOB,
+        entity_id=job.id,
+        config_id=job._task.config_id,  # OK
```
jrobinAV: The previous comment may have an impact on this line.

jrobinAV: Please remove `# OK`

src/taipy/core/notification/event.py
```diff
@@ -71,30 +74,60 @@ class Event:
...
             raise InvalidEventOperation
-        return operation
+
+        # Check attribute name:
+        if self.operation in _NO_ATTRIBUTE_NAME_OPERATIONS and self.attribute_name is not None:
+            raise InvalidEventAttributeName
+
+
+@singledispatch
+def make_event(
```
jrobinAV: We should rename the method `_make_event` for now. Once we want to open this as an API for users, we can name it without the `_` prefix, but this will require a lot of documentation, and compatibility with future versions. I believe we are not ready for that.

src/taipy/core/notification/event.py
gmarabout: Yes, it is standard way to separate positional only from named parameters...
See https://peps.python.org/pep-0570/

src/taipy/core/job/_job_manager.py
gmarabout: We can pass it to medata if it makes more sense.
And yes we can name it `task_config_id`.

However, for a scenario or a data node, does it make sense NOT to give the "config ID"? How do we know which scenario or datanode configs we are talking about?

src/taipy/core/_manager/_manager.py
gmarabout: The problem I point at is not the fact that there is two calls instead of one (we can make it a single call if you want to).
The problem is that the caller of `_publish_event` must currently implement all the logic: how to get the ID, how to get the metadata/extra information for the event etc.
As a consequence, this logic (which is very entity related) is scattered all over the code base. Just search for all the calls for `_publish_event` and you get an idea of what needs to change if you want to add just one more field!
My proposal is to make it more centralized and generic.
My two cents.

src/taipy/core/job/_job_manager.py
jrobinAV: Scenarios and datanodes depend on a configuration so they have a `config_id` attribute.
It does NOT make sense not to provide the config ID. It must be passed.

Jobs don't depend on a configuration. They depend on a task that depends on a configuration.
What you want to provide here is not the `config_id` of the job (it does not exist) but the `config_id` of the task attached to the job.

Does it make sens to you?

src/taipy/core/notification/event.py
gmarabout: Ok, but notice that `Notifier.publish(event)` is public and documented...

src/taipy/core/notification/event.py
gmarabout: Actually, your remark makes me think the `*` might have been more appropriate in this case... 🤔 

```
def f(pos1, pos2, /, pos_or_kwd, *, kwd1, kwd2):
      -----------    ----------     ----------
        |             |                  |
        |        Positional or keyword   |
        |                                - Keyword only
         -- Positional only
```

src/taipy/core/job/_job_manager.py
gmarabout: Yes, it does !
Thank you for your feedback!

src/taipy/core/notification/event.py
gmarabout: The method has been renamed into `_make_event`.

jrobinAV (approved):

src/taipy/core/_entity/_entity.py
toan-quach: that's awesome!!

src/taipy/core/notification/event.py
toan-quach: hmmm interesting, do you think it make sense for us to do this across all other functions?

toan-quach (approved):
Looks really good to me!!! also learnt something new too! Thanks @gmarabout !

src/taipy/core/notification/event.py
gmarabout: Honestly, I can't say. This would depend on the case I guess. 
But beware that applying this to existing functions would break API compatibility.
Let's discuss this offline.

joaoandre-avaiga (approved):

PR #832, opened 2023-11-27
Feature/new update submission status algorithm

toan-quach:
## Purpose:

This PR is an attempt to improve the submission status change algorithm performance.

## Changes:

Instead of looping over all jobs of a submission entity when 1 job status is updated, we will update the submission status based only on the recently updated job status.

## Remaing tasks:

- Looking into adding check for submission status in `test_orchestrator.py`

## Result

Previously, when running `pytest tests/enterprise/core/test_end_to_end.py::test_without_authorization_standalone` on my local (the previous test used during our discussion), it took 22.55s to run, after applying this changes, it now takes only 5.39s to run!

**Results from running profiling:**

### ------------ Old version ------------
Web capture_27-11-2023_202952_127 0 0 1

### ------------ New version ------------
Web capture_27-11-2023_202959_127 0 0 1

### DIFF
--- a/src/taipy/core/_orchestrator/_orchestrator.py
+++ b/src/taipy/core/_orchestrator/_orchestrator.py
@@ -67,7 +67,10 @@ class _Orchestrator(_AbstractOrchestrator):
         Returns:
             The created Jobs.
         """
-        submission = _SubmissionManagerFactory._build_manager()._create(submittable.id, submittable._ID_PREFIX)  # type: ignore
+        submission = _SubmissionManagerFactory._build_manager()._create(
+            submittable.id, submittable._ID_PREFIX  # type: ignore
+        )
+
         jobs = []
         tasks = submittable._get_sorted_tasks()
         with cls.lock:
--- a/src/taipy/core/_orchestrator/_orchestrator.py
+++ b/src/taipy/core/_orchestrator/_orchestrator.py
@@ -225,7 +225,6 @@ class _Orchestrator(_AbstractOrchestrator):
         if job.is_completed() or job.is_skipped():
             cls.__unblock_jobs()
         elif job.is_failed():
-            print(f"\nJob {job.id} failed, abandoning subsequent jobs.\n")
             cls._fail_subsequent_jobs(job)
 
     @classmethod
@@ -298,7 +297,6 @@ class _Orchestrator(_AbstractOrchestrator):
                 cls.__find_subsequent_jobs(failed_job.submit_id, set(failed_job.task.output.keys()))
             )
             for job in to_fail_or_abandon_jobs:
-                print(f"Abandoning job: {job.id}")
                 job.abandoned()
             to_fail_or_abandon_jobs.update([failed_job])
             cls.__remove_blocked_jobs(to_fail_or_abandon_jobs)
--- a/src/taipy/core/_orchestrator/_orchestrator.py
+++ b/src/taipy/core/_orchestrator/_orchestrator.py
@@ -67,7 +67,7 @@ class _Orchestrator(_AbstractOrchestrator):
         Returns:
             The created Jobs.
         """
-        submission = _SubmissionManagerFactory._build_manager()._create(submittable.id)  # type: ignore
+        submission = _SubmissionManagerFactory._build_manager()._create(submittable.id, submittable._ID_PREFIX)  # type: ignore
         jobs = []
         tasks = submittable._get_sorted_tasks()
         with cls.lock:
@@ -118,7 +118,7 @@ class _Orchestrator(_AbstractOrchestrator):
         Returns:
             The created `Job^`.
         """
-        submission = _SubmissionManagerFactory._build_manager()._create(task.id)
+        submission = _SubmissionManagerFactory._build_manager()._create(task.id, task._ID_PREFIX)
         submit_id = submission.id
         with cls.lock:
             job = cls._lock_dn_output_and_create_job(
--- a/src/taipy/core/submission/_submission_converter.py
+++ b/src/taipy/core/submission/_submission_converter.py
@@ -25,7 +25,7 @@ class _SubmissionConverter(_AbstractConverter):
             id=submission.id,
             entity_id=submission._entity_id,
             entity_type=submission.entity_type,
-            job_ids=submission._job_ids,
+            job_ids=[job.id if isinstance(job, Job) else JobId(str(job)) for job in list(submission._jobs)],
             creation_date=submission._creation_date.isoformat(),
             submission_status=submission._submission_status,
             version=submission._version,
--- a/src/taipy/core/submission/_submission_converter.py
+++ b/src/taipy/core/submission/_submission_converter.py
@@ -24,7 +24,8 @@ class _SubmissionConverter(_AbstractConverter):
         return _SubmissionModel(
             id=submission.id,
             entity_id=submission._entity_id,
-            job_ids=[job.id if isinstance(job, Job) else JobId(str(job)) for job in list(submission._jobs)],
+            entity_type=submission.entity_type,
+            job_ids=submission._job_ids,
             creation_date=submission._creation_date.isoformat(),
             submission_status=submission._submission_status,
             version=submission._version,
@@ -34,6 +35,7 @@ class _SubmissionConverter(_AbstractConverter):
     def _model_to_entity(cls, model: _SubmissionModel) -> Submission:
         submission = Submission(
             entity_id=model.entity_id,
+            entity_type=model.entity_type,
             id=SubmissionId(model.id),
             jobs=model.job_ids,
             creation_date=datetime.fromisoformat(model.creation_date),
--- a/src/taipy/core/submission/submission.py
+++ b/src/taipy/core/submission/submission.py
@@ -18,7 +18,7 @@ from .._entity._labeled import _Labeled
 from .._entity._reload import _self_reload, _self_setter
 from .._version._version_manager_factory import _VersionManagerFactory
 from ..job._job_manager_factory import _JobManagerFactory
-from ..job.job import Job, JobId
+from ..job.job import Job, JobId, Status
 from ..notification.event import Event, EventEntityType, EventOperation, _make_event
 from .submission_id import SubmissionId
 from .submission_status import SubmissionStatus
@@ -157,27 +157,29 @@ class Submission(_Entity, _Labeled):
         if self.__failed or self.__canceled:
             return
 
-        if job.is_failed():
+        job_status = job.status
+
+        if job_status == Status.FAILED:
             self.submission_status = SubmissionStatus.FAILED  # type: ignore
             self.__failed = True
             return
-        if job.is_canceled():
+        if job_status == Status.CANCELED:
             self.submission_status = SubmissionStatus.CANCELED  # type: ignore
             self.__canceled = True
             return
 
-        if job.is_blocked():
+        if job_status == Status.BLOCKED:
             self.__blocked_jobs.add(job.id)
-        if job.is_pending() or job.is_submitted():
+        if job_status == Status.PENDING or job_status == Status.SUBMITTED:
             self.__pending_jobs.add(job.id)
             self.__blocked_jobs.discard(job.id)
-        if job.is_running():
+        if job_status == Status.RUNNING:
             self.__running_jobs.add(job.id)
             self.__pending_jobs.discard(job.id)
-        if job.is_completed() or job.is_skipped():
+        if job_status == Status.COMPLETED or job_status == Status.SKIPPED:
             self.__completed = True
             self.__running_jobs.discard(job.id)
-        if job.is_abandoned():
+        if job_status == Status.ABANDONED:
             self.__abandoned = True
 
         if self.__abandoned:
--- a/src/taipy/core/submission/submission.py
+++ b/src/taipy/core/submission/submission.py
@@ -9,6 +9,7 @@
 # an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the
 # specific language governing permissions and limitations under the License.
 
+import threading
 import uuid
 from datetime import datetime
 from typing import Any, List, MutableSet, Optional, Union
@@ -40,6 +41,7 @@ class Submission(_Entity, _Labeled):
     _ID_PREFIX = "SUBMISSION"
     _MANAGER_NAME = "submission"
     __SEPARATOR = "_"
+    lock = threading.Lock()
 
     def __init__(
         self,
@@ -54,15 +56,13 @@ class Submission(_Entity, _Labeled):
         self._entity_id = entity_id
         self._entity_type = entity_type
         self.id = id or self.__new_id()
-        self._job_ids = [job if isinstance(job, str) else job.id for job in jobs] if jobs else []
+        self._jobs: Union[List[Job], List[JobId], List] = jobs or []
         self._creation_date = creation_date or datetime.now()
         self._submission_status = submission_status or SubmissionStatus.SUBMITTED
         self._version = version or _VersionManagerFactory._build_manager()._get_latest_version()
 
         self.__abandoned = False
-        self.__canceled = False
         self.__completed = False
-        self.__failed = False
 
         self.__running_jobs: MutableSet[str] = set()
         self.__blocked_jobs: MutableSet[str] = set()
@@ -107,20 +107,15 @@ class Submission(_Entity, _Labeled):
         jobs = []
         job_manager = _JobManagerFactory._build_manager()
 
-        for job_id in self._job_ids:
-            jobs.append(job_manager._get(job_id))
+        for job in self._jobs:
+            jobs.append(job_manager._get(job))
 
         return jobs
-
-    @property  # type: ignore
-    @_self_reload(_MANAGER_NAME)
-    def job_ids(self) -> List[JobId]:
-        return self._job_ids
 
     @jobs.setter  # type: ignore
     @_self_setter(_MANAGER_NAME)
     def jobs(self, jobs: Union[List[Job], List[JobId]]):
-        self._job_ids = [job if isinstance(job, str) else job.id for job in jobs]
+        self._jobs = jobs
 
     def __hash__(self):
         return hash(self.id)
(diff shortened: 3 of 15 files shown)


src/taipy/core/submission/_submission_converter.py
```diff
@@ -24,7 +24,8 @@ def _entity_to_model(cls, submission: Submission) -> _SubmissionModel:
         return _SubmissionModel(
             id=submission.id,
             entity_id=submission._entity_id,
-            job_ids=[job.id if isinstance(job, Job) else JobId(str(job)) for job in list(submission._jobs)],
+            entity_type=submission.entity_type,
```
gmarabout, line 27: :heart: Cool!!

src/taipy/core/submission/submission.py
```diff
@@ -44,19 +44,30 @@ class Submission(_Entity, _Labeled):
...
 
+        self.__abandoned = False
+        self.__canceled = False
+        self.__completed = False
+        self.__failed = False
+
+        self.__running_jobs: MutableSet[str] = set()
+        self.__blocked_jobs: MutableSet[str] = set()
+        self.__pending_jobs: MutableSet[str] = set()
+
```
gmarabout, line 70: :question: Are the fields `__running_jobs`, `__blocked_jobs` and `__pending_jobs` persisted on disk/database? Is it safe for concurrent access?

src/taipy/core/submission/submission.py
```diff
@@ -130,56 +150,50 @@ def __gt__(self, other):
...
-            if job.is_completed() or job.is_skipped():
-                completed = True
-                continue
-            if job.is_abandoned():
-                abandoned = True
-        if canceled:
+    def _update_submission_status(self, job: Job):
+
+        if job.id not in self.job_ids:
+            return
```
gmarabout: 🤔 Could that really happen (receiving a job update for a job that is not in the list of jobs)? Shouldn't we raise an error? We should at least log something, IMHO.

src/taipy/core/submission/submission.py
```diff
@@ -44,19 +44,30 @@ class Submission(_Entity, _Labeled):
...
-        self._jobs: Union[List[Job], List[JobId], List] = jobs or []
+        self._job_ids = [job if isinstance(job, str) else job.id for job in jobs] if jobs else []
         self._creation_date = creation_date or datetime.now()
         self._submission_status = submission_status or SubmissionStatus.SUBMITTED
         self._version = version or _VersionManagerFactory._build_manager()._get_latest_version()
 
+        self.__abandoned = False
+        self.__canceled = False
+        self.__completed = False
+        self.__failed = False
```
gmarabout: 🤔 Those additional fields look redundant with there `_submission_status`...

src/taipy/core/submission/submission.py
gmarabout, line 70: ❓ What is the purpose of these additional `set` fields? As far as I can see, we add/remove job ids from it in `_update_submission_status` but we never read from it!
if it is just to track the statuses of the jobs, why not just using one single `Dict[JobId, Status]`?

gmarabout (commented):
There are actually two independent parts in this PR.

The first is to add the `entity_type` to the submission. 
I fully approve this part ! 👏🏻 

The second is to change how the status of the submission is updated. I have comments on how it's done: duplicated status information (the `_submission_status`and the various boolean fields), the additional sets to store job IDs per status that looks unused... I added some comments on this.

src/taipy/core/submission/submission.py
toan-quach: - It won't be persisted on the database so that we can avoid loading time.
- concurrent access is an interesting point, I think we should be safe but since there will be writing to the "job sets",  wouldnt hurt to put in some locking mechanism just in case. What do you think? @gmarabout

src/taipy/core/submission/submission.py
toan-quach: Haha nope I don't think it's possible! Let me remove it then haha

src/taipy/core/submission/submission.py
toan-quach: Oh it's just something to hold if we still have jobs running, blocked, or pending or not. As we still need access to other job statuses, if we use a dict, we will need to loop through all job statuses to do this check, while if we keep each status to an equivalent set, we can quickly identify the existence of "running jobs" or "blocked jobs", etc. in O(1) time. Does that make sense to you?

src/taipy/core/submission/submission.py
toan-quach: I have removed canceld and failed, but decided to keep abandoned and completed. I think they're still necessary, will look into if we can remove them also :D

src/taipy/core/submission/submission.py
gmarabout: Ok, but is it used anywhere in the code? I can't find any...
If it is not used, why do we need those sets?

gmarabout: Sorry, but I have concerns about this implementation...
Let me elaborate, so you can tell me I shouldn't actually be worried :-)

By keeping such "states" in the object without any db persistence will make any Taipy-core service stateful. Example: a pod running in Kubernetes. 

I see at least two problems now:
 - If we need to "scale up" the pod (because we need more throughput/horse power), then the state remains in on pod but the other has an inconsistent state. The behaviour might become erratic.
 - If the pod dies for some reason, then the state/data is lost forever.

So either this state is important, and thus we need to somehow persist it, either it's not useful and we should just get rid of it.

let me know if we need to discuss it with more details.

toan-quach: Hmmm for now, from what I can understand, as long as orchestrator and dispatcher is on the same application, sharing the same memory pool, this shouldn't be an issue. Previously we also rely on this fact when initiating the callback from job to update the submission status:
- To run callbacks https://github.com/Avaiga/taipy-core/blob/develop/src/taipy/core/job/job.py#L31 this function relies on callbacks store in memory and not in disk/db
- Assigning callbacks to jobs https://github.com/Avaiga/taipy-core/blob/develop/src/taipy/core/_orchestrator/_orchestrator.py#L81 we never store these callbacks all the way until after we have finished running the job in https://github.com/Avaiga/taipy-core/blob/develop/src/taipy/core/_orchestrator/_dispatcher/_job_dispatcher.py#L124 where we set everything and consequently also saving the callbacks of the job into the disk/db

But you made a very valid point, in the scenario where we want to scale up, separating the dispatcher to a separate application or even, a separate machine, this won't work anymore. I think we should discuss this point further :D and we should also involve @jrobinAV as I think this topic would be interesting for this ticket also https://app.zenhub.com/workspaces/taipy-613ef328affde6000ff77fb3/issues/gh/avaiga/taipy-enterprise/268

src/taipy/core/submission/submission.py
```diff
@@ -130,56 +145,46 @@ def __gt__(self, other):
...
-            if job.is_running():
-                running = True
-                continue
-            if job.is_completed() or job.is_skipped():
-                completed = True
-                continue
-            if job.is_abandoned():
-                abandoned = True
-        if canceled:
+    def _update_submission_status(self, job: Job):
```
jrobinAV, line 148: Should we pass the status of the job directly instead of the job?
1. So we don't reload the job at all.
2. So we are sure to process all the statuses, in case the status already changed when reloading the job.

src/taipy/core/submission/submission.py
```diff
@@ -130,56 +145,46 @@ def __gt__(self, other):
...
+        if self._submission_status == SubmissionStatus.FAILED or self._submission_status == SubmissionStatus.CANCELED:
+            return
+
+        job_status = job.status
```
jrobinAV, line 152: This is related to the previous comment.

Here, the job status might have already changed when reloaded.

src/taipy/core/submission/submission.py
```diff
@@ -9,16 +9,17 @@
 # an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the
 # specific language governing permissions and limitations under the License.
 
+import threading
 import uuid
 from datetime import datetime
-from typing import Any, List, Optional, Union
+from typing import Any, List, MutableSet, Optional, Union
```
jrobinAV, line 15: Use [collections.abc.MutableSet](https://docs.python.org/3/library/collections.abc.html#collections.abc.MutableSet) instead?

src/taipy/core/submission/submission.py
```diff
@@ -130,56 +145,46 @@ def __gt__(self, other):
...
+
+        with self.lock:
+            if job_status == Status.BLOCKED:
+                self.__blocked_jobs.add(job.id)
+            if job_status == Status.PENDING or job_status == Status.SUBMITTED:
+                self.__pending_jobs.add(job.id)
+                self.__blocked_jobs.discard(job.id)
+            if job_status == Status.RUNNING:
+                self.__running_jobs.add(job.id)
+                self.__pending_jobs.discard(job.id)
```
jrobinAV, line 169: We must remove it from `__blocked_jobs` as well, in case the job went from blocked to running directly.

src/taipy/core/submission/submission.py
```diff
@@ -130,56 +145,46 @@ def __gt__(self, other):
...
+            if job_status == Status.COMPLETED or job_status == Status.SKIPPED:
+                self.__completed = True
+                self.__running_jobs.discard(job.id)
+            if job_status == Status.ABANDONED:
+                self.__abandoned = True
```
jrobinAV, line 174: if the job is abandonned, don't you need to discard it from the `__running_jobs`, `__blocked_jobs`, and `__pending_jobs`?

src/taipy/core/submission/submission.py
```diff
@@ -130,56 +145,46 @@ def __gt__(self, other):
...
+                self.__blocked_jobs.add(job.id)
+            if job_status == Status.PENDING or job_status == Status.SUBMITTED:
+                self.__pending_jobs.add(job.id)
+                self.__blocked_jobs.discard(job.id)
+            if job_status == Status.RUNNING:
+                self.__running_jobs.add(job.id)
+                self.__pending_jobs.discard(job.id)
+            if job_status == Status.COMPLETED or job_status == Status.SKIPPED:
+                self.__completed = True
+                self.__running_jobs.discard(job.id)
```
jrobinAV, line 172: Once it is completed, we should remove it from all the sets. No?

src/taipy/core/submission/submission.py
trgiangdo, line 174: Could we use `if ... elif` here?

src/taipy/core/submission/submission.py
```diff
@@ -130,56 +145,46 @@ def __gt__(self, other):
...
+        elif self.__blocked_jobs:
             self.submission_status = SubmissionStatus.BLOCKED  # type: ignore
-            return
-        if completed:
+        elif self.__completed:
             self.submission_status = SubmissionStatus.COMPLETED  # type: ignore
-            return
-        self.submission_status = SubmissionStatus.UNDEFINED  # type: ignore
+        else:
+            self.submission_status = SubmissionStatus.UNDEFINED  # type: ignore
```
trgiangdo, line 187: Why do we need the `# type: ignore` here and all other `self.submission_status = SubmissionStatus.STATUS`?

src/taipy/core/_orchestrator/_orchestrator.py
```diff
@@ -222,7 +225,6 @@ def _on_status_change(cls, job: Job):
         if job.is_completed() or job.is_skipped():
             cls.__unblock_jobs()
         elif job.is_failed():
-            print(f"\nJob {job.id} failed, abandoning subsequent jobs.\n")
```
trgiangdo, line 225: Can we keep this message?
Or log it instead of `print`?

Or when the job is failed, we log a message saying why the job fails. It's better for the user, but probably quite tricky.

src/taipy/core/_orchestrator/_orchestrator.py
```diff
@@ -295,7 +297,6 @@ def _fail_subsequent_jobs(cls, failed_job: Job):
                 cls.__find_subsequent_jobs(failed_job.submit_id, set(failed_job.task.output.keys()))
             )
             for job in to_fail_or_abandon_jobs:
-                print(f"Abandoning job: {job.id}")
```
trgiangdo, line 298: Can we keep this message?
Or log it instead of `print`?

Or when the job is abandoned, we log a message saying why the job is abandoned. It's better for the user to understand the problem, but probably quite tricky.

src/taipy/core/submission/submission.py
trgiangdo: Isn't that the dispatcher and orchestrator should always on the same machine, and we only split the jobs?

src/taipy/core/_orchestrator/_orchestrator.py
gmarabout: We should definitely log (instead of `print`)

tests/core/submission/test_submission.py
```diff
@@ -162,6 +160,7 @@ def test_update_submission_status_with_one_failed_job_in_jobs(job_ids, expected_
         (["job2_canceled", "job6_completed"], SubmissionStatus.CANCELED),
         (["job2_canceled", "job7_skipped"], SubmissionStatus.CANCELED),
         (["job2_canceled", "job8_abandoned"], SubmissionStatus.CANCELED),
+        (["job2_canceled", "job1_failed"], SubmissionStatus.CANCELED),
```
jrobinAV, line 163: Haha. 🤣 You changed the test to match your implementation. 
This test case is incorrect.
The expected submission status is not `CANCELED` but `FAILED`.

tests/core/submission/test_submission.py
```diff
@@ -139,7 +138,6 @@ def test_update_single_submission_status(job_ids, expected_submission_status):
         (["job1_failed", "job6_completed"], SubmissionStatus.FAILED),
         (["job1_failed", "job7_skipped"], SubmissionStatus.FAILED),
         (["job1_failed", "job8_abandoned"], SubmissionStatus.FAILED),
-        (["job2_canceled", "job1_failed"], SubmissionStatus.FAILED),
```
jrobinAV, line 142: Same comment. You changed the test to match your implementation. 
The test case was valid and still is.
The expected submission status is `FAILED` but not `CANCELED`.

src/taipy/core/submission/submission.py
gmarabout: @trgiangdo In theory, if you consider Kubernetes as the deployment platform, you shouldn't make such assumptions. Services running on K8s must natively be stateless and restartable without breaking the app (the so-called "cloud native"). 
And if it can only run once (i.e. a single process that won't scale), then it must also be ready to be restarted on other nodes without notice (K8S does that). It means the state must be easily restored or recomputed. In this case, it would just be emptied...
I don't want to block the PR, but design choices imply things that go beyond the scope of the code.

src/taipy/core/submission/submission.py
jrobinAV, line 148: I am concern about the the possibility to call this method for the same job for 2 (or more) status changes in the wrong order.

tests/core/submission/test_submission.py
jrobinAV, line 1: In this file we only tests the update_submission method without initial state.
I am concern it does not cover all the cases.

src/taipy/core/submission/submission.py
toan-quach: weird thing is that for setter, we have to or the linter will complain, I noticed we used `type: ignore` everywhere when we use a setter that has self_setting enabled

src/taipy/core/_orchestrator/_orchestrator.py
toan-quach: this was a print that I used when debugging an issue that I forgot to remove :D

toan-quach: and yes I do agree that we need better logging

@jrobinAV should we have a different ticket to tackle this?

src/taipy/core/submission/submission.py
toan-quach: Good idea, but the current implementation indicates that we're passing the job to the callbacks needed to be run and we don't have an approach to say "we will be updating the submission status now" to determine that we should pass the status only instead of the whole job

src/taipy/core/submission/submission.py
toan-quach: I have answered your question above so will resolve this comemnt for now 😃

src/taipy/core/submission/submission.py
toan-quach: if it's completed, it must be in running first, and by that, it has already been removed from blocked jobs and pending jobs

src/taipy/core/submission/submission.py
toan-quach: the job will be set to pending and triggered the function to remove the job from blocked_jobs and add it to pending_jobs so no need to discard it from blocked_jobs imo as it has already been done

src/taipy/core/_orchestrator/_orchestrator.py
toan-quach: It's a print I used for debugging previously that I forgot to remove

As for the logging issue, I think we need a different ticket for it as it might require more than a simple change @jrobinAV

jrobinAV: We have a log drped from the `job.update_status` method. Is it not enough?

src/taipy/core/_orchestrator/_orchestrator.py
jrobinAV: We can log a message when canceling or abandoning a job in the method `_orchestrator._cancel_job`.

tests/core/submission/test_submission.py
toan-quach: I have added tests regarding the tests for the initial states

src/taipy/core/submission/submission.py
toan-quach: as discussed, the status of a job will be finished updated by the orchestrator before handing it over to the dispatcher so no update to job status shall create a collision. The only scenario this might happen is when we try to cancel a job as this action will be trigger directly from the main thread while the job could be running in the subthread. But we have various conditions to check if we can cancel a job or not, if it's running then we can't, if it's not running, I think this process can still be triggered safely when we update the status of job to `canceled`

src/taipy/core/_orchestrator/_orchestrator.py
toan-quach: We have already log that information https://github.com/Avaiga/taipy-core/blob/develop/src/taipy/core/_orchestrator/_orchestrator.py#L245

src/taipy/core/submission/submission.py
toan-quach: I believe we can close this topic for now.
To summarize:
- We acknowledge that there are several places that are stateful that we don't have an approach to serialize them across multiple machines yet (including job callbacks, state of the submission entity, the event notification, etc.)
- We will not these issues right now and will review them again later

toan-quach: I believe the discussion has been made and further talk is linked in the comment below

gmarabout: Yes, we can close the topic.

src/taipy/core/_orchestrator/_orchestrator.py
toan-quach: just had a look and I think it's sufficient

src/taipy/core/_orchestrator/_orchestrator.py
jrobinAV: No log in the else block which correspond to the case where we do cancel and abandon jobs.

jrobinAV:
The code is already merged in taipy repo, right?
Can we close this PR?

toan-quach:
@jrobinAV yep we can! Sorry for that!

PR #833, opened 2023-11-29
feature/ added entity type to submission 

toan-quach:


jrobinAV (dismissed):
It looks good to me.
We can merge it if we want, but the most important is to do it again in taipy repository once core repo is merged into taipy repo.

@toan-quach Please let us know when the PR is opened on taipy.

Thx!

tests/core/job/test_job_manager.py
```diff
@@ -349,8 +349,8 @@ def test_cancel_subsequent_jobs():
     task_3 = Task("task_config_3", {}, print, [dn_4], id="task_3")
 
     # Can't get tasks under 1 scenario due to partial not serializable
-    submission_1 = submission_manager._create("scenario_id")
-    submission_2 = submission_manager._create("scenario_id")
+    submission_1 = submission_manager._create("scenario_id", "SCENARIO")
```
gmarabout: 🥜 It's really a neat, but maybe we could use `Scenario._ID_PREFIX` instead...

src/taipy/core/submission/submission.py
```diff
@@ -44,13 +44,15 @@ class Submission(_Entity, _Labeled):
     def __init__(
         self,
         entity_id: str,
+        entity_type: str,
```
gmarabout, line 47: 💬 It means that we can pass any str (i.e. not an enum). 
What should we expect if we pass a string that does not correspond to a real entity?

gmarabout (dismissed):
I left a couple of comments (they are more questions actually), but looks great to me!

src/taipy/core/submission/submission.py
toan-quach: hmmm good point, as this is only an entity for internal usage, I hope this won't be an issue 😅

tests/core/job/test_job_manager.py
toan-quach: ah haha sure thing!

gmarabout (approved):

jrobinAV:
The code is already merged in taipy repo, right?
Can we close this PR?

toan-quach:
Yep can! Let me close it!

PR #834, opened 2023-11-29
Event Creation: Add version and make sure we don't call autoreload attributes

gmarabout:
# Goal
* Make sure all events are created without accessing fields that would "auto reload" the entity (and thus could create performance issues)
* Add `version` in metadata of entities 

# Changed
* Data_node: add `_version`
* Job: Use internal fields to by-pass autoreload
* Scenario: Use internal fields to by-pass autoreload
* Submission: small refactoring
* Cycle: no change, there is no version for it?

### DIFF
--- a/src/taipy/core/data/data_node.py
+++ b/src/taipy/core/data/data_node.py
@@ -567,7 +567,7 @@ def make_event_for_datanode(
     attribute_value: Optional[Any] = None,
     **kwargs,
 ) -> Event:
-    metadata = {"config_id": data_node.config_id, **kwargs}
+    metadata = {"config_id": data_node.config_id, "version": data_node._version, **kwargs}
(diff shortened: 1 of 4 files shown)


jrobinAV (approved):

toan-quach (approved):
looks good to me! but can you quickly explain why we wouldn't want to use auto reload here? Is the consistency of the value already guaranteed?

gmarabout:
> looks good to me! but can you quickly explain why we wouldn't want to use auto reload here? Is the consistency of the value already guaranteed?

@toan-quach I actually *don't want* autoreload to happen! This function must have the least performance penalty possible 🚀

jrobinAV (commented):
It must be re done in taipy repository !

